PR #5, opened 2023-04-18
Tests for validation of DCAT-US records

rshewitt:
# Pull Request

Related to [ticket](https://github.com/GSA/data.gov/issues/4256)

## About

Tests for validation of DCAT-US records. All JSON files in this [directory](https://github.com/GSA/ckanext-datajson/tree/main/ckanext/datajson/tests/datajson-samples) are used. 

## PR TASKS

- [ ] Any changes to docs?
- [ ] Bumped version number in [setup.py](https://github.com/GSA/ckanext-geodatagov/blob/main/setup.py#L13) (also checked on [PyPi](https://pypi.org/project/ckanext-geodatagov/#history)).


tests/test_demo.py
```diff
@@ -1,8 +1,40 @@
...
+    json_data = open_json(file_path)
+
+    output = None
+
+    try:
+        jsonschema.validate(json_data, schema=dataset_schema)
+        output = is_valid
+    except jsonschema.ValidationError as e:
+        error_message = "error: " + e.message + ". offending element: " + e.json_path
+        print(error_message)
```
jbrown-xentity: Think of this as the function we are trying to test. So we should define an "extract" folder, and that folder should have a dcat_us.py file, and that file should have a function called extract, and this is the logic (more or less) that should go in it.
Then this test calls out to said extract function, and then validates the expected output given the various json inputs.

tests/test_demo.py
```diff
@@ -1,8 +1,40 @@
...
-    assert hello('name') == "Hello name!"
\ No newline at end of file
+    assert hello('name') == "Hello name!"
+
+
+@pytest.mark.parametrize(
+    "schema, dataset, is_valid",
+    open_json(TESTS_DIR / "fixtures" / "parametrizers" /
+              "params.json")["test_validate_dcat_us"]
+)
```
jbrown-xentity: This is really cool and fancy! But maybe overkill, and may make it difficult to read in the future.
I would have built a "parsing" function that abstracts all the "getting json fixture" code. Then I would have a separate test for each json use case, and it would be some copy/paste. In general it would look like the following, with slight customizations for each json file. This allows for comments, and explicit naming for each test case.
```
def test_numerical_title()
    json_to_parse = custom_fixture_function("numerical-title.data.json")
    try:
        jsonschema.validate(json_data, schema=dataset_schema)
        assert false #this should fail validation, numerical titles aren't allowed
    except jsonschema.ValidationError as e:
        error_message = "error: " + e.message + ". offending element: " + e.json_path
        print(error_message)
        assert true
```

jbrown-xentity (approved):

tests/test_demo.py
jbrown-xentity: I clearly was in the wrong headspace.. replace everything above that says `extract` with `validate`

rshewitt:
Spent some time better understanding [the pyproject.toml file](https://python-poetry.org/docs/pyproject/#packages). Mistakenly deleted the example package and replaced the `--cov` value in the action yml.

jbrown-xentity (approved):

datagovharvester/utils/json_utilities.py
```diff
@@ -0,0 +1,6 @@
+import json
+
+
+def open_json(file_path):
+    with open(file_path) as fp:
+        return json.load(fp)
```
nickumia-reisys, line 6: Is this function really necessary?  What's the long-term benefit of it?  It seems like a frivolous abstraction...

datagovharvester/validate/dcat_us.py
```diff
@@ -0,0 +1,31 @@
...
+
+def parse_errors(errors):
+    error_message = ""
+
+    for error in errors:
+        error_message += (
+            f"error: {error.message}. offending element: {error.json_path} \n"
+        )
+
+    return error_message
```
nickumia-reisys, line 13: Is this the best consolidation of errors that makes sense?  A single string of an unknown number of errors?  Isn't a list of strings more flexible?

tests/validate/dcatus/conftest.py
```diff
@@ -0,0 +1,111 @@
+import pytest
+from pathlib import Path
+from datagovharvester.utils.json_utilities import open_json
+
+BASE_DIR = Path(__file__).parents[3]
+DATA_DIR = BASE_DIR / "data" / "dcatus"
+SCHEMA_DIR = DATA_DIR / "schemas"
+JSON_DIR = DATA_DIR / "jsons"
```
nickumia-reisys, line 8: This is really cool 👍

tests/validate/dcatus/conftest.py
```diff
@@ -0,0 +1,111 @@
...
+def open_geospatial_json():
+    json_file = JSON_DIR / "geospatial.data.json"
+    return open_json(json_file)
+
+
+# valid
+@pytest.fixture
+def open_null_spatial_json():
+    json_file = JSON_DIR / "null-spatial.data.json"
+    return open_json(json_file)
```
nickumia-reisys, line 111: These tests seem a bit wasteful... Can we have a datajson creator that we pass different parameters to generate the datasets instead of hardcoding the files and having a growing number of open functions?  It is a different design choice, but I think it'll be more sustainable than this historical way of running tests.

datagovharvester/validate/dcat_us.py
```diff
@@ -0,0 +1,31 @@
...
+    try:
+        validator.validate(json_data)
+        success = True
+        error_message = "no errors"
+    except ValidationError:
+        success = False
+        errors = validator.iter_errors(json_data)
+        error_message = parse_errors(errors)
+
+    return success, error_message
```
nickumia-reisys, line 31: Can we somehow highlight that this is the only real core code that was added?  Add docstrings and a feature list in the README?

nickumia-reisys (commented):
🦯

PR #7, opened 2023-05-09
Extract for DCAT-US

rshewitt:
# Pull Request

Related to [Develop extract for DCAT-US ticket](https://github.com/orgs/GSA/projects/11/views/1?pane=issue&itemId=23777726)

## About

- Added DCAT-US extraction functionality.
- Added tests to verify DCAT-US extraction works.
- Added nginx and localstack services to docker compose yml.
- Added 3 python libraries to poetry dependencies. 

## PR TASKS

- [x] Any changes to docs?
- [ ] Bumped version number in [setup.py](https://github.com/GSA/ckanext-geodatagov/blob/main/setup.py#L13) (also checked on [PyPi](https://pypi.org/project/ckanext-geodatagov/#history)).


.env
```diff
@@ -0,0 +1,7 @@
+S3FILESTORE__AWS_BUCKET_NAME=test-bucket
+S3FILESTORE__AWS_STORAGE_PATH=local
+S3FILESTORE__REGION_NAME=us-east-1
+S3FILESTORE__HOST_NAME=http://127.0.0.1:4566
+S3FILESTORE__AWS_ACCESS_KEY_ID=_placeholder
+S3FILESTORE__AWS_SECRET_ACCESS_KEY=_placeholder
+S3FILESTORE__SIGNATURE_VERSION=s3v4
```
nickumia-reisys: Are we keeping the `S3FILESTORE` naming convention?  Also, are we keeping the `ckanext-envvar` variable naming conventions (with the `_` `__` and `___`)?

.gitignore
```diff
@@ -8,4 +8,7 @@ venv/
 
 # previous versions of schemas
 **/catalog_d4.json
-**/dataset_d4.json
\ No newline at end of file
+**/dataset_d4.json
```
btylerburton: can glob all with `**/dataset_**.json`

datagovharvester/extract/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    except requests.exceptions.RequestException as e:
+        error = e
+
+    return resp, error
+
+
+def fetch_url(url):
+    success = False
+
+    resp, error_msg = download_catalog(url)
```
btylerburton: what are we doing with the error_msg?

datagovharvester/extract/utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+    S3, S3_create_error_msg = create_s3_client(S3_config)
+
+    bucket_name = os.getenv("S3FILESTORE__AWS_BUCKET_NAME")
+    bucket_name, bucket_error_msg = create_bucket(S3, bucket_name)
+
+    for idx, record in enumerate(data["dataset"]):
+        record = json.dumps(record)
+        key_name = f"{job_id}_{idx}_extract.json"
+        s3_payload = create_s3_payload(record, bucket_name, key_name)
```
nickumia-reisys: Is this a `dcat-us` specific schema operation?  Do we want a separate function for this?  Or is this function `dcat-us` specific?

datagovharvester/utils/s3_utilities.py
```diff
@@ -0,0 +1,45 @@
...
+def create_s3_payload(json_str, bucket_name, key_name):
+    return {
+        "Body": json_str,
+        "Bucket": bucket_name,
+        "Key": key_name,
+        "ContentType": "application/json",
+    }
+
+
+def upload_dcatus_to_S3(S3, s3_payload):
```
nickumia-reisys: Is this `dcatus` specific?  Or will this be used for all uploads to `s3`?

datagovharvester/extract/utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+        upload_data, upload_error_message = upload_dcatus_to_S3(S3, s3_payload)
+
+    return upload_data, upload_error_message
```
btylerburton: don't we need to push `upload_data` & `upload_error_message` into something outside the for loop and then return that?

tests/extract/test_dcatus.py
```diff
@@ -0,0 +1,18 @@
+from datagovharvester.extract.utils import extract_catalog
+
+
+def test_extract_dcatus(create_s3_client_config, get_dcatus_job):
+    success = False
+
+    url = get_dcatus_job["url"]
+    job_id = get_dcatus_job["job_id"]
+
+    S3_config = create_s3_client_config
```
nickumia-reisys: This seems like an extraneous line of code...

datagovharvester/utils/s3_utilities.py
```diff
@@ -0,0 +1,45 @@
...
+
+def create_bucket(S3, bucket_name):
+    error_message = None
+
+    try:
+        S3.create_bucket(Bucket=bucket_name)
+    except Exception as e:
+        error_message = e
+
+    return bucket_name, error_message
```
btylerburton: does `create_bucket` return anything, do we need to capture it?

docker-compose.yml
```diff
@@ -0,0 +1,28 @@
...
+    privileged: true
+    image: localstack/localstack:1.3.1
+    ports:
+      - "4566:4566"
+      - "8081:8081"
+    environment:
+      - SERVICES=s3
+      - DEBUG=1
+      - DATA_DIR=/tmp/localstack/data
+      - HOSTNAME=
```
btylerburton: should this be empty?

tests/extract/conftest.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+from dotenv import load_dotenv
+import os
+from uuid import uuid4
+
+
+@pytest.fixture
+def get_dcatus_job():
+    return {
+        "url": "http://localhost:80/data.json",
```
btylerburton: don't think we need port `80` as that's default

datagovharvester/extract/utils.py
```diff
@@ -0,0 +1,52 @@
...
+    upload_dcatus_to_S3,
+    create_s3_client,
+    create_bucket,
+    create_s3_payload,
+)
+
+
+def download_catalog(url):
+    resp = None
+    error = None
```
nickumia-reisys: I think having a docstring would be more useful than adding unnecessary initializers for python code.

nickumia-reisys (dismissed):
Please add the new features to the README file or some other doc file.

btylerburton:
*Testing Slack webhook...*

datagovharvester/extract/utils.py
rshewitt: I would consider this to be a `dcatus` schema operation. I think isolating this as a separate function makes sense.

datagovharvester/utils/s3_utilities.py
rshewitt: All `s3` uploads. I figure the payload could change depending on the format.

tests/extract/test_dcatus.py
rshewitt: I can see that. Would the idea be to pass `get_dcatus_job["url"]`, `get_dcatus_job["job_id"]`, and `create_s3_client_config` to the intended function `extract_catalog` like `extract_catalog( get_dcatus_job["url"], get_dcatus_job["job_id"], create_s3_client_config)`?

.env
btylerburton: Yeah no reason not to follow a more generic pythonic convention.

datagovharvester/extract/utils.py
rshewitt: Can you explain this further? I follow this pattern so my outputs are always at the top, I only return once ( at the end of the function ), I always return the same number of items, and without any undefined issues. I don't always follow my own rules though haha.

datagovharvester/extract/utils.py
rshewitt: Nothing is done with error messages at the moment. I added this with the intention of using the error message at a later date.

.env
rshewitt: [constants in PEP8](https://peps.python.org/pep-0008/#constants)?

.gitignore
rshewitt: I like this. thanks for this.

datagovharvester/extract/utils.py
rshewitt: we can definitely package the results and return them.

datagovharvester/utils/s3_utilities.py
rshewitt: It returns a dict as `ResponseMetadata` with the location of the newly created bucket ( e.g. `Location: /test-bucket`.

tests/extract/conftest.py
rshewitt: gotcha, I can omit that.

datagovharvester/utils/s3_utilities.py
nickumia-reisys: To that point, maybe take `dcatus` out of the name?

.github/workflows/test.yml
```diff
@@ -22,20 +24,25 @@ jobs:
...
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: ${{ env.PY_VERSION }}
 
       - name: Install Poetry
-        run: curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
+        uses: abatilo/actions-poetry@v2
+        with:
+          poetry-version: ${{ env.POETRY_VERSION }}
```
robert-bryson, line 34: Any reason to install ourselves vs use an action?

datagovharvester/extract/utils.py
nickumia-reisys: Yeah, so, to that point, extract has three components: (1) Download, (2) Parse, (3) Upload to S3.

The `Download` needs special logic to handle WAF vs. a single file download.  This could be abstracted to the "fetching_url" function.  The `Parsing` would require some if-then logic or dictionary mapping to decide in which way to parse the catalog into individual records.  The `Upload to S3` sounds like it could be neutral across all cases and I don't see a lot of issue with that aspect yet.

tests/extract/test_dcatus.py
nickumia-reisys: I would do something like: 
```python
final_res, final_error_msg = extract_catalog(
    get_dcatus_job["url"],
    create_s3_client_config,
    get_dcatus_job["job_id"]
)
```

...until such time where a special variable is needed (which I don't envision).

datagovharvester/extract/utils.py
nickumia-reisys: Oh yeah, I didn't even realize that it's just being lost in the function.  I agree we need a more robust error capturing algoirthm.  Each function returning errors leads to holes where errors can get lost.

datagovharvester/extract/utils.py
nickumia-reisys: I agree with @btylerburton... Right now, the `upload_data, upload_error_message` is being continuously overwritten.. I'm not sure what the intent to returning it is.  Once it's uploaded to S3, I don't think we need the data for anything, just the error messages which requires a different solution as mentioned in a different comment.

datagovharvester/utils/s3_utilities.py
nickumia-reisys: I'm not sure how I feel about this.  We only need to actually `create_bucket` for local testing @btylerburton.  On `prod`, the bucket will exist forever.  So it is less critical to capture anything from the output here.  But for what it's worth, this function is trivial as well.  I think it only makes sense to have this as a function if it was part of a class.  But the only added feature is "error handling" which should probably be done in a different way...

For context, there is a school of thought surrounding encapsulated functions (in the functional programming world) whereby there could be a theoretical `error_handling` function that wraps every single function call, so there would only be a need to write error_handling in one function as opposed to each individual function.  However, this makes sense if we are going for a functional programming style.  With classes, there should be an error handling class that all other classes inherit from... So... I think we need to talk about this more...

tests/extract/conftest.py
nickumia-reisys: I agree it seems unnecessary..

datagovharvester/extract/utils.py
rshewitt: Yeah those components make sense. My current approach for WAF is parsing the html, filtering anchors with hyperlinks, then select all files with some file extension or a folder to continue the search. For the WAF, are we to expect all urls to be the root or will we have to traverse upwards? Yeah i agree about `Upload to S3`.

datagovharvester/utils/s3_utilities.py
rshewitt: yeah will do.

.env
rshewitt: I don't mind having `S3FILESTORE` in the beginning. it provides more context on what the values represent. Omitting it (e.g. `AWS_BUCKET_NAME`) does simplify things though and if having them more general won't conflict with anything then i'm down for that. I'm gonna follow PEP8 and use `_`.

docker-compose.yml
rshewitt: I copy/pasted this service from [inventory](https://github.com/GSA/inventory-app/blob/main/docker-compose.yml#L55-L71). The [ localstack documentation](https://docs.localstack.cloud/references/configuration/#debugging:~:text=all%20API%20invocations.-,HOSTNAME,-localhost%20(default)) indicates `HOSTNAME` defaults to localhost so I think we can remove it? There's no mention of `START_WEB` which could indicate it's not used anymore? not sure. gonna test it.

nickumia-reisys:
@rshewitt If you feel stuck, please reach out to the team and we can have an interactive review to clarify things..

datagovharvester/utils/s3_utilities.py
btylerburton: As personal preference, I subscribe to the functional paradigm, but I'm open to either, and agree that this should be something we chat about further in depth

datagovharvester/extract/dcatus.py
```diff
@@ -0,0 +1,26 @@
+from .utils import fetch_url
+import json
+from datagovharvester.utils.s3_utilities import (
+    upload_to_S3,
+    create_s3_payload,
+)
+
+
+def extract_dcatus_catalog(url, job_id, S3_client, bucket_name):
```
btylerburton: I might be missing something, but are we making this specific to `dcatus` because we anticipate making this function more `dcatus` specific?

datagovharvester/extract/dcatus.py
```diff
@@ -0,0 +1,26 @@
...
+    try:
+        data, fetch_success = fetch_url(url)
+
+        for idx, record in enumerate(data["dataset"]):
+            record = json.dumps(record)
+            key_name = f"{job_id}_{idx}_extract.json"
+            s3_payload = create_s3_payload(record, bucket_name, key_name)
+
+            upload_data, upload_error_message = upload_to_S3(S3_client, s3_payload)
+            extracted_data.append([upload_data, upload_error_message])
```
btylerburton: do we want to refactor this into a series of try/catch blocks here? like, one for `fetch_url` and one nested in the for loop for `upload_to_s3`?

datagovharvester/extract/dcatus.py
rshewitt: The only `dcatus`-ness in the function is the `dataset` key access and the json-stringify of each record/object in the `dataset` array ( i.e. `json.dumps(record)` ). Besides that everything else should be format agnostic. Actually that's not true. The content-type of the payload is hard-coded as `application/json` in the `create_s3_payload` function. This will obviously change.

datagovharvester/extract/dcatus.py
rshewitt: Yeah i think that's better. Attempt ( try/catch) to `fetch_url`, add a check to see if `dataset` is a key in `data` ( new addition I think helps ), and attempt to assemble the S3 payload and put the object.

.env
nickumia-reisys: Yeah, definitely stick with at least one `_`, but the point was that `ckanext-envvars` associated different meanings to 1, 2 or 3 underscores.  I think it's important to use a descriptive name.  I just personally never liked the name "s3filestore".  This is not a blocking comment for me.  And I won't suggest a change unless others also don't like the name.  

If we leave it as is, I say we stay consistent with `<feature>__<variable>` with the two underscores separating context and variable name.

datagovharvester/extract/dcatus.py
btylerburton: cool. looks like you've abstracted out most of the generic stuff, so we should be good.

datagovharvester/extract/utils.py
btylerburton: feels to me like we could actually do the `get` inside `fetch_url`

datagovharvester/extract/utils.py
nickumia-reisys: Python is a loosely-typed language.  There is also no concept of variable initialization in Python.  To those points, there are very very very .... very few usecases that you need to explicitly initialize a variable.  Since `resp` and `error` are simple variable assignments, no need to initialize them with `None`.

To track "outputs", you can have a docstring that outlines the inputs/outputs and general purpose of the function... In this case....

```python
def download_catalog(url):
    ''' Download agency source catalog data
    url (str)     : https link to download data
    resp (str)    : data download
    error (error) : errors encountered when downloading
    '''
    ...the rest of the function...
```

nickumia-reisys: ... but be more specific, is `error` an instance of error class? or it is a list of errors?  Or something else?

.github/workflows/test.yml
rshewitt: I think using an action is preferable.

.env
rshewitt: I think using `<feature>__<variable>` is good.

datagovharvester/extract/utils.py
rshewitt: Pin this topic to a pairing session for group discussion.

datagovharvester/extract/utils.py
rshewitt: Yeah we can combine `download_catalog` and `fetch_url`.

datagovharvester/extract/utils.py
rshewitt: The return of `upload_to_s3` will be handled at a later date. currently, nothing is done with the return.

datagovharvester/utils/s3_utilities.py
rshewitt: Handling errors at a later date. [reference](https://github.com/GSA/datagov-harvesting-logic/pull/7/files/e48cf1894b656f00cbd4a0e9a6350e485935752f#r1192794433)

datagovharvester/extract/utils.py
rshewitt: Should one function handle one exception though?

nickumia-reisys:
Things to decide on as a team before we continue with this PR and future ones:
- [ ] What should live in a class vs. a function?  (i.e. what are our boundaries for object-oriented vs functional programming)
  - [ ] By extension, how do we implement error handling?
- [ ] What is our documentation style?  (e.g. are docstrings a must, where are readmes, et cetera...)
- [ ] ...

There was something else, @rshewitt has the full list.

robert-bryson:
> Things to decide on as a team before we continue with this PR and future ones:
> 
> * [ ]  What should live in a class vs. a function?  (i.e. what are our boundaries for object-oriented vs functional programming)
>   
>   * [ ]  By extension, how do we implement error handling?
> * [ ]  What is our documentation style?  (e.g. are docstrings a must, where are readmes, et cetera...)
> * [ ]  ...
> 
> There was something else, @rshewitt has the full list.

I think these are good considerations to be having, especially now as we are adding new code. 

As per OO/functional paradigm, I don't have strong feelings but from experience know performance can be an issue in Python if you are sending large amounts of data to each and every function. 

For documentation, we should at least be adding docstrings to new code. I also like type hints. I know that code quality has been contentious on the past so I'm glad you're bringing it up for discussion, @nickumia-reisys .

nickumia-reisys:
Python is not a true object-oriented programming language.  It supports OOP.  But with functional programming, you don't need to pass large amounts of data back and forth.  Pointers are a great utility for functional programming to optimize functional communication.

rshewitt:
> Things to decide on as a team before we continue with this PR and future ones:
> 
> * [ ]  What should live in a class vs. a function?  (i.e. what are our boundaries for object-oriented vs functional programming)
>   
>   * [ ]  By extension, how do we implement error handling?
> * [ ]  What is our documentation style?  (e.g. are docstrings a must, where are readmes, et cetera...)
> * [ ]  ...
> 
> There was something else, @rshewitt has the full list.

What format we want to use for env vars? We spoke about `<feature>__<variable>`

datagovharvester/extract/dcatus.py
```diff
@@ -0,0 +1,33 @@
...
+            record = json.dumps(record)
+            key_name = f"{job_id}_{idx}_extract.json"
+            s3_payload = create_s3_payload(record, bucket_name, key_name)
+            upload_to_S3(S3_client, s3_payload)
+        except Exception as e:
+            # do something with the error
+            errors.append(e)
+            continue
+
+    return errors
```
jbrown-xentity: Return the job id's and the s3 location, not just the errors.
Any errors that occur, should be logged in the DB via a help function TBD. However, we need some process to end the pipeline for an extract if it is an error that means there is nothing left for the pipeline to do. If it is a record error or soft error, log and continue.

jbrown-xentity: Instead of having a helper function for the URL request and the main function do the record processing, have the main function do the request and call a helper function that processes the record. This will make for easier unit testing.

datagovharvester/utils/s3_utilities.py
```diff
@@ -0,0 +1,45 @@
...
+
+def create_s3_bucket(S3, bucket_name):
+    error_message = None
+
+    try:
+        S3.create_bucket(Bucket=bucket_name)
+    except Exception as e:
+        error_message = e
+
+    return bucket_name, error_message
```
jbrown-xentity: This is an unlikely function. Can probably just be run wholesale in the tests, and not implemented here.

datagovharvester/extract/dcatus.py
```diff
@@ -0,0 +1,33 @@
...
+        return e
+
+    if not data.get("dataset", None):
+        # do something with the error
+        return "no dataset key"
+
+    for idx, record in enumerate(data["dataset"]):
+        try:
+            record = json.dumps(record)
+            key_name = f"{job_id}_{idx}_extract.json"
```
jbrown-xentity: I would actually have this as `source_id/job_id/idx.json`. This allows for easier browsing by humans, if we have to debug manually.

tests/extract/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+
+@pytest.fixture
+def create_client(create_client_config):
+    return create_s3_client(create_client_config)
+
+
+@pytest.fixture
+def create_bucket(create_client):
+    s3_client, s3_client_msg = create_client
+    return create_s3_bucket(s3_client, "test-bucket")
```
jbrown-xentity: We'll need to add tests after this. In the future we will follow [Railway Oriented Programming](https://swlaschin.gitbooks.io/fsharpforfunandprofit/content/posts/recipe-part2.html), in thinking this way we want to capture make sure each "track" (ie every exception) is working as we expect. This should include the following use cases:

1. request error
2. request doesn't give a 200 response code
3. request is not a valid JSON
4. record has no `dataset` key
5. record does not have an `id` (not positive if this is a warning or an error; it won't continue past compare process but maybe we still want to save it's state? I lean towards not saving it, this is a really edge case)
6. s3 upload failure
7. Happy path (record is good and is processed successfully)

jbrown-xentity (commented):
Thanks for the discussion @rshewitt @robert-bryson @nickumia-reisys @btylerburton @FuhuXia , please add anything else you see that we need to implement!

datagovharvester/extract/dcatus.py
rshewitt: changes are reflected [here](https://github.com/GSA/datagov-harvesting-logic/commit/60fa67bec6eed6660ff30c71ee75b485d68270d3)

rshewitt: resolved [here](https://github.com/GSA/datagov-harvesting-logic/commit/60fa67bec6eed6660ff30c71ee75b485d68270d3)

datagovharvester/utils/s3_utilities.py
rshewitt: removed the function and moved bucket creation [here](https://github.com/GSA/datagov-harvesting-logic/blob/54226c79f9e595315b8f4d3e961eb643628be08e/tests/extract/test_dcatus.py#L15)

datagovharvester/extract/dcatus.py
rshewitt: updated [here](https://github.com/GSA/datagov-harvesting-logic/commit/41209f80fc7566d403ccb7400aade30b7f0ac551) and corresponding location in store_record_in_s3 function.

.vscode/launch.json
```diff
@@ -0,0 +1,16 @@
...
+        {
+            "name": "Python: Current File",
+            "type": "python",
+            "request": "launch",
+            "program": "${file}",
+            "console": "integratedTerminal",
+            "justMyCode": true
+        }
+    ]
+}
```
nickumia-reisys: Was this added by accident?

datagovharvester/extract/dcatus.py
```diff
@@ -0,0 +1,56 @@
...
+    record (dict)               :   dcatus record
+    source_id (str)             :   uuid 
+    job_id (str)                :   uuid
+    record_idx (int)            :   index position of record in records list
+    S3_client (boto3.client)    :   boto3 s3 client. 
+    bucket_name (str)           :   name of bucket to store data. 
+    """
+    try:
+        record = json.dumps(record)
+        key_name = f"{source_id}/{job_id}/{record_idx}.json"
```
nickumia-reisys: We should define this in a single place.  I see it referenced in two places. (https://github.com/GSA/datagov-harvesting-logic/pull/7/files#diff-9d01608e3885086371828eef6b60f3143ad31367f557b6019cde7b24559ed8f9R50)

tests/nginx.conf
```diff
@@ -0,0 +1,49 @@
...
+    #    include        fastcgi_params;
+    #}
+
+    # deny access to .htaccess files, if Apache's document root
+    # concurs with nginx's one
+    #
+    #location ~ /\.ht {
+    #    deny  all;
+    #}
+}
```
nickumia-reisys: Do we need all of these comments?

.vscode/launch.json
rshewitt: @jbrown-xentity had mentioned a while ago about adding a debugger configuration to the project. I think it doesn't make sense to add it through this branch though. I'm also okay with omitting it entirely, adding it to gitignore, as people may have different debugging preferences.

datagovharvester/extract/dcatus.py
rshewitt: I agree. I'll change that now.

.vscode/launch.json
nickumia-reisys: Oh, this seems like a vscode specific debugger though?

tests/nginx.conf
rshewitt: No I don't think so. The example config file I had used as reference had these. I'll remove them.

.vscode/launch.json
rshewitt: yeah it's a vscode debugger. Is that a topic for discussion amongst the team?

tests/nginx.conf
rshewitt: [removed comments](https://github.com/GSA/datagov-harvesting-logic/commit/eec840c51ede39b2bb6e11bd6380aaaa0bde0cab)

datagovharvester/extract/dcatus.py
rshewitt: [updated to use 1 reference](https://github.com/GSA/datagov-harvesting-logic/commit/7049e38c44ab4e0a1b4db38c915d849eeb6ffee6)

.vscode/launch.json
rshewitt: I [removed the folder](https://github.com/GSA/datagov-harvesting-logic/commit/b359ab0e9890fc0e26e6c12a6b50263967c3df74)  and [added it to gitignore](https://github.com/GSA/datagov-harvesting-logic/commit/8b09e812d62a7b5e5361668523d5d907c4186fbe) but can easily add it back if need be.

jbrown-xentity: I think the hope was to document the vs code "happy path", but not necessarily put it in code (or save the configuration files). I see this ideally as a wiki page that the readme links to. We don't all have to use the same tools if people don't like vs code, but having a working proven use case in something like vs code allows others to follow documentation in their preferred ide to get a similarly working debugger.

rshewitt: yeah okay that makes sense ( providing users with a practical use case of debugging harvester 2.0 functionality ). It could also be helpful to include an example of debugging tests.

datagovharvester/extract/dcatus.py
```diff
@@ -0,0 +1,57 @@
...
+)
+
+def store_record_in_s3( record, source_id, job_id, record_idx, S3_client, 
+    key_name, bucket_name ):
+    """ store the input record in a s3 bucket
+    record (dict)               :   dcatus record
+    source_id (str)             :   uuid 
+    job_id (str)                :   uuid
+    record_idx (int)            :   index position of record in records list
+    S3_client (boto3.client)    :   boto3 s3 client.
```
robert-bryson: Looks like you intended to include `key_name` here? It's probably worth adding since you return it, also

datagovharvester/extract/dcatus.py
```diff
@@ -0,0 +1,57 @@
+from .utils import download_catalog
```
robert-bryson: The pattern below of `from datagovharvester.utils.xxx` makes sense to me. Why did you want to do something differently here?

datagovharvester/extract/dcatus.py
rshewitt: I ended up nixing this function but good find! I'm gonna double check my docstrings to make sure i'm including everything.

datagovharvester/extract/dcatus.py
rshewitt: Most likely for convenience but with autocomplete that's not really an excuse haha. good find. i've updated my imports to be absolute rather than relative.

datagovharvester/extract/dcatus.py
rshewitt: removal [here](https://github.com/GSA/datagov-harvesting-logic/commit/0a6c08677b3b93cc44a2b30291ec9585f9a8414c)

datagovharvester/extract/dcatus.py
rshewitt: I refactored a bit so i'm not importing the same thing(s). [similar lines in the same module](https://github.com/GSA/datagov-harvesting-logic/blob/ba005adfef0b0f1c2d94fe630bf509d637bf57a1/datagovharvester/extract/dcatus.py#L2-L3)

robert-bryson: Sorry, I don't think I was very clear. Why do we have `utils` here and a `utils` dir with sub `utils`? Do you think it would make more sense to group all of them in the same place (honest question, do you think it is worth it to separate them)?

robert-bryson (approved):

datagovharvester/extract/dcatus.py
rshewitt: My thinking is `datagovharvester.utils` are modules applicable anywhere. For example, the `datagovharvester.utils.s3_utilities.upload_to_S3` function will be, or is currently, used in the `extract`, `transform`, and `load` main functions found in their respective `__init__.py` modules ( currently only `extract` ). `datagovharvester.extract.utils` are functions which are primarily applicable to `extract`. All the utils could be placed in `datagovharvester.utils`. It would simplify things and having all of them in one place could make navigation and the general structure easier to work with.

robert-bryson: Hmm, sounds like both approaches have merit. I bringing it up since it would set a precedent for us: unified or separate `utils`. What does the rest of the @GSA/data-gov-team think?

To be clear @rshewitt, I'm happy to do the refactor. I'm not suggesting you do the work, more how should we approach this now and in the future.

jbrown-xentity (approved):

PR #10, opened 2023-06-15
Setup s3 service

rshewitt:
# Pull Request

Related to [#4308](https://github.com/GSA/data.gov/issues/4308)

## About

- Add `localstack` `S3` docker container. Add test checking container health. 
- Timed `S3` bucket object count of 100k catalogs using `boto3` ( 26 seconds ) and `aws cli` ( 11 seconds ) with `localstack` test bucket. 
- This ticket only introduced a new feature used in development: an s3 service we can test. 

## PR TASKS

- [ ] Any changes to docs?


jbrown-xentity (dismissed):
This all looks good, confirms the errors we are supposed to get and verifies the "happy path". I know that some of this error handling is going to change with https://github.com/GSA/data.gov/issues/4324, we might want to put a few TODO's noting that to make it easier when we get to that point.
Looks good to me!

docker-compose.yml
```diff
@@ -16,6 +16,11 @@ services:
     ports:
       - "4566:4566"
       - "8081:8081"
+    healthcheck:
+      test: ["CMD", "curl", "localhost:4566"]
```
btylerburton: if we're using curl we'd want it fail, like `curl --fail`, I think. also, might be worth adding a TODO to write a custom healthcheck

tests/extract/test_dcatus.py
```diff
@@ -25,6 +25,8 @@ def test_extract_bad_url(get_bad_url, create_client):
 
     if str(main(get_bad_url, S3_client)) == "non-200 status code":
```
btylerburton: you can just assert `VAL = VAL` and skip the conditional in all these.

tests/s3/test_s3.py
```diff
@@ -0,0 +1,12 @@
+from time import sleep 
+
+def test_localstack_is_healthy(get_docker_api_client):
+
+    api_client = get_docker_api_client
```
btylerburton: why are we testing localstack?

btylerburton (changes requested):

docker-compose.yml
rshewitt: yeah I think adding `--fail` makes sense. I didn't know that flag existed. I'll add a TODO for the custom health check.

tests/extract/test_dcatus.py
rshewitt: totally agree. thanks for the recommendation.

tests/s3/test_s3.py
rshewitt: I think it provides added reassurance in the services we use in our testing. we already include a healthcheck on it in the compose file though which could make this test redundant. I've kept the test in for now.

docker-compose.yml
rshewitt: resolved [here](https://github.com/GSA/datagov-harvesting-logic/commit/3e5b54136e0aa81a9598fd0bfe906cc0a540b776)

tests/extract/test_dcatus.py
rshewitt: resolved [here](https://github.com/GSA/datagov-harvesting-logic/commit/17a5d09e3c16f4c30de6d327c88e8b7fbb0c205e)

tests/s3/test_s3.py
rshewitt: removed localstack test [here](https://github.com/GSA/datagov-harvesting-logic/pull/10/commits/78e7320f86505509fa384cd30694f556f56d377f)

tests/extract/conftest.py
```diff
@@ -62,9 +62,42 @@ def create_client_config():
...
+
+
 @pytest.fixture
 def create_client(create_client_config):
     """create a boto3.client
     create_client_config (dict)     :   configuration file.
     """
-    return create_s3_client(create_client_config)
+    S3_client = create_s3_client(create_client_config)
+    yield S3_client
```
btylerburton, line 86: is `yield` not effectively the same as a return in python? in JS it's to return a generator and all lines following will be ignored

rshewitt: no, they're not the same. lines proceeding `yield` are executed. this provides an opportunity for cleanup to occur after the fixture object has served its purpose. pytest [documentation](https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended) on yield fixtures.

btylerburton: read the docs and now i get what you were doing. nice!

btylerburton (approved):

nickumia-reisys (approved):
This is not a comment about this PR... but we are working counter-intuitively because there's a lot of rewriting that we'll have to do when we actually start to work on this more actively.  As one example, the amount of `try...except` blocks that do nothing is exhausting for me to look at..

🪨

rshewitt:
@nickumia-reisys if you're not feeling good about my approach let's chat. I want to avoid re-working things. For the `try...except` we can either not include them at all, include them and do nothing, or include them and do something.

PR #14, opened 2023-08-01
Add MVP Initial Flask App

nickumia-reisys:
Related to
- https://github.com/GSA/data.gov/issues/4398

Notes:
- Create basic flask app
- Abstract different elements of the Harvesting API
  - `/source/...` manages harvest source information (`create`, `view`)
  - `/dataset/...` manages dataset information (`create`, `view`)
  - `/job/...` manages harvest job information (`run`, `status`)
  - Note: More routes to be created in the future (these were just added to not have an empty website)
- Dummy `db` variables were created to mock the real data store options.
- The general harvest pipeline structure was mocked to provide a skeleton
- Add `flask` as a dependency
- Create a barebones cloud.gov manifest


controller/source/create.py
```diff
@@ -0,0 +1,22 @@
+import uuid
+from controller.source import bp, db
+from flask import request
+
+
+def valid(x):
+    return x
```
robert-bryson, line 9: Is this intended to be a placeholder for the future?

requirements.txt
```diff
@@ -0,0 +1,4 @@
+# poetry not supported by cloud.gov?
```
robert-bryson, line 1: What is the context on this? Poetry can export to a `requirements.txt` if need be with something like `poetry export --without-hashes --format=requirements.txt > requirements-poetry.txt`.

robert-bryson (approved):
I did a lint pr, hope that's ok! Looks good to me, great work!

controller/source/create.py
nickumia-reisys: Yes.  I'm not sure what a "valid" name/url is for a harvest source.  But we'll have to define those specifications at some point in the future.

requirements.txt
nickumia-reisys: That's a useful command.  We'll definitely need it in the repo somewhere.  We thought that cloud.gov would be able to recognize the poetry file, but apparently it only deals with `requirements.txt`.  So we can generate it with that command.  @Jin-Sun-tts Do you want to run it for this repo and add it to the code someplace?

Jin-Sun-tts (approved):

requirements.txt
Jin-Sun-tts: We may do it next time. Thank you for the useful command.

PR #17, opened 2023-09-11
Harvesting pipeline documentation

nickumia-reisys:
Notes:
- Adds mermaid flowchart documentation of our current harvesting pipelines, plus some new system diagrams too
- Also adds `npm run makeDoc` NPM script to generate new SVGs.

This is not an exhaustive list of all the new docs for Harvester 2.0, but currently this is good to go as is, and can be added to in future.

References:
- https://mermaid.js.org/syntax/flowchart.html


btylerburton (approved):
:shipit:

PR #22, opened 2023-10-24
Refactor repo

rshewitt:
# Pull Request

Related to [#4502](https://github.com/GSA/data.gov/issues/4502)

## About

Removed the flask controller, airflow implementation, sqlalchemy, postgres, s3, and refactored the remaining functionality. We can `git revert` anything we feel needs to be included in the future. This work is intended to remove anything that we don't immediately need within the context of using this repo as a python module in a workflow orchestration system like airflow.  

## PR TASKS

- [x] Any changes to docs?


harvester/utils/json.py
```diff
@@ -1,38 +1,11 @@
...
-
-    if resp.status_code != 200:
-        raise Exception("non-200 status code")
-
-    try:
-        data = resp.json()
-    except json.JSONDecodeError as e:
-        raise Exception(e)
-
-    return data
```
jbrown-xentity, line 38: I don't think we want to remove this code; this will be needed in the final version of this library.

jbrown-xentity (commented):
Close, I would keep one piece of this... Otherwise it's ready!

harvester/utils/json.py
rshewitt: okay I'll add that back in.

rshewitt: added in bea96d7aef757d3168899a1e4e23d882872e7f7a

robert-bryson (approved):

PR #6, opened 2023-04-28
Minor Optimizations

nickumia-reisys:
Related to:
- Review of #5 

Changes:
- I didn't have python 3.9 installed... I needed to specify the python version for poetry to work properly...  https://github.com/python-poetry/poetry/issues/1735#issuecomment-565501611
- Don't limit python version to `^3.9`, but we should support all newer version as well
- Add titles to make the Github Action page easier to read

References:
- https://python-poetry.org/docs/dependency-specification/

PR #23, opened 2023-12-12
init pypi github actions config

robert-bryson:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4544

## About

Related to https://github.com/GSA/datagov-harvester/pull/6

## PR TASKS

- [ ] Any changes to docs?


rshewitt:
catalog proxy on dev needs to be cherry picked back to allow package creation. restarting the actions in a moment should be good.

setup.py
```diff
@@ -0,0 +1,36 @@
...
+    author="Data.gov",
+    author_email="datagovhelp@gsa.gov",
+    url="https://github.com/GSA/datagov-harvesting-logic/",
+    license="Public Domain",
+    packages=find_packages(exclude=["ez_setup", "examples", "tests"]),
+    namespace_packages=["harvester"],
+    include_package_data=True,
+    zip_safe=False,
+    install_requires=[
+        # TODO anything?
```
rshewitt: i'm not familiar with this option so i can't say whether something already accounted for in our pyproject.toml would be better placed here. I think we can probably leave this blank though.

setup.py
rshewitt, line 1: can't we use the pyproject.toml for packaging instead of this?

setup.py
```diff
@@ -0,0 +1,36 @@
...
+    ],
+    setup_requires=["wheel"],
+    entry_points="""
+        # TODO anything?
```
rshewitt: seems like an advanced feature we don't need at least for now

setup.py
btylerburton: thanks for noting that @rshewitt 
can we follow the new standard?
https://www.reddit.com/r/learnpython/comments/yqq551/pyprojecttoml_setupcfg_setuppy_whats_the/

jbrown-xentity (dismissed):
Looks great!

rshewitt (approved):

PR #28, opened 2024-01-03
Feature/compare

rshewitt:
# Pull Request

Related to [#4557](https://github.com/GSA/data.gov/issues/4557)

## About

Add first-pass comparison function including unit and integration tests

## PR TASKS

- [x] Any changes to docs?


README.md
```diff
@@ -5,35 +5,86 @@ transformation, and loading into the data.gov catalog.
...
+- Run tests with the command `poetry run pytest ./tests/load/ckan`
+
 ### Harvester testing
-- These tests are found in `extract`, and `validate`. Some of them rely on services in the `docker-compose.yml`. run using docker `docker compose up -d` and with the command `poetry run pytest --ignore=./tests/load/ckan`. 
+
+- These tests are found in `extract`, and `validate`. Some of them rely on services in the `docker-compose.yml`. Run using docker `docker compose up -d` and with the command `poetry run pytest --ignore=./tests/load/ckan`.
 
 If you followed the instructions for `CKAN load testing` and `Harvester testing` you can simply run `poetry run pytest` to run all tests.
 
+## Comparison
```
btylerburton, line 42: these diffs feel like they belong next to their respective test cases

harvester/load.py
```diff
@@ -93,14 +105,17 @@ def simple_transform(dcatus_catalog):
...
         if isinstance(mapping[k], dict):
             temp = {}
+            to_skip = ["@type"]
             for k2, v2 in v.items():
                 if k2 == "hasEmail":
                     v2 = get_email_from_str(v2)
+                if k2 in to_skip:
+                    continue
                 temp[mapping[k][k2]] = v2
             output = {**output, **temp}
```
btylerburton, line 120: is there any chance we overwrite good values in `**output` with values in `**temp`?

harvester/load.py
btylerburton, line 1: the methods in this file are complex enough to merit some unit test cases.

btylerburton (commented):
overall great progress. a bit more documentation and some unit test cases should be sufficient to merge.

harvester/load.py
rshewitt: at the moment, there's no chance of overwriting because there's no overlap of keys between `output` and the values of `mapping`. unless there's duplicate keys in what ckan expects in a dataset in may not ever happen? i think this is a good thing to be aware of though.

README.md
rshewitt: this would require moving them to 2 places ( unit and integration ) which seems redundant because the datasets with these changes are used in tests in both places. I agree that it makes sense to include these comments next to their tests though.

harvester/load.py
rshewitt: added tests in [commit](https://github.com/GSA/datagov-harvesting-logic/pull/28/commits/c31a32ee2f1cad23c39db718001677e698fcd014)

robert-bryson (approved):

PR #30, opened 2024-01-18
Refactor to class

rshewitt:
# Pull Request

Related to [#4577](https://github.com/GSA/data.gov/issues/4577)

## About

- refactoring harvester functionality and tests as classes
- i'm using the `@property` decorator to make some fields read-only. it serves a similar benefit to `const` in node. 
  - i'm not using the decorator in some cases where the dtype is a dictionary because we'd need to create a custom dict class and overload the `__setitem__` method. is it better to be consistent?
  - the non-dict fields which will be assigned after initialization have setter methods configured to them with basic type checking
- a majority of tests have been removed. this is for a couple reasons
  - the number of tests aren't needed ( e.g. validation and extract ) 
  - some tests weren't added back into this refactor in an effort to get this codebase available to @btylerburton more quickly so he can perform his load test. 
  - furthermore, I simply wanted to get the coverage reasonably high while also minimizing our codebase as a sort-of fresh start.
- nothing is being done with the transform ( i.e no tests or functionality ). we're not transforming anything in our [load test](https://github.com/GSA/data.gov/issues/4578) because our transformation service isn't ready although progress is actively being made so maybe we include it? 
- `RemoteCKAN` is created outside of any dataclass to avoid any potential serialization issue when/if we decide to store an intermediate `HarvestSource` instance in somewhere like `s3`. 
- `Record` instantiation occurs primarily in `harvest_to_id_hash` but only for records that need to be created or updated NOT deleted. for the sake of record-keeping ( no pun-intended ) I instantiated `Record` for deleted records immediately before deleting them on ckan. 
- there's purposefully no try/excepts ( excluding the validation function because i copy/pasted that ) because we don't have a concrete idea of our exception handling. 

## PR TASKS

- [ ] Any changes to docs?


jbrown-xentity:
As someone who's been tracking this progress all week, my suggestion for reviewing this is to look at the code state at the latest commit, not the changes. GitHub doesn't know how to show this many changes, it's messy. But the code is much cleaner.

tests/unit/load/test_ckan_load.py
```diff
@@ -0,0 +1,104 @@
...
+        self.status = "updated"
+
+    def create_mock(self):
+        self.status = "created"
+
+    @patch.object(Record, "create_record", create_mock)
+    @patch.object(Record, "update_record", update_mock)
+    @patch.object(Record, "delete_record", delete_mock)
+    @patch.object(HarvestSource, "get_ckan_records")
+    def test_sync(self, get_ckan_records_mock, dcatus_compare_config, ckan_compare):
```
rshewitt, line 22: i'm not a fan of this many patches but since it's testing against the synchronization ( delete, update, create ) of the records I need to mock all of those methods in addition to mocking the retrieval of ckan records. an alternative to this would be to have harvest_source.compare_data only contain records to either delete, update, or create.

rshewitt: it's a matter of whether we want to have 3 tests ( one for delete, update, create ) or one test which does all of them.

harvester/harvest.py
```diff
@@ -0,0 +1,497 @@
...
+    # making these fields read-only
+    # (these values can be set during initialization though)
+    _title: str
+    _url: str
+    _extract_type: str  # dcatus vs waf
+    _waf_config: dict = field(default_factory=lambda: {})
+    _extra_source_name: str = "harvest_source_name"  # TODO: update this
+    _dataset_schema: dict = field(
+        default_factory=lambda: open_json(
+            ROOT_DIR / "data" / "dcatus" / "schemas" / "dataset.json"
```
btylerburton, line 50: reminder to move that schema dir into harvester dir

harvester/harvest.py
```diff
@@ -0,0 +1,497 @@
...
+                self,
+                i,
+                record,
+                h,
+            )
+
+    def ckan_to_id_hash(self, results: list[dict]) -> None:
+        for record in results:
+            self.ckan_records[record["identifier"]] = dataset_to_hash(
+                eval(
```
btylerburton: why do we need eval here?

harvester/harvest.py
```diff
@@ -0,0 +1,497 @@
...
+        ckan_ids = set(self.ckan_records.keys())
+        same_ids = harvest_ids & ckan_ids
+
+        # since python 3.7 dicts are insertion ordered so deletions will occur first
+        self.compare_data = {"delete": None, "create": None, "update": None}
+        self.compare_data["delete"] = ckan_ids - harvest_ids
+        self.compare_data["create"] = harvest_ids - ckan_ids
+        self.compare_data["update"] = set(
+            [
+                i
```
btylerburton: list comprehension is easier for me to read on a single line.

harvester/harvest.py
```diff
@@ -0,0 +1,497 @@
...
+        return output
+
+    def create_ckan_publisher_hierarchy(self, pub_dict: dict, data: list = []):
+        for k, v in pub_dict.items():
+            if k == "name":
+                data.append(v)
+            if isinstance(v, dict):
+                self.create_ckan_publisher_hierarchy(v, data)
+
+        return " > ".join(data[::-1])
```
btylerburton, line 473: remind me again what `[::-1]` denotes

btylerburton: nvm. https://stackoverflow.com/questions/31633635/what-is-the-meaning-of-inta-1-in-python

what exactly is it doing here then?

rshewitt: it joins the reversed list of strings together by ` > `. for example
```python
test_publisher_dict = {
                "name": "U.S. Commodity Futures Trading Commission",
                "subOrganizationOf": {
                    "name": "U.S. Government"
                }
            }

# after traversing through all the nested dictionaries "data" will look like this...
data = [ "U.S. Commodity Futures Trading Commission", "U.S. Government" ]

" > ".join(data[::-1]) # produces the string "U.S. Government > U.S. Commodity Futures Trading Commission"

data[::-1] # produces [ "U.S. Government", "U.S. Commodity Futures Trading Commission" ]
```

the structure of the publisher dictionary goes from more specific ( agency ) to less specific ( government ) and we want the reverse of that order

slicing syntax is [ start: stop: step ]. a negative step is backwards and not including a start or stop means the entire thing.

harvester/harvest.py
rshewitt: yeah i totally agree. this is a result of the auto-formatting on save i have configured with black in vsc. any other examples where the formatting looks weird is caused by that too. increasing the maximum line length should hopefully fix it?

harvester/harvest.py
rshewitt: this was my shortcut to convert a json str into a dict. i was running into situations where the input json str needed to be scrubbed of backslashes and single quotes. I recall trying to implement something like...
```python
record["dcat_metadata"].replace("\\","").replace("'",'"') 
```
but I still ran into a json decode error when using json.load( str ). I'll go back and get something working without eval

harvester/harvest.py
rshewitt: resolved in d65bed0e328133c937957b5d499cbf5ce0968bb6

jbrown-xentity (approved):
This all looks good to me, reviewed in person.

PR #34, opened 2024-02-06
ckanification utilities

rshewitt:
# Pull Request

Related to [#4599](https://github.com/GSA/data.gov/issues/4599)

## About

add ckan string scrubbing utilities 

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/harvester/ckan_utils.py
+++ b/harvester/ckan_utils.py
@@ -1,0 +1,150 @@
+import re
+
+# all of these are copy/pasted from ckan core
+# https://github.com/ckan/ckan/blob/master/ckan/lib/munge.py
+
+PACKAGE_NAME_MAX_LENGTH = 100
+PACKAGE_NAME_MIN_LENGTH = 2
+
+MAX_TAG_LENGTH = 100
+MIN_TAG_LENGTH = 2
+
+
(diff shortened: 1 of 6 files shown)


harvester/ckan_utils.py
```diff
@@ -0,0 +1,150 @@
...
+        0xF6: "o",
+        0xF8: "o",
+        0xF9: "u",
+        0xFA: "u",
+        0xFB: "u",
+        0xFC: "u",
+        0xFD: "y",
+        0xFE: "th",
+        0xFF: "y",
+        # 0xa1: '!', 0xa2: '{cent}', 0xa3: '{pound}', 0xa4: '{currency}',
```
btylerburton, line 97: is there a reason for commenting these out?

btylerburton (approved):
Looks good!

harvester/ckan_utils.py
rshewitt: no, this was a direct copy/paste from the [source](https://github.com/ckan/ckan/blob/f2550929feed5fbd8f155e71d855aa8d4fb1e1b6/ckan/lib/munge.py#L95-L103)

PR #36, opened 2024-02-26
Create harvest db

Jin-Sun-tts:
# Pull Request

Related to [Create Harvest DB and Table Structure](https://github.com/GSA/data.gov/issues/4612)

## About

-  Defined Table Structure in SQLAlchemy
-  Initialized this structure into a local Docker PostgresSQL container
-  Created interfaces to interact with the DB
-  Created Flask app (https://harvesting-logic.app.cloud.gov/) to test the harvester operations into the DB 
-  Created manifest file for deployment to cloud.gov with PostgresSQL aws-rds micro-sql service.  
-  Added tests with local PostgresSQL container.


Jin-Sun-tts:
moved to datagov-harvest-orchestrator repo

Jin-Sun-tts:
back from datagov-harvest-orchestrator repo, reopen this.

Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Use an official Python runtime as a base image
+FROM python:3.11
```
btylerburton: mismatch with `.python-version`

README.md
```diff
@@ -88,3 +88,59 @@ If you followed the instructions for `CKAN load testing` and `Harvester testing`
...
+Any created service needs to be bound to an app with `cf bind-service <app> <service>`. With the above example, the db can be bound with 
+`cf bind-service harvesting-logic harvesting-logic-db`.
+
+Accessing the service can be done with service keys. They can be created with `cf create-service-keys`, listed with `cf service-keys`, and shown with 
+
+`cf service-key <service-key-name>`.
+
+#### Manually Deploying the Flask Application
+
+1. Ensure you have a `manifest.yml` and `vars.yml` file configured for your Flask application. The `vars.yml` file should include variables such as `FLASK_APP` and database service bindings.
```
btylerburton: would it be helpful to provide a more exhaustive list of vars?

app.py
```diff
@@ -0,0 +1,87 @@
...
+    for rule in app.url_map.iter_rules():
+        if 'static' not in rule.endpoint and 'index' not in rule.endpoint:
+            html += (f"<li>{rule.endpoint} : "
+                     f"<a href='{rule.rule}'>{rule.rule}</a></li><br>")
+    html += "</ul>"
+    return html
+
+@app.route('/add_source', methods=['GET'])
+def add_harvest_source():
+    source=db.add_harvest_source(new_source)
```
btylerburton, line 22: did you want to wire in actual data in this PR?

app.py
```diff
@@ -0,0 +1,87 @@
...
+    return(f"Added new source with ID: {source.id}")
+
+@app.route('/add_job', methods=['GET'])
+def add_harvest_job():
+    source_id = request.args.get('source_id', None)
+    if source_id is None:
+        return 'Please provide source_id: /add_job?source_id=xxx'
+    else:
+        job=db.add_harvest_job(new_job, source_id)
```
btylerburton, line 31: same as above for `new_source`

app.py
```diff
@@ -0,0 +1,87 @@
...
+        return(f"Added new job with ID: {job.id}")
+
+@app.route('/add_error', methods=['GET'])
+def add_harvest_error():
+    job_id = request.args.get('job_id', None)
+    if job_id is None:
+        return 'Please provide job_id: /add_error?job_id=xxx'
+    else:
+        err=db.add_harvest_error(new_error, job_id)
```
btylerburton, line 40: same as above

app.py
```diff
@@ -0,0 +1,87 @@
...
+    result = db.get_all_harvest_jobs()
+    return result
+
+@app.route('/harvest_errors', methods=['GET'])
+def get_all_harvest_errors():
+    result = db.get_all_harvest_errors()
+    return result
+
+@app.route('/harvest_source/<source_id>', methods=['GET'])
+def get_harvest_source(source_id):
```
btylerburton, line 63: is it more pythonic to put the var in the function signature or to get at like above using `requests.get...`? or are those doing two different things?

.profile
```diff
@@ -0,0 +1,22 @@
...
+
+export APP_NAME=$(echo $VCAP_APPLICATION | jq -r '.application_name')
+
+export db_username=$(vcap_get_service aws-rds .credentials.username)
+export db_password=$(vcap_get_service aws-rds .credentials.password)
+export db_host=$(vcap_get_service aws-rds .credentials.host)
+export db_port=$(vcap_get_service aws-rds .credentials.port)
+export db_name=$(vcap_get_service aws-rds .credentials.db_name)
+
+export DATABASE_URI=postgresql://$db_username:$db_password@$db_host:$db_port/$db_name
```
jbrown-xentity: I think all the intermediate steps may be unnecessary; I think you can just grab the URI for the database if you want to. Please double check though in the actual vcap service object.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+# Make port 8080 available to the world outside this container
+EXPOSE 8080
+
+# Define environment variable
+ENV FLASK_APP=app.py
+ENV FLASK_RUN_HOST=0.0.0.0
+ENV FLASK_RUN_PORT=8080
+
+# Run Flask
+CMD ["poetry", "run", "flask", "run"]
```
jbrown-xentity: We can keep this file, but I don't recommend using it. Using the flask app directly on our local machine allows for much better debugging (in my experience)

harvester/database/models.py
```diff
@@ -0,0 +1,58 @@
...
+    id = mapped_column(
+        UUID(as_uuid=True), primary_key=True, server_default=text("gen_random_uuid()")
+    )
+
+class HarvestSource(Base):
+    __tablename__ = 'harvest_source'
+    
+    name = mapped_column(String, nullable=False)
+    notification_emails = mapped_column(ARRAY(String))
+    organization_name = mapped_column(String)
```
btylerburton: change to `organization_id = mapped_column(String, nullable=False)`?

harvester/database/models.py
```diff
@@ -0,0 +1,58 @@
...
+    frequency = mapped_column(String, nullable=False)
+    url = mapped_column(String, nullable=False)
+    schema_type = mapped_column(String, nullable=False)
+    source_type = mapped_column(String, nullable=False)
+    harvest_source_id = mapped_column(String)
+    harvest_source_name = mapped_column(String)
```
btylerburton, line 22: what is difference between `name` & `harvest_source_name`?

app.py
```diff
@@ -0,0 +1,87 @@
...
+        return(f"Added new job with ID: {job.id}")
+
+@app.route('/add_error', methods=['GET'])
+def add_harvest_error():
+    job_id = request.args.get('job_id', None)
+    if job_id is None:
+        return 'Please provide job_id: /add_error?job_id=xxx'
+    else:
+        err=db.add_harvest_error(new_error, job_id)
+        return(f"Added new error with ID: {err.id}")
```
jbrown-xentity, line 41: I don't think this should exist; however that's my opinion and would default to how others feel about this. I think we should never be manually creating errors through the API, it should be happening by the runner directly into the database.

create_cloudgov_service.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Get the current space and trim leading whitespace
+space=$(cf target | grep space | cut -d : -f 2 | xargs)
+
+# Production and staging should use bigger DB instances
+if [ "$space" = "prod" ] || [ "$space" = "staging" ]; then
+    cf service "${app_name}-db"    > /dev/null 2>&1 || cf create-service aws-rds medium-psql-redundant "${app_name}-db" --wait&
+else
+    cf service "${app_name}-db"    > /dev/null 2>&1 || cf create-service aws-rds micro-psql "${app_name}-db" --wait&
+fi
```
jbrown-xentity, line 17: Unless we're going to use these in a github action to auto-deploy the necessary services I don't see this as being useful...

harvester/database/interface.py
```diff
@@ -0,0 +1,79 @@
...
+    def _to_dict(obj):
+        return {c.key: getattr(obj, c.key) 
+                for c in inspect(obj).mapper.column_attrs}
+    
+    def add_harvest_source(self, source_data):
+        new_source = HarvestSource(**source_data)
+        self.db.add(new_source)
+        self.db.commit()
+        self.db.refresh(new_source)
+        return new_source
```
jbrown-xentity, line 27: In the future, it would be nice to just accept a HarvestSource object, and do the manipulation/crosswalk here to load it.

harvester/database/interface.py
```diff
@@ -0,0 +1,79 @@
...
+        self.db.add(new_error)
+        self.db.commit()
+        self.db.refresh(new_error)
+        return new_error
+
+    def get_all_harvest_errors(self):
+        harvest_errors = self.db.query(HarvestError).all()
+        harvest_errors_data = [
+            HarvesterDBInterface._to_dict(err) for err in harvest_errors]
+        return harvest_errors_data
```
jbrown-xentity: That could turn into a dangerous method in the future; I expect this to be millions of records. We should maybe only have a "get_all_harvest_errors_by_job" function to start.

jbrown-xentity (dismissed):
Thanks @Jin-Sun-tts ! Some comments for your and the @GSA/data-gov-team 's awareness/review, some of these are general comments and some should be addressed before merging. I'm going to approve though, in case we need to get this merged and circle back to the improvements later.

tests/database/test_db.py
```diff
@@ -0,0 +1,121 @@
...
+
+    with engine.begin() as connection:
+        connection.execute(text(f"DROP SCHEMA IF EXISTS {TEST_SCHEMA} CASCADE;"))
+
+@pytest.fixture(scope='session')
+def db_interface(db_session):
+    return HarvesterDBInterface(db_session)
+
+def test_add_harvest_source(db_interface):
+    source_data = {'name': 'Test Source',
```
btylerburton, line 40: why not use test data here?

README.md
btylerburton, line 1: I really appreciate the README!

.profile
btylerburton: Pretty sure I remember that being present as well

app.py
btylerburton: 100% agree with that.

README.md
Jin-Sun-tts: Done.

app.py
Jin-Sun-tts: This is just for easy test. Will have the edit option later to allow input actual data.

app.py
Jin-Sun-tts: This is just for easy test. Will have the edit option later to allow input actual data.

app.py
Jin-Sun-tts: This is just for easy test. Will have the edit option later to allow input actual data.

tests/database/test_db.py
Jin-Sun-tts: test data file will be removed. That file is just for initial quick test.

harvester/database/models.py
Jin-Sun-tts: good point. I think they are the same thing.  We added the `harvest_source_id` and `harvest_source_name` which are from ckan database.  We should only keep one. Will do the modification later along with other database changes.

harvester/database/models.py
Jin-Sun-tts: done.

Dockerfile
Jin-Sun-tts: fixed.

create_cloudgov_service.sh
Jin-Sun-tts: removed. It is not be used anyway. Thanks.

app.py
Jin-Sun-tts: agree. I just list here for initial manually testing database. Will remove it later.

Dockerfile
Jin-Sun-tts: sure, Done.

harvester/database/interface.py
rshewitt: paginate the results? include a page count and start count? we can do this with the ckan api

.profile
Jin-Sun-tts: Yes, you are right. Done.

harvester/database/interface.py
Jin-Sun-tts: agree. Will do.

harvester/database/interface.py
Jin-Sun-tts: True. I just list here for initial test database. already changed it.

harvester/database/models.py
rshewitt: agree 100%. with the addition of the `organization` table we should also add a FK.

app.py
Jin-Sun-tts: Using path parameter like above aligns more with REST API conventions.  
The query parameters are for manual record entry during testing phases.   
Moving forward, we will accept an object instead of id for the creation of  source, job, error etc.

harvester/database/interface.py
Jin-Sun-tts: @rshewitt good point.  If we choose to implement the function/method, we might consider different approaches for its display. We should have a discussion to determine its necessity. For the time being, I've made modifications to display only by job id.

FuhuXia (approved):

PR #38, opened 2024-02-27
update gh action

FuhuXia:


jbrown-xentity (approved):
Good improvement, linting and tests take about 2-2.5 minutes, seems a reasonable time to wait to make sure it's a good deployment.

PR #42, opened 2024-03-12
Added routes for each space

Jin-Sun-tts:
added routes for each space to fix the deploy error.


FuhuXia:
How do internal routes get used?

Jin-Sun-tts:
I just removed it as we do not use it for now. We may add it back later when there is a need. @FuhuXia

FuhuXia (approved):

PR #43, opened 2024-03-12
Exception handling

rshewitt:
# Pull Request

Related to [#4620](https://github.com/GSA/data.gov/issues/4620)

## About

- add exception handling using custom exception classes. db error creation and critical/error logs occur within the exception classes as a way to centralize where these things occur

## PR TASKS

- [ ] Any changes to docs?


harvester/logger_config.py
```diff
@@ -0,0 +1,25 @@
...
+        "standard": {
+            "format": (
+                "[%(asctime)s] %(levelname)s "
+                "[%(name)s.%(funcName)s:%(lineno)d] %(message)s"
+            )
+        },
+    },
+    "handlers": {
+        "console": {
+            "level": "INFO",
```
jbrown-xentity, line 13: Someday we may change this up to "WARN", however for today more logging is better.

jbrown-xentity (approved):
LGTM! I really like how we're using exceptions to define what happens (logging and DB error creation) instead of copy pasting over and over, or using some generic function.

PR #44, opened 2024-03-12
fix the vars file name

Jin-Sun-tts:
Adjusted variable file names based on varying spaces.


FuhuXia (approved):
it seems about time to make it a template to avoid future edits like this.

PR #45, opened 2024-03-14
Create `update` route

robert-bryson:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4641

## About

Added an update route to the app and tests. Triggering an update to a `<source-id>` is done via a `PUT` request to `/harvests/<source-id>`. 

More work should be done on error handling, SQL sanitization, and things like that, but those are larger conversations for the team.

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Python Lint
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - uses: chartboost/ruff-action@v1
 
   test:
@@ -21,7 +21,7 @@ jobs:
     name: Pytests
     steps:
       - name: Check out the code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           fetch-depth: 1
 
@@ -61,7 +61,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Install Poetry
         uses: abatilo/actions-poetry@v2
         with:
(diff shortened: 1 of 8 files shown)


app.py
```diff
@@ -83,6 +83,17 @@ def get_harvest_error(error_id):
     except Exception:
         return "provide error_id"
 
+
+@app.route("/harvests/<string:source_id>", methods=["PUT"])
+def update_harvest_source():
+    source_id = request.args.get("source_id", None)
```
FuhuXia: `source_id` is available as a function argument, why it is getting from querystring here?

rshewitt: on a different note I like the use of variables in the route itself like you have here instead of parsing them via `requests.get( "example" )` from the query string. i think it's more descriptive of what's expected vs just `/harvests`

robert-bryson: Sorry, good catch. That is how it was in the `harvesting-orchestrator` so I just copied.

FuhuXia:
This PR can merge after https://github.com/GSA/datagov-harvesting-logic/pull/46, where files and routes are restructured.

robert-bryson:
> This PR can merge after #46, where files and routes are restructured.

@FuhuXia, I rebased. I think this looks good to go now.

FuhuXia:
I dont see any changes on how to get `source_id` from route `/harvests/<string:source_id>` ?

robert-bryson:
> I dont see any changes on how to get `source_id` from route `/harvests/<string:source_id>` ?

Should be fixed. I think I blew away the change with clumsily rebasing.

robert-bryson:
Also, looking more closely, I don't think our test command has been updated for the new changes: https://github.com/GSA/datagov-harvesting-logic/blob/main/.github/workflows/commit.yml#L47

FuhuXia (approved):

PR #46, opened 2024-03-14
Harvest db update

Jin-Sun-tts:
# Pull Request

Related to (https://github.com/GSA/data.gov/issues/4647)

- Switched to Flask-Alchemy 
- Reorganized the Flask app into separate directory. 
- Added Flask-Migrate for managing database DDL changes.
- Added table indexes.


app/models.py
```diff
@@ -0,0 +1,78 @@
...
+                                  db.ForeignKey('harvest_source.id'),
+                                  nullable=False)
+    date_created = db.Column(db.DateTime)
+    date_finished = db.Column(db.DateTime)
+    records_added = db.Column(db.Integer)
+    records_updated = db.Column(db.Integer)
+    records_deleted = db.Column(db.Integer)
+    records_errored = db.Column(db.Integer)
+    records_ignored = db.Column(db.Integer)
+    errors = db.relationship('HarvestError', backref='job', lazy=True)
```
FuhuXia: It feels like we are missing some indexes. We will list jobs by time, by finished status quite often. There is no index for them. We may use integer (not uuid) as primary key, it is sortable, so that we can avoid using indexes for dates fields?

app/models.py
```diff
@@ -0,0 +1,78 @@
...
+    id = db.Column(UUID(as_uuid=True), primary_key=True,
+                   server_default=text("gen_random_uuid()"))
+
+
+class HarvestSource(Base):
+    __tablename__ = 'harvest_source'
+    
+    name = db.Column(db.String, nullable=False)
+    notification_emails = db.Column(ARRAY(db.String))
+    organization_id = db.Column(db.String, nullable=False)
```
FuhuXia: add index?

app/models.py
```diff
@@ -0,0 +1,78 @@
...
+    frequency = db.Column(db.String, nullable=False)
+    url = db.Column(db.String, nullable=False)
```
FuhuXia: add unique index?

app/models.py
```diff
@@ -0,0 +1,78 @@
...
+    schema_type = db.Column(db.String, nullable=False)
+    source_type = db.Column(db.String, nullable=False)
+    harvest_source_id = db.Column(db.String)
+    harvest_source_name = db.Column(db.String)
```
FuhuXia: what is the difference between id/harvest_source_id and name/harvest_source_name ?

app/models.py
```diff
@@ -0,0 +1,78 @@
...
+    type = db.Column(db.String(), nullable=False)
+    source_metadata = db.Column(db.String(), nullable=True)
+    name = db.Column(db.String(), nullable=False)
+    owner_org = db.Column(UUID(as_uuid=True),
+                          db.ForeignKey('organization.id'),
+                          nullable=False)
+    author = db.Column(db.String(), nullable=True)
+    author_email = db.Column(db.String(), nullable=True)
+    maintainer = db.Column(db.String(), nullable=True)
+    maintainer_email = db.Column(db.String(), nullable=True)
```
FuhuXia: i see the point to have ckan_id (indexed) and source_metadata, no point to have other fields.

requirements.txt
```diff
@@ -1,3 +1,8 @@
...
-# run this command to generate the file:
-# poetry export -f requirements.txt --output requirements.txt --without-hashes
\ No newline at end of file
+
+flask-migrate==4.0.7 
+flask-sqlalchemy==3.1.1 
+flask==3.0.2 
+psycopg2-binary==2.9.9 
+pytest==7.4.4 
+python-dotenv==1.0.1
```
FuhuXia: Is this file supposed to be auto generated for cf push purpose, no need to keep an offline copy?

app/models.py
Jin-Sun-tts: added index for date_created and status.

app/models.py
Jin-Sun-tts: added index.

app/models.py
Jin-Sun-tts: added unique id for url.

app/models.py
Jin-Sun-tts: @btylerburton we had this discussion one time in the meeting, and added this two columns https://github.com/GSA/data.gov/issues/4612#issuecomment-1960391070, do we still need them?

app/models.py
Jin-Sun-tts: @btylerburton @rshewitt @jbrown-xentity we may need group discussion for the purpose of this table and which fields we need to keep.

requirements.txt
Jin-Sun-tts: you are right, just removed it.

app/models.py
Jin-Sun-tts: after discussion, we decide to remove these two columns.

app/models.py
Jin-Sun-tts: after discussion, removed extra fields for now.

FuhuXia (approved):

PR #54, opened 2024-04-23
Fix cf integration tests

rshewitt:
# Pull Request

Related to [4706](https://github.com/GSA/data.gov/issues/4706)

## PR TASKS

- [x] Any changes to docs?


rshewitt:
Screenshot 2024-04-23 at 7 21 58 AM

rshewitt:
I ended up consolidating some of the tests into a single integration test because I need the `guid` of the task assigned by cf to get the task, delete the task, or read logs from the task. I originally passed the task name which I controlled in my fixture but that was proving unreliable.

rshewitt:
I also ended up removing the example cf credential vars from the .env file to avoid having env vars be in more than 1 place ( secret vs .env ).  i updated the readme with local integration testing info.

btylerburton (approved):
Nice fixes

PR #61, opened 2024-05-03
Refactor

rshewitt:
# Pull Request

Related to [#4720](https://github.com/GSA/data.gov/issues/4720), 
- [#4718](https://github.com/GSA/data.gov/issues/4718)
  - [function](https://github.com/GSA/datagov-harvesting-logic/blob/57ecdcd762218b6b44ece2fe7cf38fc8e85c7540/harvester/harvest.py#L121)
  - [test](https://github.com/GSA/datagov-harvesting-logic/blob/57ecdcd762218b6b44ece2fe7cf38fc8e85c7540/tests/unit/exception/test_exception_handling.py#L36)
- [#4712](https://github.com/GSA/data.gov/issues/4712)
  - [function](https://github.com/GSA/datagov-harvesting-logic/blob/57ecdcd762218b6b44ece2fe7cf38fc8e85c7540/harvester/harvest.py#L161)
  - [compare test](https://github.com/GSA/datagov-harvesting-logic/blob/57ecdcd762218b6b44ece2fe7cf38fc8e85c7540/tests/unit/compare/test_compare.py#L20) which involves pulling harvest records from the harvest db. could also include exception test or a simple read. 
- [#4668](https://github.com/GSA/data.gov/issues/4668)
  - [compare test](https://github.com/GSA/datagov-harvesting-logic/blob/57ecdcd762218b6b44ece2fe7cf38fc8e85c7540/tests/unit/compare/test_compare.py#L20)

This PR relates to multiple tickets because I figured while I was refactoring our code I may as well complete some of these other things

Hopefully something like this won't be needed again in the near future

## About
- reorganize file structure
  - moved all database stuff into a separate folder
  - moved example data to root in "example_data"
  - moved json schema to root in "schemas"
  - moved nginx config to root
- tests
  - consolidate tests to classes
  - moved database tests to unit test folder
  - consolidated extract tests into one module
  - update tests to work with new harvest source class 
  - add necessary record insertions to satisfy FK constrains in all relevant tests
  - removed integration tests for compare and extract
    - these tests originally read from ckan but that's changed to reading from the harvesting db. the integration test would be reading from a database living in cloud.gov but I don't think we're there yet? we can make those tests later. 
- harvest.py
  - add function for reading harvest source information from job id
  - replace ckan read functionality with harvest db read
  - isolate some utility functions out of harvest.py into the respective util module. 
- general clean up
  - removed mdtranslator and localstack services in docker compose
  - remove s3 handler class and associated tests
  - remove boto3 as dependency

## PR TASKS

- [ ] Any changes to docs?


harvester/ckan_utils.py
```diff
@@ -148,3 +148,147 @@ def munge_tag(tag: str) -> str:
...
+            output.append(
+                {
+                    "key": "publisher_hierarchy",
+                    "value": create_ckan_publisher_hierarchy(val, []),
+                }
+            )
+
+        else:
+            if isinstance(val, list):  # TODO: confirm this is what we want.
+                val = val[0]
```
btylerburton, line 182: why do we want to truncate that if it's a list?

harvester/ckan_utils.py
```diff
@@ -148,3 +148,147 @@ def munge_tag(tag: str) -> str:
...
+    # )
+
+    # output.append(
+    #     {
+    #         "key": self.harvest_source.extra_source_name,
+    #         "value": self.harvest_source.title,
+    #     }
+    # )
+
+    output.append({"key": "identifier", "value": metadata["identifier"]})
```
btylerburton, line 201: do you want to strike 'identifier' from the earlier list since we're appending it here?

harvester/ckan_utils.py
```diff
@@ -148,3 +148,147 @@ def munge_tag(tag: str) -> str:
...
+    for k, v in metadata.items():
+        if k not in mapping:
+            continue
+        if isinstance(mapping[k], dict):
+            temp = {}
+            to_skip = ["@type"]
+            for k2, v2 in v.items():
+                if k2 == "hasEmail":
+                    v2 = get_email_from_str(v2)
+                if k2 in to_skip:
```
btylerburton, line 275: what exactly is going on here? might be worth a comment to explain this control flow.

jbrown-xentity (approved):
All relevant tests seem to be ported, and this looks ready to go.

rshewitt:
@btylerburton 3 of the tickets ( read records from harvest db, use harvest records in compare, and implement preprocess function for pulling harvest source info via job id ) are found in this [commit](https://github.com/GSA/datagov-harvesting-logic/pull/61/commits/ab680c5fcb0847515f308203b919af60c24b862e). unfortunately, there was also some refactoring which occurred in that commit as well. i updated the description of the pr with links to what updates were made as they pertain to each ticket

btylerburton (commented):
A few post merge comments I had stored...

harvester/ckan_utils.py
rshewitt: sometimes "bureauCode" comes as a list and at the time of making this logic I assumed ckan would only accept 1 value as the bureau code so this grabs the first element and uses it

harvester/ckan_utils.py
rshewitt: ah yeah sure

harvester/ckan_utils.py
rshewitt: all those functions contribute to ckanifying a dcatus dataset

PR #64, opened 2024-05-07
refactor bulk add records to follow ORM addall pattern

### LINKED ISSUE #4736: Refactor DB interface add_harvest_records method
## User Story

In order to keep track of the correlation between the harvest records db table and the records in CKAN, datagovteam needs to add the sqlalchemy generated UUID to CKAN metadata. To accomplish that goal, we need to generate the HarvestRecord ORM object prior to inserting into the DB.

Related to:
- https://github.com/GSA/data.gov/issues/4667

## Acceptance Criteria

- [ ] GIVEN I have a series of records output by the harvest runner's compare step
  THEN I want to create the HarvestRecords ORM object in a for loop, saving the newly generated UUID to the accompanying CKAN record, prior to inerting the objects into the DB.

## Background

Work done in the prior [PR](https://github.com/GSA/data.gov/issues/4667) would not have allowed for the generation of UUID's individually. Any performance gains by bulk records insertion would be mititagated by a second for loop to recover that UUID prior to posting to CKAN.

## Sketch

- [ ] After compare step, any records that should be created/updated/deleted should be cast into HarvestRecords ORM objects
- [ ] In the same for loop, we take the return object of that ORM generation and add the `id` to the outgoing CKAN record
- [ ] Push the newly generated list of ORM records into DB in a single commmit using `ds.add_all()`

btylerburton:
# Pull Request

Resolves https://github.com/GSA/data.gov/issues/4736

## About

Need a bulk method that allows for creating of foreign key relationships. 

Validated based on the tests here: https://docs.sqlalchemy.org/en/13/faq/performance.html#i-m-inserting-400-000-rows-with-the-orm-and-it-s-really-slow

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/database/interface.py
+++ b/database/interface.py
@@ -232,7 +232,11 @@ class HarvesterDBInterface:
         :raises Exception: if the records_data contains records with errors
         """
         try:
-            self.db.bulk_insert_mappings(HarvestRecord, records_data)
+            for i, record_data in enumerate(records_data):
+                new_record = HarvestRecord(**record_data)
+                self.db.add(new_record)
+                if i % 1000 == 0:
+                    self.db.flush()
             self.db.commit()
             return True
         except Exception as e:
--- a/database/interface.py
+++ b/database/interface.py
@@ -1,4 +1,5 @@
 import os
+import uuid
 from sqlalchemy import create_engine, inspect, or_
 from sqlalchemy.exc import NoResultFound
 from sqlalchemy.orm import scoped_session, sessionmaker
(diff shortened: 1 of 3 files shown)


jbrown-xentity (dismissed):

database/interface.py
```diff
@@ -232,7 +232,11 @@ def add_harvest_records(self, records_data: list) -> bool:
         :raises Exception: if the records_data contains records with errors
         """
         try:
-            self.db.bulk_insert_mappings(HarvestRecord, records_data)
+            for i, record_data in enumerate(records_data):
+                new_record = HarvestRecord(**record_data)
+                self.db.add(new_record)
+                if i % 1000 == 0:
+                    self.db.flush()
```
jbrown-xentity: I don't fully understand how this is better. I thought we wanted to track the ID in code, or insert individually into the DB and get back the ID such that we didn't have to make a lookup call later.
This seems to not do the bulk insert, but I don't see how it adds/tracks the generated ID through the rest of the pipeline.

jbrown-xentity (commented):
Please let me know what I'm misunderstanding!

database/interface.py
rshewitt: you could update the return to a dict like `{ <dataset id>: <new_record.id> }` then pass it along to ckan

btylerburton: Glad I had to update the branch and we got this comment, James! I was solely focused on the issue of the foreign keys, and totally forgot that we wanted to export the id's for use in CKAN.

What is best form for that, @rshewitt? 

- Attach `id` to `record_data` and return the list of dicts
- Cast newly created HarvestRecords to a list of dicts
- A single dict in the form you mentioned above?

All are easy enough.

rshewitt: i think returning a dictionary would be more efficient than a list in this circumstance. using a list would require a loop to find the right record but a dict with the format of `{ <dataset id>: <new_record.id> }` you can derive the new_record.id instantly ( as far as i know dict lookup is O(1) time )

btylerburton: Sounds good. The only thing I don't like is our `add_harvest_record` method returns a different signature, but maybe we can think more about that as a simple lookup table is definitely the right way to go in this case.

rshewitt (approved):

PR #66, opened 2024-05-09
sqlite foreign key & db fixes

btylerburton:
- Adds foreign key constraints to sqlite. 
- fixes tests; 
- adds db.create_all to flask app init; 
- fixes dependency update script; 
- runs update;

## About

Issue noticed after `make clean` that tables were not getting created on app init. Also noticed that foreign key constraints were not respected in tests.

## PR TASKS

- [ ] Any changes to docs?


app/__init__.py
```diff
@@ -34,6 +34,7 @@ def create_app(testing=False):
         register_routes(app)
 
     with app.app_context():
+        db.create_all()
```
rshewitt, line 37: eventually the tables would already exist right? or does this create them if they don't already?

btylerburton: If I start and stop the app they do not get reinitialized, but I also had the same question. I was unsure needed to be wrapped in a conditional, though from what I see in examples online, it does not.

ex. https://stackoverflow.com/questions/20744277/sqlalchemy-create-all-does-not-create-tables

rshewitt (approved):

btylerburton:
based on this failed action run there is still some work to be done on actions setup with env vars. going to merge this anyway though as it fixes the db issues

https://github.com/GSA/datagov-harvesting-logic/actions/runs/9022796421/job/24793348195

app/__init__.py
btylerburton: Funny. All I had to do was look at the official docs... 😸 
https://flask-sqlalchemy.palletsprojects.com/en/3.1.x/quickstart/#create-the-tables

PR #67, opened 2024-05-10
Fix ci deploy

btylerburton:
## About

Fixes issues with cloud.gov deploy

## PR TASKS

- [ ] Any changes to docs?
- [X] Bumped version number in [pyproject.toml](https://github.com/GSA/datagov-harvesting-logic/blob/8078c5b9f015ca7fb8a142e4e4b2e22ad2fd49b1/pyproject.toml#L3) (also checked on [PyPi](https://pypi.org/project/ckanext-geodatagov/#history)).


.github/workflows/deploy.yml
```diff
@@ -33,11 +30,6 @@
             cat requirements.txt
         - name: deploy DHL
           uses: cloud-gov/cg-cli-tools@main
-          env:
-              CF_API_URL: https://api.fr.cloud.gov
-              CF_SERVICE_USER: ${{secrets.CF_SERVICE_USER}}
-              CF_SERVICE_AUTH: ${{secrets.CF_SERVICE_AUTH}}
-              LM_RUNNER_APP_GUID: ${{secrets.LM_RUNNER_APP_GUID}}
```
jbrown-xentity, line 40: Any detail about why we can remove these without impact?

jbrown-xentity (approved):
LGTM

.github/workflows/deploy.yml
btylerburton: That was my mistake. I added them in my last DB fix PR and didn't have time to clean them up. The correct place to put them was in `.profile`

PR #68, opened 2024-05-14
Update db tables

rshewitt:
# Pull Request

Related to [#4731](https://github.com/GSA/data.gov/issues/4731)

## About

update database tables to reflect what's in the definition document. remove use of sqllite in testing environment because it doesn't support arrays as a dtype which we need for notification_emails in the harvest source table

## PR TASKS

- [ ] Any changes to docs?


app/__init__.py
```diff
@@ -14,15 +14,11 @@
...
-    else:
-        app.config["SQLALCHEMY_DATABASE_URI"] = os.getenv("DATABASE_URI")
-        app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
-        app.config["SECRET_KEY"] = os.urandom(16)
-        Bootstrap(app)
+    app.config["TESTING"] = testing
+    app.config["SQLALCHEMY_DATABASE_URI"] = os.getenv("DATABASE_URI")
+    app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
+    app.config["SECRET_KEY"] = os.urandom(16)
+    Bootstrap(app)
```
FuhuXia: Do these configs apply to environments other than testing?

rshewitt: i figure SQLALCHEMY_DATABASE_URI/DATABASE_URI would apply to other environments. everything else can be moved to the [fixture](https://github.com/GSA/datagov-harvesting-logic/blob/03f52c656e0feeff89a9b47ea42fac1b1c3e98d1/tests/conftest.py#L20)

rshewitt: is `SECRET_KEY` something we want to use?

FuhuXia: `SECRET_KEY` can anything for the local environment; in cloud.gov it should be a cf env variable.

btylerburton: set `app.config["TESTING"] = True` in conftest.py and then do conditional logic here based on that.

app/__init__.py
```diff
@@ -14,15 +14,11 @@
...
-    if testing:
-        app.config["TESTING"] = True
-        app.config["SQLALCHEMY_DATABASE_URI"] = "sqlite:///:memory:"
-        app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
-    else:
-        app.config["SQLALCHEMY_DATABASE_URI"] = os.getenv("DATABASE_URI")
-        app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
-        app.config["SECRET_KEY"] = os.urandom(16)
-        Bootstrap(app)
+    app.config["TESTING"] = testing
```
btylerburton: we should move this to `conftest.py`

btylerburton (changes requested):
this is good to after you update the `app.config["TESTING"]` line

btylerburton:
Compare is telling me maybe we should just merge this one as it's got the superset of changes

https://github.com/GSA/datagov-harvesting-logic/compare/separate-database...update-db-tables

btylerburton (dismissed):
looks good

btylerburton:
same as other PR, we should get @Jin-Sun-tts or @FuhuXia to review so we get fresh eyes.

jbrown-xentity (approved):
This looks like it's contained in #72, but looks good!

PR #69, opened 2024-05-14
make flask and harvester apps separate containers

rshewitt:
# Pull Request

Related to [#4750](https://github.com/GSA/data.gov/issues/4750)

## About
- separate flask and harvester apps and dedicating 1 postgres instance to each

## PR TASKS

- [ ] Any changes to docs?


jbrown-xentity:
I think this will have the same impact as before; there is no overwriting of the `.env` file and spinning up the app would use the same DB as integration testing. I think we want a makefile command, something like `integration_test`, that would do the following:

1. spin up docker with a specific app name, specifying the port using [-e in the command line](https://docs.docker.com/compose/environment-variables/set-environment-variables/#set-environment-variables-with-docker-compose-run---env) to be something other than `5432` (say `5433`)
2. run the python integration tests, somehow passing the DB connection information to be 5433 instead of 5432 (maybe we just default the integration tests to know what the port is?)
3. turn off docker app that was spun up

We can then use this in integrated github actions without needing a separate command (just use the pre-defined make command).

rshewitt:
Screenshot 2024-05-14 at 4 33 42 PM

rshewitt:
Screenshot 2024-05-14 at 9 59 31 PM

this is supposed to show the result of adding a record into the harvester app db (5433) and how it doesn't appear in the flask app db (5432)

rshewitt:
Screenshot 2024-05-15 at 8 32 09 AM

separate volumes per db

btylerburton (approved):
LGTM but I'd like someone who didn't touch this code to review. 

@Jin-Sun-tts or @FuhuXia can you take a look?

jbrown-xentity (approved):
Looks great!

PR #70, opened 2024-05-15
James separate db

jbrown-xentity:
# Pull Request

## About

Separates out testing from app development and user testing.

Separates out concerns between integration tests and unit tests.

## PR TASKS

- [ ] Any changes to docs?


rshewitt:
the checks needs to be updated. this can be merged without a review. oh wait i didn't know this was going into mine and not main. i'm cool with leaving the check as-is

btylerburton:
Tests pass in CLI but hang in vscode. I like using VSCODE to debug so I want to pursue some hybrid of my work and this stuff here. Fine to merge this into the parent branch, but we should hold off on merging to main until we solve that problem.

PR #74, opened 2024-05-22
adds db healthcheck to prevent app starting before db is ready

btylerburton:
# Pull Request

Related to:
  - ongoing db headaches
  - restoring my sanity

## About

One issue with moving to postgres is the DB now takes a bit of time to initialize and sometimes the app crashes. This healthcheck prevents that.

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,8 +47,9 @@ down: ## Tears down the flask and harvester containers
 up-debug: ## Sets up local docker environment
 	docker compose -f docker-compose.yml -f docker-compose_debug.yml up -d
 
-clean: down ## Cleans docker images
+clean: ## Cleans docker images
 	docker compose down -v --remove-orphans
+	docker compose -p harvest-app down -v --remove-orphans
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -47,8 +47,9 @@ down: ## Tears down the flask and harvester containers
 up-debug: ## Sets up local docker environment
 	docker compose -f docker-compose.yml -f docker-compose_debug.yml up -d
 
-clean: down ## Cleans docker images
+clean: ## Cleans docker images
```
rshewitt, line 50: clean and down do similar things. could we just have clean?

btylerburton: similar but not the same...

down takes the containers down
clean takes them down and deletes the data volumes associated with them.

rshewitt: ah gotcha cool ok

rshewitt (approved):

PR #75, opened 2024-05-23
Update db after validate

### LINKED ISSUE #4747: Update DB After Validate
## User Story

In order to keep an accurate count of the state of our datasets, datagovteam wants to update the Harvesting database with the results of the Validate operation.

## Acceptance Criteria

- [ ] GIVEN a record needs to validated to conform to DCAT-US spec
THEN that dataset should be passed along to the sync operation

- [ ] GIVEN a record needs to validated 
AND we should create a new error in the harvest_error table
AND that dataset should be added to a list to be marked as "status: error" in the harvest_records db

## Background

## Sketch

- [x] Loop over records to be deleted
- [x] Create an error if necessary at runtime
- [x] Log the result of the validate operation

### LINKED ISSUE #4764: Improve db session management
## User Story

In order to simplify db session management, datagov wants to implement a single session for a given environment ( i.e. testing and deployment ) per application ( flask and harvest runner )

## Acceptance Criteria

- [ ] GIVEN the harvesting logic repo  \
  WHEN a db session is expected for either testing or deployment for a given app \
  THEN 1 session is created and used

## Background
- need to organize our `HarvestDBInterface` instances. there's multiple instances throughout the repo which means there's multiple separate db sessions. depending on which session is used the db will return back different results

## Sketch
- create a session manager class 
- or create a db interface instance in the __init__ file for harvest.py which will be used as the default in `HarvestSource`. pass a difference db interface for testing. in this case, there's 2 session types: testing and deployment.

btylerburton:
# Pull Request

Resolves https://github.com/GSA/data.gov/issues/4747
Resolves https://github.com/GSA/data.gov/issues/4764

## About

Allows us to patch the db_interface by outputting it from a common exported instance in `/harvest/__init__`.

Adds global patch to conftest to reduce testing boilerplate.

Note: this moves attaching of the db_interface to Harvest Source to the __post_init__ method due to the particularities of the python dataclass

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/harvester/exceptions.py
+++ b/harvester/exceptions.py
@@ -43,33 +43,36 @@ class CompareException(HarvestCriticalException):
 
 # non-critical exceptions
 class HarvestNonCriticalException(Exception):
-    def __init__(self, msg, harvest_job_id, title):
-        super().__init__(msg, harvest_job_id, title)
+    def __init__(self, msg, harvest_job_id, record_id):
+        super().__init__(msg, harvest_job_id, record_id)
 
-        self.title = title
         self.msg = msg
         self.harvest_job_id = harvest_job_id
         self.severity = "ERROR"
         self.type = "record"
+        self.harvest_record_id = record_id
 
         self.db_interface = HarvesterDBInterface()
         self.logger = logging.getLogger("harvest_runner")
 
-        error_data = {
+        self.error_data = {
             "harvest_job_id": self.harvest_job_id,
             "message": self.msg,
             "severity": self.severity,
             "type": self.type,
             "date_created": datetime.utcnow(),
-            "harvest_record_id": self.title,  # to-do
+            "harvest_record_id": record_id,  # to-do
         }
 
-        self.db_interface.add_harvest_error(error_data)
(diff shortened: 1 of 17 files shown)


harvester/exceptions.py
```diff
@@ -64,15 +64,15 @@ def __init__(self, msg, harvest_job_id, record_id):
...
+        self.db_interface.add_harvest_error(error_data)
         self.logger.error(self.msg, exc_info=True)
 
 
 class ValidationException(HarvestNonCriticalException):
     def __init__(self, msg, harvest_job_id, record_id):
         super().__init__(msg, harvest_job_id, record_id)
 
-        # self.db_interface.update_harvest_record(record_id, {"status": "error"})
+        self.db_interface.update_harvest_record(record_id, {"status": "error"})
```
rshewitt: don't we want to update the harvest record for any non-critical error? if that's the case then we can move the record update into `HarvestNonCriticalException`

btylerburton: Good point! I think so.

rshewitt (approved):

PR #76, opened 2024-05-24
Writes harvest error to db when harvester fails to delete a record;

### LINKED ISSUE #4746: Update DB After CKAN Delete
## User Story

In order to keep an accurate count of the state of our datasets, datagovteam wants to update the Harvesting database with the results of the CKAN delete operation.

## Acceptance Criteria

- [ ] GIVEN a record is marked to be deleted from CKAN Solr
  THEN we should log that success
  AND we should add mark that dataset as `"state: success"` in the `harvest_records` db.

- [ ] GIVEN a record is marked to be deleted from CKAN Solr
  AND we should create a new error in the `harvest_error` table
  AND we should mark that dataset `"state: error"` in the `harvest_records` db

## Background

## Sketch

- [ ] Loop over records to be deleted
- [ ] Log the result of the delete operation
- [ ] Create an error if necessary at runtime
- [ ] Push the dataset into a failed_delete list if it fails to delete

btylerburton:
adds new get_harvest_errors_by_record_id method
updates exceptions to map type to error class
fixes deprecation issue with datetime;

# Pull Request

Resolves https://github.com/GSA/data.gov/issues/4746

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/harvester/exceptions.py
+++ b/harvester/exceptions.py
@@ -1,5 +1,5 @@
 import logging
-from datetime import datetime
+from datetime import datetime, timezone
 
 from . import HarvesterDBInterface, db_interface
 
@@ -12,7 +12,6 @@ class HarvestCriticalException(Exception):
         self.msg = msg
         self.harvest_job_id = harvest_job_id
         self.severity = "CRITICAL"
-        self.type = "job"
 
         self.db_interface: HarvesterDBInterface = db_interface
         self.logger = logging.getLogger("harvest_runner")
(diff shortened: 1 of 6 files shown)


harvester/exceptions.py
```diff
@@ -21,8 +20,8 @@ def __init__(self, msg, harvest_job_id):
             "harvest_job_id": self.harvest_job_id,
             "message": self.msg,
             "severity": self.severity,
-            "type": self.type,
-            "date_created": datetime.utcnow(),
+            "type": self.__class__.__name__,
```
rshewitt, line 23: good idea

harvester/exceptions.py
```diff
@@ -21,8 +20,8 @@ def __init__(self, msg, harvest_job_id):
...
+            "date_created": datetime.now(timezone.utc),
```
rshewitt, line 24: i didn't know if we needed the time zone. good addition

rshewitt (approved):

PR #77, opened 2024-05-28
Ensures all tests are mocked and no http requests are made

### LINKED ISSUE #4767: mock unit & integration tests to avoid network calls
## User Story

In order to ensure speed and consistency of test runs, datagovteam wants to ensure that all unit & integration tests aren't making live network calls

## Acceptance Criteria

- [ ] GIVEN a unit or integration test
  WHEN it is run without a connection to the internet
  THEN it should behave exactly as it does with a connection

## Background

None. This makes our app boundary more explicit.

## Sketch

- [ ] Bucket specs into the requisite directories: unit, integration, functional
- [ ] Exclude functional specs from normal pytest discovery and `make test`
- [ ] Allow `make test-ci` to run functional specs
- [ ] Ensure that tests pass with internet disconnected

btylerburton:
## Pull request

Resolves https://github.com/GSA/data.gov/issues/4767

- makes sure all unit and intergration tests are not making network calls
- moves single func spec into a new dir and adds it to be run during `make test-ci`

## About

Some tests were making live calls and we don't want that

## PR TASKS

- [ ] Any changes to docs?


tests/conftest.py
```diff
@@ -22,6 +22,15 @@
 
 HARVEST_SOURCE_URL = os.getenv("HARVEST_SOURCE_URL")
 
+# ignore tests in the functional dir
+collect_ignore_glob = ["functional/*"]
```
rshewitt, line 26: i figured this was some kind of global pytest looks for. [source](https://docs.pytest.org/en/7.1.x/reference/reference.html#globalvar-collect_ignore_glob)

tests/unit/cf/test_cf_tasks.py
```diff
@@ -0,0 +1,87 @@
...
+        return collections.defaultdict(list)
+
+    def get(self, a):
+        return collections.defaultdict(list)
+
+    def cancel(self, a):
+        return "cancelled"
+
+
+with patch("harvester.utils.CloudFoundryClient", CFClientMock), patch(
```
rshewitt, line 56: didn't know you could do multiple things in a context manager like this

tests/unit/cf/test_cf_tasks.py
```diff
@@ -0,0 +1,87 @@
...
+    "harvester.utils.TaskManager", TaskManagerMock
+):
+    CFUtil = CFHandler("url", "user", "password")
+
+
```
rshewitt, line 61: could we just patch the CFHandler class instead of all these underlying classes in the libraries?

btylerburton: We could, but then we're just testing our mocks. By only patching the supporting libraries we aren't responsible for, we ensure that the CFHandler logic is sound. 

The mock objects could probably be improved though.

rshewitt: okay yeah i see what you're doing

rshewitt (approved):

PR #78, opened 2024-05-29
fixes error signature for DCATUSToCKANException; modifies test case; …

### LINKED ISSUE #4752: Update DB after CKAN Sync
## User Story

In order to keep an accurate count of the state of our datasets, datagovteam wants to update the Harvesting database with the results of the CKAN Sync operation.

Sync here refers to records that need to created or updated in CKAN DB.

## Acceptance Criteria

- [ ] GIVEN a record needs to synced to CKAN
THEN we should add mark that dataset as "state: success" in the harvest_records db.

- [ ] GIVEN a record needs to validated 
AND we should create a new error in the harvest_error table
AND that dataset should be added to a list to be marked as "state: error" in the harvest_records db

## Background

## Sketch

- [ ] Loop over records to be synced
- [ ] Log the result of the sync operation
- [ ] Create an error if necessary at runtime

btylerburton:
# Pull Request

Resolves https://github.com/GSA/data.gov/issues/4752

## About

Updates tests to validate exception arguments. 

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/tests/integration/exception/test_exception_handling.py
+++ b/tests/integration/exception/test_exception_handling.py
@@ -5,14 +5,7 @@ from unittest.mock import patch
 import ckanapi
 import pytest
 
-import harvester
-from harvester.exceptions import (
-    DCATUSToCKANException,
-    ExtractExternalException,
-    ExtractInternalException,
-    SynchronizeException,
-    ValidationException,
-)
+from harvester.exceptions import ExtractExternalException, ExtractInternalException
 from harvester.harvest import HarvestSource
 
 
@@ -20,7 +13,7 @@ def download_mock(_, __):
     return dict({"dataset": []})
 
 
-class TestExceptionHandling:
+class TestCriticalExceptionHandling:
     def test_bad_harvest_source_url_exception(
         self,
         interface,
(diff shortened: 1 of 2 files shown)


tests/integration/exception/test_exception_handling.py
```diff
@@ -111,24 +106,40 @@ def test_validation_exception(
...
     ):
         interface.add_organization(organization_data)
-        source = interface.add_harvest_source(source_data_dcatus_invalid)
-        harvest_job = interface.add_harvest_job(job_data_dcatus_invalid)
+        interface.add_harvest_source(source_data_dcatus)
+        harvest_job = interface.add_harvest_job(job_data_dcatus)
 
         harvest_source = HarvestSource(harvest_job.id)
-        harvest_source.prepare_external_data()
+        harvest_source.get_record_changes()
```
rshewitt, line 123: like seeing all the things working together

tests/integration/exception/test_exception_handling.py
```diff
@@ -111,24 +106,40 @@ def test_validation_exception(
...
+        harvest_source.write_compare_to_db()
+        harvest_source.synchronize_records()
```
rshewitt, line 125: like seeing the things working together

tests/integration/exception/test_exception_handling.py
```diff
@@ -5,22 +5,15 @@
...
+from harvester.exceptions import ExtractExternalException, ExtractInternalException
 from harvester.harvest import HarvestSource
 
 
 def download_mock(_, __):
     return dict({"dataset": []})
 
 
-class TestExceptionHandling:
+class TestCriticalExceptionHandling:
```
rshewitt, line 16: +1 and same for non critical exception handling

rshewitt (approved):
it's allllll coming together

PR #80, opened 2024-06-03
organize by model; update with new models definitions

rshewitt:
# Pull Request

Related to [#4757](https://github.com/GSA/data.gov/issues/4757)

## About

update model definitions and use of models 

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -207,9 +207,7 @@ def edit_harvest_source(source_id=None):
 
 @mod.route("/harvest_source/harvest/<source_id>", methods=["GET"])
 def trigger_harvest_source(source_id):
-    job = db.add_harvest_job(
-        {"harvest_source_id": source_id, "status": "pending_manual"}
-    )
+    job = db.add_harvest_job({"harvest_source_id": source_id, "status": "manual"})
     if job:
         flash(f"Triggered harvest of source with ID: {source_id}")
     else:
@@ -232,18 +230,6 @@ def delete_harvest_source(source_id):
 
 
 ## Harvest Job
-@mod.route("/harvest_job/add", methods=["POST"])
-def add_harvest_job():
-    if request.is_json:
-        job = db.add_harvest_job(request.json)
-        if job:
-            return jsonify({"message": f"Added new harvest job with ID: {job.id}"})
-        else:
-            return jsonify({"error": "Failed to add harvest job."}), 400
-    else:
-        return jsonify({"Please provide harvest job with json format."})
-
-
 @mod.route("/harvest_job/", methods=["GET"])
 @mod.route("/harvest_job/<job_id>", methods=["GET"])
 def get_harvest_job(job_id=None):
@@ -265,6 +251,15 @@ def get_harvest_job(job_id=None):
         return "Please provide correct job_id or harvest_source_id"
 
 
+@mod.route("/harvest_job/<job_id>/error", methods=["GET"])
+def get_harvest_job_error(job_id):
+    try:
+        job_error = db.get_harvest_job_error(job_id)
+        return jsonify(job_error) if job_id else ("Not Found", 404)
+    except Exception:
+        return "Please provide correct job_id"
+
+
 @mod.route("/harvest_job/<job_id>", methods=["PUT"])
 def update_harvest_job(job_id):
     result = db.update_harvest_job(job_id, request.json)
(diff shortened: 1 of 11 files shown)


database/interface.py
```diff
@@ -199,19 +209,26 @@ def add_harvest_error(self, error_data):
             self.db.rollback()
             return None
 
-    def get_harvest_error(self, error_id):
-        result = self.db.query(HarvestError).filter_by(id=error_id).first()
+    def get_harvest_job_error(self, job_id: str) -> dict:
```
btylerburton, line 224: now that we have HarvestJobError and HarvestRecordError, and we will likely never have both in a single job, we could just query both tables with a job_id to remain consistent.

as there is definitely a case for all errors associated with a job_id for reporting results of a harvest run.

app/routes.py
```diff
@@ -265,6 +251,15 @@ def get_harvest_job(job_id=None):
         return "Please provide correct job_id or harvest_source_id"
 
 
+@mod.route("/harvest_job/<job_id>/error", methods=["GET"])
+def get_harvest_job_error(job_id):
+    try:
+        job_error = db.get_harvest_job_error(job_id)
+        return jsonify(job_error) if job_id else ("Not Found", 404)
```
btylerburton: shouldn't do the query if we don't have job_id

app/routes.py
```diff
@@ -352,6 +312,36 @@ def get_harvest_record(record_id=None):
...
+@mod.route("/harvest_record/<record_id>/errors", methods=["GET"])
+def get_all_harvest_record_errors(record_id: str) -> list:
+    try:
+        record_errors = db.get_harvest_record_errors(record_id)
+        return record_errors if record_errors else ("Not Found", 404)
+    except Exception:
+        return "Please provide correct record_id"
+
+
+@mod.route("/harvest_record/<error_id>/error", methods=["GET"])
```
btylerburton: if passing `error_id` i'd think route would be `harvest_error/<error_id>` and then we'd want to query both Error tables.

database/interface.py
```diff
@@ -187,9 +194,12 @@ def delete_harvest_job(self, job_id):
         self.db.commit()
         return "Harvest job deleted successfully"
 
-    def add_harvest_error(self, error_data):
+    def add_harvest_error(self, error_data: dict, error_type: str):
```
btylerburton, line 197: probably want a specific response for not passing `error_type` as that's going to be easy enough to forget now.

database/models.py
```diff
@@ -1,17 +1,26 @@
...
 from flask_sqlalchemy import SQLAlchemy
-from sqlalchemy import Enum, func
+from sqlalchemy import Enum, func, String, Column
+from sqlalchemy.orm import DeclarativeBase
 
-db = SQLAlchemy()
 
-
-class Base(db.Model):
+class Base(DeclarativeBase):
```
btylerburton, line 8: what's the advantage of shifting to this as base inheritance class?

database/models.py
```diff
@@ -1,17 +1,26 @@
...
     __abstract__ = True  # Indicates that this class should not be created as a table
-    id = db.Column(db.String(36), primary_key=True, default=lambda: str(uuid.uuid4()))
+    id = Column(String(36), primary_key=True, default=lambda: str(uuid.uuid4()))
+
+
+db = SQLAlchemy(model_class=Base)
 
 
-class Organization(Base):
+class Error(db.Model):
```
btylerburton, line 16: we should include `harvest_job_id` as common to both errors as we will need to correlate record level errors to the job unless i'm missing something.

btylerburton (changes requested):
This complicates record level error lookups by jobid which we'll need for harvest reporting.

database/interface.py
rshewitt: add [here](https://github.com/GSA/datagov-harvesting-logic/blob/8baa21616de9693b43ff78e40f8e05d0cfeba56e/database/interface.py#L214)

app/routes.py
rshewitt: paired and discussed

database/interface.py
rshewitt: added [here](https://github.com/GSA/datagov-harvesting-logic/blob/8baa21616de9693b43ff78e40f8e05d0cfeba56e/database/interface.py#L199)

database/models.py
rshewitt: paired and discussed

database/models.py
rshewitt: [resolved with](https://github.com/GSA/datagov-harvesting-logic/blob/8baa21616de9693b43ff78e40f8e05d0cfeba56e/database/interface.py#L214)

app/routes.py
```diff
@@ -352,6 +319,36 @@ def get_harvest_record(record_id=None):
...
```
btylerburton: shouldn't this be `harvest_error/<error_id>` for consistency?

tests/integration/database/test_db.py
```diff
@@ -101,6 +102,43 @@ def test_harvest_source_by_jobid(
...
+        source_data_dcatus,
+        job_data_dcatus,
+        job_error_data,
+    ):
+        interface.add_organization(organization_data)
+        interface.add_harvest_source(source_data_dcatus)
+        interface.add_harvest_job(job_data_dcatus)
+
+        harvest_job_error = interface.add_harvest_error(job_error_data, "job")
+        assert isinstance(harvest_job_error, HarvestJobError)
```
btylerburton, line 118: nice assertion!

btylerburton (dismissed):
Great cleanup!

app/routes.py
rshewitt: [resolved](https://github.com/GSA/datagov-harvesting-logic/pull/80/commits/6e745b6ccf238eb8d42e37cf3aff6248672aa2a5)

btylerburton (approved):

PR #81, opened 2024-06-04
Adds main function

### LINKED ISSUE #4715: Create harvest runner main function for dcatus
## User Story

In order for the harvest runner to process a harvest source, datagov wants to create a `main` function in `harvest.py` responsible for processing a harvest source.

## Acceptance Criteria

- [ ] GIVEN harvest.py \
  WHEN `main` is called \
  THEN a dcatus source should be harvested

## Background

## Sketch
- main function steps
  - parse sys.argv and get the jobid 
  - pull the associated harvest source information from the DB via the jobid 
  - create a `HarvestSource` instance using the derived harvest source information
  - get records changes
  - synchronize records
- get a simple harvest source as a test
- this'll be an integration test

btylerburton:
# Pull Request

Resolves https://github.com/GSA/data.gov/issues/4715

## About

Creates a main function to use when calling the app from the command line

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/.env
+++ b/.env
@@ -10,3 +10,4 @@ HARVEST_SOURCE_URL=http://localhost:${HARVEST_SOURCE_PORT}
 
 CF_API_URL=https://api.fr.cloud.gov
 LM_RUNNER_APP_GUID=f4ab7f86-bee0-44fd-8806-1dca7f8e215a
+CKAN_API_URL=https://catalog-dev.data.gov
--- a/.env
+++ b/.env
@@ -10,4 +10,3 @@ HARVEST_SOURCE_URL=http://localhost:${HARVEST_SOURCE_PORT}
 
 CF_API_URL=https://api.fr.cloud.gov
 LM_RUNNER_APP_GUID=f4ab7f86-bee0-44fd-8806-1dca7f8e215a
-CKAN_API_URL=https://catalog-dev.data.gov
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ up: ## Sets up local flask and harvest runner docker environments. harvest runne
 	DATABASE_PORT=5433 docker compose up -d
 	docker compose -p harvest-app up db -d
 
+up-unified: ## For testing when you want a shared db between flask and harvester
+	docker compose up -d
+
 down: ## Tears down the flask and harvester containers
 	docker compose down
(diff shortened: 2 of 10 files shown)


harvester/harvest.py
```diff
@@ -455,3 +456,26 @@ def sync(self) -> None:
...
+
+
+if __name__ == "__main__":
+    import argparse
+    import sys
+
+    parser = argparse.ArgumentParser()
+    try:
+        parser.add_argument("jobId", help="job id for harvest job")
+        args = parser.parse_args()
```
rshewitt: there's an existing argument parse function in [utils](https://github.com/GSA/datagov-harvesting-logic/blob/9527e6d320a250577721288919031cf8b0590532/harvester/utils.py#L22). you would just pass sys.argv[1:] to it

Makefile
```diff
@@ -44,6 +44,9 @@ up: ## Sets up local flask and harvest runner docker environments. harvest runne
 	DATABASE_PORT=5433 docker compose up -d
 	docker compose -p harvest-app up db -d
 
+up-unified: ## For testing when you want a shared db between flask and harvester
```
rshewitt, line 47: +1

harvester/harvest.py
rshewitt: if that feels too granular we can toss it in favor of doing it this way

rshewitt (dismissed):

harvester/harvest.py
btylerburton: found it. thanks!

.env
```diff
@@ -10,3 +10,4 @@ HARVEST_SOURCE_URL=http://localhost:${HARVEST_SOURCE_PORT}
 
 CF_API_URL=https://api.fr.cloud.gov
 LM_RUNNER_APP_GUID=f4ab7f86-bee0-44fd-8806-1dca7f8e215a
+CKAN_API_URL=https://catalog-dev.data.gov
```
FuhuXia: why does local docker use catalog-dev url?

btylerburton: good point. it shouldn't need that. it was for testing.

btylerburton: updated

FuhuXia (approved):
LGTM

PR #85, opened 2024-06-25
Update with ckan_id when create record in Ckan database

Jin-Sun-tts:
# Pull Request

Related to (https://github.com/GSA/data.gov/issues/4791)

## About

Retrieve the ckan_id from the CKAN package_create API call and update it in the record table in the harvestDB.


harvester/harvest.py
```diff
@@ -439,20 +451,24 @@ def validate(self) -> None:
             )
 
     def create_record(self):
-        ckan.action.package_create(**self.ckanified_metadata)
+        result = ckan.action.package_create(**self.ckanified_metadata)
+        self.ckanid = result['id']
 
     def update_record(self) -> dict:
         ckan.action.package_update(**self.ckanified_metadata)
```
btylerburton: did you test `package_update`? I believe `ckanified_metadata` method needs to be updated to include that field as `id`.

btylerburton (dismissed):
🚀

btylerburton (dismissed):

btylerburton (approved):

PR #86, opened 2024-06-26
Adds Scheduled Jobs Support to Load Manager

btylerburton:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4779

## About

Summary of changes:
- adds scheduled jobs and better support for manual jobs to load manager;
- removes 'manual' from job status enum;
- removes 'dict' type returns from interface in favor of ORM; cleans up tests after removal of to_dict;

## PR TASKS

- [X] Any changes to docs?


database/models.py
```diff
@@ -40,6 +40,18 @@ class HarvestSource(db.Model):
...
+            "manual",
+            "daily",
+            "weekly",
+            "biweekly",
+            "monthly",
+            name="frequency",
+        ),
+        nullable=False,
+        index=True,
+    )
```
jbrown-xentity: Aren't these incompatible? Does the second overwrite the first?

harvester/utils/utils.py
```diff
@@ -3,15 +3,14 @@
 import json
 import os
```
jbrown-xentity: Why the file name change? harvester/utils/utils.py seems a little over-nested...

tests/integration/app/test_load_manager.py
```diff
@@ -0,0 +1,192 @@
+from copy import copy
+from datetime import datetime, timedelta
+from unittest.mock import patch
```
jbrown-xentity: I'm not sure that this is an integration test, I feel that it is more of a unit test since all outside integration is mocked. If it was hitting something that resembled/copied cloudfoundry functionality then this would be an integration test...

jbrown-xentity (dismissed):
Ton of good cleanup in here, nice work @btylerburton ! Just a few comments to consider...

database/models.py
btylerburton: ooh good catch! that should be a replace.

harvester/utils/utils.py
btylerburton: We have `utils/ckan_utils.py` as well so I wanted to group them. I'd be open to a name change suggestion, or combining the two files and then just parking them at `harvester/utils.py`. Any preferences?

tests/integration/app/test_load_manager.py
btylerburton: Good suggestion. I'm going to rewrite the tests, particularly the first two, so we're inheriting the interface rather than mocking it, and we are just mocking the CF library dependency's responses.

harvester/utils/utils.py
btylerburton: renamed to `utils/general_utils.py`

jbrown-xentity (approved):

PR #87, opened 2024-07-01
add email list field

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4795

## About

fixed the email list bug. 
updated the login.gov CLIENT_ID for local docker.


jbrown-xentity (approved):
That looks to have been the problem we were seeing!

jbrown-xentity:
@btylerburton candidate for a UI test, as this bug seems to have derived from the UI and not from the model/processing code...

btylerburton:
Nice work, Jin! Glad to see this fixed.
Definitely a candidate for UI testing.

btylerburton (approved):

PR #88, opened 2024-07-08
Integrate harvester UI

### LINKED ISSUE #4792: Port Harvester UI elements to Flask App
## User Story

In order to streamline our codebase, datagovteam wants incorporate the templates created in `datagov-harvest-ui` into the `datagov-harvest-admin` app.

These elements will not require a login and will be read-only.

Adding these features to our Flask app allows us to shrink our application footprint (by removing the datagov-harvest-ui app from the stack), and allows us to leverage the already-created Flask backend.

## Acceptance Criteria

- [ ] GIVEN I have view logic that has been built in datagov-harvest-ui
  THEN I want to port that to the datagov-harvest-admin app

## Background

To install packages only available via NPM (ex. USWDS), follow this guide here: https://stackoverflow.com/questions/24514936/how-can-i-serve-npm-packages-using-flask

## Sketch

- [ ] port all template work done in datagov-harvest-ui into datagov-harvest-admin
- [ ] hook into pre-existing API endpoints to fetch required data
- [ ] confirm none of the view templates requires a login

btylerburton:
# Pull Request

Resolves https://github.com/GSA/data.gov/issues/4792

## About

Integrates Harvest UI elements from https://github.com/GSA/datagov-harvest-ui

## PR TASKS

- [ ] Any changes to docs?


btylerburton:
Comments out the session tests until I can isolate the GH Actions issue

app/routes.py
```diff
@@ -351,54 +351,100 @@ def add_harvest_source():
...
+                "backgroundColor": "black",
+            },
+            {
+                "label": "Errored",
+                "data": [job.records_errored for job in jobs],
+                "borderColor": "red",
+                "backgroundColor": "red",
+            },
+            {
+                "label": "Igrored",
```
rshewitt: typo. not trying to nitpick just wanna make sure your stuff gets looked at!

btylerburton: good catch!

btylerburton:
Session tests fixed. 🚀

.github/workflows/commit.yml
```diff
@@ -22,7 +22,6 @@ jobs:
         CF_SERVICE_AUTH: ${{secrets.CF_SERVICE_AUTH}}
         HARVEST_RUNNER_APP_GUID: ${{secrets.HARVEST_RUNNER_APP_GUID}}
         FLASK_APP_SECRET_KEY: ${{secrets.FLASK_APP_SECRET_KEY}}
-        OPENID_PRIVATE_KEY: ${{secrets.OPENID_PRIVATE_KEY}}
```
btylerburton, line 25: Removing this proactively as we don't need it for testing and it's not stored in GH as a secret.

btylerburton:
Just remembered that I need to add a new step to pipeline to build the static assets. This shouldn't hold up this ticket but it will hold up resolution of the issue.

btylerburton:
> Just remembered that I need to add a new step to pipeline to build the static assets. This shouldn't hold up this ticket but it will hold up resolution of the issue.

✅ Added new static assets build step

app/routes.py
```diff
@@ -245,56 +238,72 @@ def add_organization():
...
-            "edit_data.html",
-            form=form,
-            action="Edit",
-            data_type="Organization",
-            button="Update",
-        )
-    else:
-        org = db.get_all_organizations()
-    return org
+def edit_organization(org_id):
```
rshewitt, line 281: +1 for removing default

app/routes.py
```diff
@@ -351,54 +361,105 @@ def add_harvest_source():
...
+    return render_template("view_source_list.html", data=data)
 
 
 # Edit Source
-@mod.route("/harvest_source/edit/<source_id>", methods=["GET", "POST"])
+@mod.route("/harvest_source/config/edit/<source_id>", methods=["GET", "POST"])
 @login_required
-def edit_harvest_source(source_id: str = None):
+def edit_harvest_source(source_id: str):
     if source_id:
```
rshewitt, line 425: do we still need this condition if `source_id` is required?

app/routes.py
```diff
@@ -351,54 +361,105 @@ def add_harvest_source():
...
-        return render_template(
-            "edit_data.html",
-            form=form,
-            action="Edit",
-            data_type="Harvest Source",
-            button="Update",
-            source_id=source_id,
-        )
+        source = db.get_harvest_source(source_id)
+        organizations = db.get_all_organizations()
```
rshewitt, line 427: we have 133. that number probably isn't going to increase enough to make this call expensive. just a comment.

app/routes.py
btylerburton: good point. probably not. i'll make the update.

btylerburton: I'm going to defer this until the route cleanup ticket, if that's ok with you. There's some similar logic that exists in harvest_jobs and harvest_records that is inconsistent which should also be cleaned up.

rshewitt:
LGTM!

rshewitt (approved):

PR #89, opened 2024-08-13
Remove harvest source along with associated datasets in CKAN

Jin-Sun-tts:
Related to https://github.com/GSA/data.gov/issues/4691

- Corrected the API call to delete the harvest source and organization.
- Added API remote call to purge datasets in CKAN.
- Ensured all dataset records, errors, and jobs are removed from the harvestDB after harvest_source deleted.

### DIFF
--- a/database/interface.py
+++ b/database/interface.py
@@ -160,7 +160,7 @@ class HarvesterDBInterface:
         if not ckan_ids:
             self.db.delete(source)
             self.db.commit()
-            return "Harvest source deleted successfully, no records found."
+            return "Harvest source deleted successfully"
 
         ckan = RemoteCKAN(
             os.getenv("CKAN_API_URL"), apikey=os.getenv("CKAN_API_TOKEN"))
@@ -201,9 +201,9 @@ class HarvesterDBInterface:
             .all()
         )
         if remaining_records == 0:
-            return "Harvest source deleted successfully."
+            return "Harvest source deleted successfully"
         else:
-            return "Harvest source delete failed."
+            return "Harvest source delete failed"
 
 
 
--- a/database/interface.py
+++ b/database/interface.py
@@ -6,6 +6,7 @@ from sqlalchemy import create_engine, inspect, or_, text
 from sqlalchemy.exc import NoResultFound
 from sqlalchemy.orm import scoped_session, sessionmaker
 from ckanapi import RemoteCKAN, NotFound
+import logging
 
 from .models import (
     HarvestJob,
@@ -18,6 +19,9 @@ from .models import (
 )
 
 DATABASE_URI = os.getenv("DATABASE_URI")
+
+logger = logging.getLogger("harvest_admin")
+
 
 class HarvesterDBInterface:
     def __init__(self, session=None):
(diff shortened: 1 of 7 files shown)


database/interface.py
```diff
@@ -146,6 +146,27 @@ def delete_harvest_source(self, source_id):
...
+        )
+        ckan_ids = [record.ckan_id for record in records]
+        if ckan_ids:
+            ckan = RemoteCKAN(
+                os.getenv("CKAN_API_URL"), 
+                apikey=os.getenv("CKAN_API_TOKEN")
+            )
+            for pkg_id in ckan_ids:
+                try:
+                    ckan.action.dataset_purge(id=pkg_id)
```
rshewitt: i believe dataset_purge returns None when executed but can we increment a counter to indicate how many were purged without exception? ideally that number would be the same as the number of records deleted from the harvest_table after cascade. resource: https://stackoverflow.com/questions/8954116/sqlalchemy-get-rows-affected-by-bulk-delete

btylerburton: > ideally that number would be the same as the number of records deleted from the harvest_table after cascade

Is there a way to extract that record count from a foreign key cascade delete as opposed to an explicit delete as mentioned in the SO article above?

btylerburton (commented):
Adding a test to confirm DB retains no records with that Harvest Source Id would be nice.

database/interface.py
rshewitt: using [triggers](https://stackoverflow.com/questions/10124077/how-can-i-get-id-of-cascade-deleted-or-marked-for-delete-items-in-sqlalchemy)? i feel like triggers would have a noticeable impact on db performance so i'm immediately hesitant

btylerburton: I think we should write up a ticket to encompass this. It's only something we're going to be doing very rarely so db performance will be less of an issue, but we would want to ensure that the operation completes successfully.

Definitely a candidate for ongoing load testing. Does deleting 20k records cripple the admin app, and if so, for how long? Would it make more sense to run a special task in the runner?

Some more on triggers:
- https://docs.sqlalchemy.org/en/14/core/ddl.html#controlling-ddl-sequences
- https://www.postgresql.org/docs/current/sql-createtrigger.html

Jin-Sun-tts: added check for unsuccessfully dataset_purge. Also confirm no records remains after delete harvest_source.

database/interface.py
```diff
@@ -146,9 +150,62 @@ def delete_harvest_source(self, source_id):
...
+            return "Harvest source deleted successfully"
+
+        ckan = RemoteCKAN(
+            os.getenv("CKAN_API_URL"), apikey=os.getenv("CKAN_API_TOKEN"))
+
+        failed_delete = []
+
+        for pkg_id in ckan_ids:
+            try:
+                ckan.action.dataset_purge(id=pkg_id)
```
btylerburton, line 172: Will `dataset_purge` effectively reset the CKAN DB of all references? 
Have you tried to delete and then reharvest yet?

database/interface.py
```diff
@@ -146,9 +150,62 @@ def delete_harvest_source(self, source_id):
...
+        else:
+            # Only delete records that were successfully purged from CKAN
+            self.db.query(HarvestRecord).filter(
+                    ~HarvestRecord.ckan_id.in_(failed_delete),
+                    HarvestRecord.harvest_source_id == source_id
+                ).delete(synchronize_session=False)
+            
+            logger.warning(
+                f"Warning: Not all CKAN datasets could be deleted. "
+                f"{len(failed_delete)} records remain."
```
btylerburton: Would it be better to print the contents of `failed_delete` so that we have the IDs?

database/interface.py
```diff
@@ -146,9 +150,62 @@ def delete_harvest_source(self, source_id):
...
+
         self.db.commit()
-        return "Harvest source deleted successfully"
+
+        remaining_records = (
+            self.db.query(HarvestRecord)
+            .filter_by(harvest_source_id=source_id)
+            .all()
+        )
+        if remaining_records == 0:
```
btylerburton: I think we would still need to execute `self.db.delete(source)` here.

btylerburton:
> Adding a test to confirm DB retains no records with that Harvest Source Id would be nice.

This is an eventual nice-to-have. To do it right we'd need to interact with a real CKAN instance.

database/interface.py
Jin-Sun-tts: (Adding a test to confirm DB retains no records with that Harvest Source Id would be nice) -- I initially thought you were referring to HarvestDB, so I added a test to confirm that no records with that Harvest Source ID are retained. Now that I understand you meant the CKAN database, we don't need this step, as the self.db.delete(source) has already been called earlier. I removed it.

database/interface.py
Jin-Sun-tts: sure, good idea, I added it.

btylerburton:
@Jin-Sun-tts I remembered what I wanted to mention at sync...

I'm not so sure we should support this in the browser at all, as I just tried to delete NASA source on Friday and it hung for many minutes just cleaning the admin db records. I would expect that a deletion with accompanying CKAN sync would mean you'd have to leave the browser tab open for multiple hours. 

We should explore making this a cf-task, like harvesting.

Jin-Sun-tts:
open new PR with different branch for the clear function.

PR #90, opened 2024-08-16
Adds support for title deconflicting to create unique ckan urls

btylerburton:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4849

## About

Harvester creates CKAN "name" based on dataset title. When titles conflict, we need to namespace them with unique characters.

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -48,10 +48,7 @@ ckan = RemoteCKAN(
     session=session,
 )
 
-CKAN_ERRORS = {"URL_IN_USE": "That URL is already in use"}
-
 # logging data
-# logging.getLogger(name) follows a singleton pattern
 logger = logging.getLogger("harvest_runner")
 
 ROOT_DIR = Path(__file__).parents[1]
@@ -473,10 +470,21 @@ class Record:
                 self.harvest_source.internal_records_lookup_table[self.identifier],
             )
 
-    def create_record(self):
-        result = ckan.action.package_create(**self.ckanified_metadata)
-        self.ckan_id = result["id"]
-        self.ckan_name = self.ckanified_metadata["name"]
+    def create_record(self, retry=False):
+        try:
+            result = ckan.action.package_create(**self.ckanified_metadata)
+            self.ckan_id = result["id"]
+            self.ckan_name = self.ckanified_metadata["name"]
+        except Exception as e:
+            if retry is False:
+                self.ckanified_metadata["name"] = add_uuid_to_package_name(
+                    self.ckanified_metadata["name"]
+                )
+                self.ckan_name = self.ckanified_metadata["name"]
+                return self.create_record(retry=True)
+            else:
+                raise e
+                # will be caught by outer SynchronizeException
 
     def update_record(self) -> dict:
         updated_metadata = {
@@ -514,14 +522,12 @@ class Record:
                 self.harvest_source.internal_records_lookup_table[self.identifier],
             )
 
-    def sync(self, iter_count=0) -> None:
-        # ruff: noqa: F841
+    def sync(self) -> None:
         if self.valid is False:
             logger.warning(f"{self.identifier} is invalid. bypassing {self.action}")
             return
 
-        if iter_count == 0:
-            self.ckanify_dcatus()
+        self.ckanify_dcatus()
 
         start = datetime.now(timezone.utc)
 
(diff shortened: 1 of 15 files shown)


harvester/harvest.py
```diff
@@ -507,12 +531,19 @@ def sync(self) -> None:
...
+                )
+                self.ckan_name = self.ckanified_metadata["name"]
+                return self.sync(iter_count=1)
+            else:
+                self.status = "error"
+                raise SynchronizeException(
+                    f"failed to {self.action} for {self.identifier} :: {repr(e)}",
+                    self.harvest_source.job_id,
+                    self.harvest_source.internal_records_lookup_table[self.identifier],
+                )
```
jbrown-xentity: Maybe this is nitpicky, but I disagree that this should be handled as an exception. I believe this is a CKAN problem, and should be integrated into the CKAN utils library. It should retry with this function if the exception occurs, not pass back to the harvester system and then go back to CKAN.

jbrown-xentity (commented):
I don't believe this is the correct way to handle this. If @rshewitt or @Jin-Sun-tts agree with @btylerburton and not myself, please just approve and we can move on.

harvester/harvest.py
btylerburton: Do you mean adding this 5-char UUID to every dataset, or checking for naming conflicts prior to the POST, or some third thing?

>  I believe this is a CKAN problem, and should be integrated into the CKAN utils library. It should retry with this function if the exception occurs, not pass back to the harvester system and then go back to CKAN.

I opted for the recursion since I didn't want to replicate the network call.

rshewitt: is the ckan exception identifiable to where we could handle it like this? or do you have to parse the message it sends to figure out what happened? I know there's redundancy in ckanifying the record again but this seems simpler maybe? i feel like the likelihood of an infinite loop is very small.
```python
#...
try:
    if self.action == "create":
        self.create_record()
    if self.action == "update":
        self.update_record()
except CKANValidation as e:
    self.ckanified_metadata["name"] = add_uuid_to_package_name(
        self.ckanified_metadata["name"]
    )
    self.sync()
# every other exception?...
except Exception as e:
    self.status = "error"
    raise SynchronizeException(
        f"failed to {self.action} for {self.identifier} :: {repr(e)}",
        self.harvest_source.job_id,
        self.harvest_source.internal_records_lookup_table[self.identifier],
    )
#...
```

rshewitt: oh wait that wouldn't work

jbrown-xentity: I guess what I'm saying is that the create and update functions are just 1-liners, and it makes more sense to handle this logic there.
Also looking from a user perspective: I wouldn't expect a title change to change the name, that changes the URL for the user. Once the item is created, the URL should stay static (it shouldn't re-munge the name). So error catching that and re-trying when updating doesn't make sense to me. So this should be error handled and re-tried recursively (limited to 1 extra try) at the [create_record](https://github.com/GSA/datagov-harvester/pull/90/files#diff-598762acc9262886a623d9784cd9a5666592e736c2ae4df37c7d921673054e6eR476) level, and not this level.

rshewitt: i included `update_record` because I assumed a title change could happen but if that's uncommon i see the reason to apply the re-try logic just to `create_record`. but even if there was a title change then a re-try wouldn't be necessary? i guess it all depends on what it changes to

btylerburton: This is great feedback. I moved the retry loop to the `create_record` method and now that we have `ckan_name` in place the name won't change on update even if the title does. Going to write a new test to confirm my nested try/except bubbles up to the outer one and then will push my changes.

harvester/utils/ckan_utils.py
```diff
@@ -292,3 +293,7 @@ def ckanify_dcatus(metadata: dict, owner_org: str) -> dict:
     ckanified_metadata["extras"] = create_ckan_extras(metadata)
 
     return ckanified_metadata
+
+
+def add_uuid_to_package_name(name: str) -> str:
+    return name + "-" + str(uuid.uuid4())[:5]
```
rshewitt, line 299: the ckan title/name munger limits the title to the length of `PACKAGE_NAME_MAX_LENGTH`. could adding the first 5 characters of a uuid str here cause the output to be longer than that? idk if ckan complains about a name being too long

btylerburton: Ah good point. Let me take a look at the original code and if anything we shave those characters off the limit.

btylerburton: I trimmed the `PACKAGE_NAME_MAX_LENGTH` to 90. Take a look at the changes and make sure they look good to you.

harvester/harvest.py
```diff
@@ -197,8 +199,8 @@ def compare(self) -> None:
         logger.info("comparing our records with theirs")
 
         try:
-            external_ids = set(self.external_records.keys())
-            internal_ids = set(self.internal_records.keys())
+            external_ids = IndexedSet(self.external_records.keys())
+            internal_ids = IndexedSet(self.internal_records.keys())
```
rshewitt, line 203: why do we need an IndexedSet?

harvester/harvest.py
```diff
@@ -239,9 +241,10 @@ def write_compare_to_db(self) -> dict:
                         "harvest_job_id": record.harvest_source.job_id,
                         "harvest_source_id": record.harvest_source.id,
                         "source_hash": record.metadata_hash,
-                        "source_raw": str(record.metadata),
+                        "source_raw": json.dumps(record.metadata),
```
rshewitt, line 244: +1

harvester/harvest.py
btylerburton: to guarantee insertion order. 

i found it necessary to make the tests in here pass reliably, then considered that we want sources to harvest in the same order every time.

rshewitt (approved):

PR #91, opened 2024-08-16
Adds new add-org click command

btylerburton:
adds new add-org click command;
updates ckan api url to point to catalog next;

# Pull Request

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -22,6 +22,8 @@ logger = logging.getLogger("harvest_admin")
 
 user = Blueprint("user", __name__)
 mod = Blueprint("harvest", __name__)
+org = Blueprint("org", __name__)
+
 db = HarvesterDBInterface()
 
 # Login authentication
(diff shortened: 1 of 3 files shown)


jbrown-xentity (approved):
LGTM

app/routes.py
```diff
@@ -178,6 +181,24 @@ def remove_user(email):
         print("Failed to remove user or user not found.")
 
 
+## Org management
+@org.cli.command("add")
+@click.argument("name")
+@click.option("--logo", default="", help="Org Logo")
+@click.option("--id", default="", help="Org ID: should correspond to CKAN ORG ID")
+def cli_add_org(name, logo, id):
```
robert-bryson, line 189: This looks good to me, but to be clear since these parameters are positional if any parameters are not included a `missing required positional arguments` error will be thrown. The `if <arg>` checks below will only check that they have a truthy value not that they exist in scope. Just want to make sure that that is the intention here, though perhaps `click` handles this.

btylerburton: `name` is the only required field

`.argument` is positional
`.option` is optional / kwarg

https://click.palletsprojects.com/en/8.1.x/options/#basic-value-options

PR #95, opened 2024-08-29
Adds Harvest Source Size to Model

btylerburton:
# Pull Request

## About

Adds a size property to the HarvestSource model to allow us to pass in bigger memory option when the job requires it.

Also adds some minor fixes for:
- displaying all jobs for a harvest source, regardless of status
- cast errors by job method response to dict to prevent internal server error
- add status column to harvest job display
- tweaks some post-job reporting metrics

## PR TASKS

- [ ] Any changes to docs?


app/scripts/load_manager.py
```diff
@@ -36,6 +41,10 @@ def create_task(job_id, cf_handler=None):
     if cf_handler is None:
         cf_handler = create_cf_handler()
 
+    if size != "small":
+        task_contract["memory_in_mb"] = TASK_SIZE_ENUM[size][0]
+        task_contract["disk_in_mb"] = TASK_SIZE_ENUM[size][1]
+
```
jbrown-xentity: This will give a key error if size is something other than `medium` or `large`. Suggest validating before utilizing.

database/models.py
```diff
@@ -52,6 +52,7 @@ class HarvestSource(db.Model):
         index=True,
     )
     url = db.Column(db.String, nullable=False, unique=True)
+    size = db.Column(Enum("small", "medium", "large", name="size"))
```
jbrown-xentity, line 55: I guess with this being an enum the odds are tiny that bad input would get passed in...

jbrown-xentity (dismissed):
Looks pretty good! Up to you if you want to add error checking or not...

database/models.py
btylerburton: Yeah, that's what I was figuring.

app/scripts/load_manager.py
btylerburton: Would you suggest: `if size != "small and TASK_SIZE_ENUM[size]`?

jbrown-xentity: I think I would suggest `if size != "small" and size in TASK_SIZE_ENUM`

btylerburton: quick change incoming

jbrown-xentity (approved):

PR #96, opened 2024-08-30
Removes size and status from harvest source model

btylerburton:
# Pull Request

## About

Removes size and status from harvest source model in favor of passing in same val for memory and disk to all tasks, with eventual focus on shaving memory requirements off after proper profiling

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/app/scripts/load_manager.py
+++ b/app/scripts/load_manager.py
@@ -10,18 +10,12 @@ CF_API_URL = os.getenv("CF_API_URL")
 CF_SERVICE_USER = os.getenv("CF_SERVICE_USER")
 CF_SERVICE_AUTH = os.getenv("CF_SERVICE_AUTH")
 HARVEST_RUNNER_APP_GUID = os.getenv("HARVEST_RUNNER_APP_GUID")
-CF_INSTANCE_INDEX = os.getenv("CF_INSTANCE_INDEX")
 
 MAX_TASKS_COUNT = 3
 
 interface = HarvesterDBInterface()
 
 logger = logging.getLogger("harvest_admin")
-
-TASK_SIZE_ENUM = {
-    "medium": (2048, 768),
-    "large": (4096, 1536),
-}
 
 
 def create_cf_handler():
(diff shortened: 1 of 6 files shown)


app/scripts/load_manager.py
```diff
@@ -10,19 +10,13 @@
 CF_SERVICE_USER = os.getenv("CF_SERVICE_USER")
 CF_SERVICE_AUTH = os.getenv("CF_SERVICE_AUTH")
 HARVEST_RUNNER_APP_GUID = os.getenv("HARVEST_RUNNER_APP_GUID")
-CF_INSTANCE_INDEX = os.getenv("CF_INSTANCE_INDEX")
```
rshewitt, line 13: why is this removed?

harvester/lib/cf_handler.py
```diff
@@ -14,10 +15,10 @@ def setup(self):
         self.client.init_with_user_credentials(self.user, self.password)
         self.task_mgr = TaskManager(self.url, self.client)
 
-    def start_task(self, app_guuid, command, task_id, memory_in_mb=512, disk_in_mb=512):
-        return self.task_mgr.create(
-            app_guuid, command, task_id, memory_in_mb, disk_in_mb
-        )
+    def start_task(self, app_guuid, command, task_id):
+        TASK_MEMORY = os.getenv("HARVEST_RUNNER_TASK_MEM", "4096")
```
btylerburton, line 19: I can't figure out how to mock these if I put them at the head of the file. I know that gets calculated at import time vs execution time, and [I could use `pytest-env` if I wanted them to be a single, static value](https://github.com/pytest-dev/pytest/discussions/10027), but does anyone on @GSA/data-gov-dev-team know how to monkeypatch them for a single test?

app/scripts/load_manager.py
btylerburton: it's imported manually here: https://github.com/GSA/datagov-harvester/pull/96/files#diff-8304ae22782c8fdf080c8374d1a6431e2ca7b7a9e4050dd71540fd7e67532072R98 so this is redundant

harvester/lib/cf_handler.py
rshewitt: does memory and disk space need to derive from the env? could they be default args similar to what you had previously?

btylerburton: i'd like them to be configurable, at least while we are trying to nail down what the correct values should be.

btylerburton:
Any further changes or is this good to merge @rshewitt?

rshewitt (approved):

PR #98, opened 2024-09-06
Add and Modify API routes for Harvest source Clear and Delete

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4787 and https://github.com/GSA/data.gov/issues/4691

## About

- Added a new API route to `clear `a harvest source. This includes invoking bulk_update_delete for remote CKAN datasets and removing entries from the harvest_record and harvest_record_error tables.

- Updated the `delete` action to ensure it only proceeds if the harvest source has no associated records. If records exist, a warning is displayed, prompting users to clear them first.

- Added a "Harvest Records" section on the harvest source page to display the total number of harvest records.

- Fixed incorrect paths for certain delete routes.


database/interface.py
```diff
@@ -142,13 +148,102 @@ def update_harvest_source(self, source_id, updates):
...
+        organization_id = source.organization_id
+
+        records = (
+            self.db.query(HarvestRecord)
+            .filter_by(harvest_source_id=source_id).all()
+        )
+        if not records:
+            return "Harvest source has no records to clear."
+
+        ckan_ids = [record.ckan_id for record in records if record.ckan_id is not None]
```
FuhuXia, line 183: what if  len(ckan_ids) < len(records)? is there an ongoing harvesting activity for this source?

app/templates/view_org_data.html
```diff
@@ -35,7 +35,7 @@ <h2>Organization Config Table</h2>
             </li>
             <li class="usa-button-group__item">
                 <button class="usa-button usa-button--secondary"
-                    onclick="confirmDelete('/organization/delete/{{data.organization.id}}')">Delete</button>
+                    onclick="confirmDelete('/organization/config/delete/{{data.organization.id}}')">Delete</button>
```
FuhuXia, line 38: why /something/**config**/action/id, not /something/action/id?

database/interface.py
```diff
@@ -142,13 +148,102 @@ def update_harvest_source(self, source_id, updates):
...
+                result = ckan.action.package_search(fq=f'owner_org:{organization_id}')
+                ckan_datasets = result['count']
+
+                if ckan_datasets == 0:
+                    _clear_harvest_records()
+                    logger.info(f"Total time: {datetime.now(timezone.utc) - start}")
+                    return "Harvest source cleared successfully."
+                else:
+                    logger.info(f"Retry {retry_count + 1}: "
+                                "{ckan_datasets} datasets remaining, retrying...")
```
FuhuXia: These lines do not belong to this try block.

database/interface.py
```diff
@@ -142,13 +148,102 @@ def update_harvest_source(self, source_id, updates):
...
+            except Exception:
```
FuhuXia: What exception we are catching here? add some comments to explain.
It feels like the only expected exception is action bulk_update_delete. Other exception should be dealt with differently.

database/interface.py
```diff
@@ -142,13 +148,102 @@ def update_harvest_source(self, source_id, updates):
...
+            .filter_by(harvest_source_id=source_id).all()
+        )
+       
+        if len(records) == 0:
+            self.db.delete(source)
+            self.db.commit()
+            return "Harvest source deleted successfully"
+        else:
+            return (f"Failed: {len(records)} records in the Harvest source, "
+                    "please Clear it first.")
```
FuhuXia, line 261: Can we add a test for this?

database/interface.py
Jin-Sun-tts: when len(ckan_ids) < len(records), one situation is maybe there is harvesting activity at same time, the other situation is when calling package_create api, if there is an error, there is no ckan_id for certain records.

app/templates/view_org_data.html
Jin-Sun-tts: not very sure, it was defined in route earlier, we may double check with Tyler @btylerburton

database/interface.py
Jin-Sun-tts: right, it runs only after clean up all dataset which outside of the try block.

database/interface.py
Jin-Sun-tts: good point, let me try to add other exception handling.

database/interface.py
FuhuXia: So len(ckan_ids) + len(error records) = len(records). We need verify this before sending the list of ckan_ids to ckan to clear. Imagine there is an ongoing harvesting activity and some ckan_id is added after this line, ckan can clear all records according to your id list but the newly added ckan_id will be left untouched and will affect action.package_search result.

Jin-Sun-tts: we may add a check to make sure the number match before clear. Or check this page for any in_progress job etc.

database/interface.py
Jin-Sun-tts: fixed.

database/interface.py
Jin-Sun-tts: Fixed.

database/interface.py
Jin-Sun-tts: Fixed

database/interface.py
Jin-Sun-tts: added the check for the two situations : len(ckan_ids) + len(error records) = len(records) and check no job in_progress

FuhuXia (approved):

app/routes.py
```diff
@@ -432,6 +432,10 @@ def add_harvest_source():
 def view_harvest_source_data(source_id: str):
     source = db.get_harvest_source(source_id)
     jobs = db.get_all_harvest_jobs_by_filter({"harvest_source_id": source.id})
+    records = db.get_harvest_record_by_source(source.id)
```
btylerburton, line 435: we need to write a proper count query that we can use for things like this...

app/routes.py
```diff
@@ -536,19 +543,30 @@ def edit_harvest_source(source_id: str):
     return db._to_dict(source)
 
 
-### Delete Source
+# Clear Source
+@mod.route("/harvest_source/config/clear/<source_id>", methods=["POST"])
```
btylerburton, line 547: when we modify the routes in https://github.com/GSA/data.gov/issues/4815 we can standardize the naming. `config` doesn't make much sense to me anymore.

app/routes.py
```diff
@@ -536,19 +543,30 @@ def edit_harvest_source(source_id: str):
...
+@login_required
+def clear_harvest_source(source_id):
+    try:
+        result = db.clear_harvest_source(source_id)
+        flash(result)
+        return {"message": "success"}
```
btylerburton, line 553: we should probably reroute the request to the home page

app/routes.py
```diff
@@ -536,19 +543,30 @@ def edit_harvest_source(source_id: str):
...
+    except Exception as e:
+        logger.error(f"Failed to clear harvest source :: {repr(e)}")
+        flash("Failed to clear harvest source")
+        return {"message": "failed"}
```
btylerburton, line 557: same here but we should reload the source page to show the flashed message

app/templates/view_org_data.html
btylerburton: we can clean up the naming in https://github.com/GSA/data.gov/issues/4815.

app/templates/view_source_data.html
```diff
@@ -9,6 +9,9 @@
 {% endblock %}
 
 {% block content %}
+<div id="flash-message" style="display:none; position:fixed; top:10px; left:50%; transform:translateX(-50%); background-color:#ffcc00; color:black; padding:10px; border:1px solid #000; z-index:1000;">
```
btylerburton, line 12: hmmm interesting option.. to have some sort of banner that shows while an op is in progress, as a standard flashed message won't do it if we need to keep the page open while the request is in progress.

app/templates/view_source_data.html
```diff
@@ -41,11 +44,33 @@ <h2>Harvest Source Config Table</h2>
...
+            <li class="usa-button-group__item">
+                <button class="usa-button usa-button--secondary"
+                    onclick="confirmAction('delete', '/harvest_source/config/delete/{{data.harvest_source.id}}')">Delete</button>
             </li>
         </ul>
     </div>
     {% endif %}
+    <br>
+    <h2>Harvest Records: </h2>
+    <div class="config-table">
```
btylerburton, line 58: this makes me think we should nail down what data we want to show in a table for each page: org, source, job, record, error

app/templates/view_source_data.html
```diff
@@ -122,8 +151,15 @@ <h2>Harvest Jobs</h2>
                 body: {}
             });
             let res = await rawResponse.json()
+
+            flashMessage.style.display = 'none';
+
             if (res.message && res.message == 'success') {
-                return window.location.replace('/')
+                if (action === 'delete') {
+                    return window.location.replace('/');
```
btylerburton, line 159: i'd like to find a way to do this in flask context, instead of relying on in page javascript.

database/interface.py
```diff
@@ -142,13 +149,117 @@ def update_harvest_source(self, source_id, updates):
...
+            return "Harvest source has no records to clear."
+
+        ckan_ids = [record.ckan_id for record in records if record.ckan_id is not None]
+        error_records = [record for record in records if record.status == 'error']
+        jobs_in_progress = self.get_all_harvest_jobs_by_filter(
+            {"harvest_source_id": source.id, "status": "in_progress"}
+        )
+
+        # Ensure no jobs are in progress
+        if jobs_in_progress:
```
btylerburton, line 190: we could move this check way up the chain since it's cheap.

database/interface.py
```diff
@@ -142,13 +149,117 @@ def update_harvest_source(self, source_id, updates):
...
+            return ("Error: A harvest job is currently in progress. "
+                    "Cannot clear datasets.")
+
+        # Ensure (error_records + ckan_ids) = total records
+        if len(error_records) + len(ckan_ids) != len(records):
```
btylerburton, line 195: can you explain this check?

*also i think there's some issue with excess double quoting here in the return val.

database/interface.py
```diff
@@ -142,13 +149,117 @@ def update_harvest_source(self, source_id, updates):
...
+            return ("Error: Not all records are either in an error state "
+                    "or have a CKAN ID. Cannot proceed without clearing the dataset.")
+
+        if not ckan_ids:  
+            _clear_harvest_records()
+            return "Harvest source cleared successfully."
+
+        ckan = RemoteCKAN(
```
btylerburton, line 203: we should make the CKAN client a class and then we can share it around

btylerburton: also, why are we not doing this as a cf-task?

database/interface.py
```diff
@@ -142,13 +149,117 @@ def update_harvest_source(self, source_id, updates):
...
+            _clear_harvest_records()
+            logger.info(f"Total time: {datetime.now(timezone.utc) - start}")
+            return "Harvest source cleared successfully."
+        else:
+            fail_message = (f"Harvest source clearance failed after {retry_count} "
+                            f"attempts. {ckan_datasets} datasets still exist in CKAN.")
+            logger.error(fail_message)
+            return fail_message
+
     def delete_harvest_source(self, source_id):
```
btylerburton, line 245: this is only for records on the flask side?
shouldn't we clear the CKAN source before deleting?

btylerburton (commented):
Added a couple things to consider for future enhancements even though this is merged. Nice work!

PR #99, opened 2024-09-12
Adds Pagination and Partial SPA-like page rerendering

btylerburton:
Features added:
- adds pagination and count decorators for large endpoints;
- adds htmx and render_block libs to support re-rendering a partial template; 
- adds USWDS pagination component and support templates; 
- adds fixture json to support local dev and a CLI command to load testdata;

# Pull Request

Related to https://github.com/GSA/data.gov/issues/4810

## PR TASKS

- [ ] Any changes to docs?


btylerburton:
Given there's a lot of front-end UI code in here, I figured it was worth recording a video of our new pagination component in action:


https://github.com/user-attachments/assets/baa6b149-82d4-4596-8acf-cd3474d159ec

app/routes.py
```diff
@@ -245,6 +249,31 @@ def cli_remove_harvest_source(id):
...
+        for item in fixture["job"]:
+            db.add_harvest_job(item)
+        for item in fixture["record"]:
+            db.add_harvest_record(item)
+        for item in fixture["record_error"]:
+            db.add_harvest_record_error(item)
+
+    click.echo("Done.")
+
+
```
jbrown-xentity: This is really to seed a database with test data, correct? The "load" click command makes me a bit nervous, just from the sense that it's a bit of a misnomer... Can we change this to load_test_data, or something?

app/templates/base.html
```diff
@@ -8,6 +8,7 @@
     <link rel="stylesheet" href="{{ url_for('static', filename='assets/uswds/css/styles.css') }}">
     <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.5.2/css/bootstrap.min.css">
     <script src="{{ url_for('static', filename='assets/uswds/js/uswds-init.js') }}"></script>
+    <script src="{{ url_for('static', filename='assets/htmx/htmx.min.js') }}"></script>
```
jbrown-xentity, line 11: Are there any processes we should put in place to track this htmx minimized version? Add notifications from their github releases to slack, or something like that?

jbrown-xentity (dismissed):
Looks really good for a massive PR @btylerburton , nice work! Just a few small comments...

app/routes.py
btylerburton: i see your point. it's currently `flask testdata load`, but i can make it `flask testdata load_test_data` just to be more explicit. i'd like to make it a script but we have the endpoints locked down with login.gov so i figure this is the best/easiest route.

app/templates/base.html
btylerburton: Good point. I can take a look at that. Since it's managed by NPM we should also be able to point SNYK at that file.

btylerburton: Added `app/static/package.json` to SNYK and HTMX release notifications to `#datagov-notifications`.

jbrown-xentity (approved):

PR #103, opened 2024-10-22
Add extra metadata fields

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4847

## About

Added three keys and their corresponding values to the extras field in the metadata.
Replaced org_id with harvest_source_id when clearing the dataset as it is available now. 
Added new route for raw record data `harvest_record/<record_id>/raw
`

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -25,7 +25,6 @@ from harvester.exceptions import (
     SynchronizeException,
     ValidationException,
 )
-from harvester.utils.ckan_utils import add_uuid_to_package_name, ckanify_dcatus
 from harvester.utils.general_utils import (
     dataset_to_hash,
     download_file,
@@ -473,6 +472,7 @@ class Record:
             )
 
     def create_record(self, retry=False):
+        from harvester.utils.ckan_utils import add_uuid_to_package_name
         try:
             result = ckan.action.package_create(**self.ckanified_metadata)
             self.ckan_id = result["id"]
@@ -512,6 +512,7 @@ class Record:
         )
 
     def ckanify_dcatus(self) -> None:
+        from harvester.utils.ckan_utils import ckanify_dcatus
         try:
             self.ckanified_metadata = ckanify_dcatus(
                 self.metadata, self.harvest_source
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -247,7 +247,6 @@ class HarvestSource:
                         "ckan_name": record.ckan_name,
                     }
                 )
-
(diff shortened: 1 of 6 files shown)


harvester/harvest.py
```diff
@@ -515,7 +514,7 @@ def update_self_in_db(self) -> bool:
     def ckanify_dcatus(self) -> None:
         try:
             self.ckanified_metadata = ckanify_dcatus(
-                self.metadata, self.harvest_source.organization_id
+                self.metadata, self.harvest_source
```
FuhuXia, line 518: `self.harvest_source.organization_id` and `self.harvest_source` are different types, how does one replace the other? could it be `self.harvest_source.id`?

harvester/utils/ckan_utils.py
```diff
@@ -151,7 +151,7 @@ def munge_tag(tag: str) -> str:
     return tag
 
 
-def create_ckan_extras(metadata: dict) -> list[dict]:
+def create_ckan_extras(metadata: dict, harvest_source) -> list[dict]:
```
FuhuXia: provide type hint to be consistent.

harvester/utils/ckan_utils.py
```diff
@@ -283,14 +301,14 @@ def simple_transform(metadata: dict, owner_org: str) -> dict:
     return output
 
 
-def ckanify_dcatus(metadata: dict, owner_org: str) -> dict:
-    ckanified_metadata = simple_transform(metadata, owner_org)
+def ckanify_dcatus(metadata: dict, harvest_source) -> dict:
```
FuhuXia: provide type hint to be consistent.

FuhuXia (changes requested):

harvester/harvest.py
Jin-Sun-tts: Previously, we only required the org_id, but now we need additional fields from harvest_source. As a result, we are passing the entire object to allow access to the other fields from harvest_source.

harvester/utils/ckan_utils.py
Jin-Sun-tts: Added.

harvester/utils/ckan_utils.py
Jin-Sun-tts: The HarvestSource class is defined in harvest.py, which imports functions from ckan_util. This results in a circular import error. To resolve this, move the import of the function to the specific location where it is used. Then added the type hint into this function.

FuhuXia (approved):

PR #104, opened 2024-10-30
Adds Gunicorn WSGI web server

btylerburton:
- adds support for gunicorn wsgi web server; 
- integrates new relic support; 
- cleans up some naive queries in favor of faceted implementations;

# Pull Request

Related to https://github.com/GSA/data.gov/issues/4940

## PR TASKS

- [ ] Any changes to docs?


docker-compose_prod.yml
```diff
@@ -0,0 +1,3 @@
+services:
+  app:
+    command: ./harvest-admin-start.sh
```
jbrown-xentity, line 3: This isn't actually used for production, right? Is this used to simulate our production startup on cloud.gov?

harvest-admin-start.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+DIR="$(dirname "${BASH_SOURCE[0]}")"
+
+exec newrelic-admin run-program gunicorn "wsgi:application" --config "$DIR/gunicorn.conf.py" -b "0.0.0.0:$PORT" --chdir $DIR --timeout 120 --worker-class gevent --workers 8 --forwarded-allow-ips='*'
```
jbrown-xentity, line 5: Curious, why did you go with 8 workers and no threads?

tests/integration/harvest/test_compare.py
```diff
@@ -36,3 +38,50 @@ def test_compare(
...
+            [
+                "cftc-dc3",
+                "cftc-dc7",
+                "cftc-dc5",
+                "cftc-dc4",
+                "cftc-dc6",
+                "cftc-dc1",
+                "cftc-dc2",
+            ]
+        )
```
jbrown-xentity, line 85: It's a little confusing that you used the sort function to create the expected; but of course it works. Just gave me pause reading it.

jbrown-xentity (approved):
No reason this PR can't move forward, it's a ton of positive changes! If you have time, I had a few comments/questions...

harvest-admin-start.sh
btylerburton: Threads only applies to the gthread worker class: https://docs.gunicorn.org/en/stable/settings.html#threads

Worker recommendation is `2-4 x CPU_CORES`, and cloud.gov docs say each VM can have "multiple cores", so basically just an educated guess

docker-compose_prod.yml
btylerburton: Correct

tests/integration/harvest/test_compare.py
btylerburton: this is an old test. i had to move it b/c it does something to the DB which interferes with the pagination tests above. more than likely it's the `harvest_source.write_compare_to_db()`. leaving it where it was mucks up pagination for some reason. definitely worth considering if this indicates a deeper bug or just a symptom of a weird test scenario though. 

i could also set conftest to clean db after every function instead of just the session but that slows things down dramatically and this seemed the better route.

btylerburton: if i left this where it was, this assertion wouldn't pass on first run, but would pass when i ran it isolation
https://github.com/GSA/datagov-harvester/pull/104/files#diff-4f46b3a8f79b223696492f1254bdf0ec30ea2b66c04bb3388e2d0aaacc55d8e0R531

PR #112, opened 2024-11-20
Add harvest mail

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4582

## About

Sends a summary email to data providers upon harvest job completion.
Email content includes a summary of job and a link to the job page for detailed information.


harvester/harvest.py
```diff
@@ -380,6 +386,50 @@ def report(self) -> None:
...
+                    msg["Subject"] = subject
+                    msg.attach(MIMEText(body, "plain"))
+
+                    server.sendmail(SMTP_CONFIG["default_sender"], [recipient],
+                                    msg.as_string())
+                    logger.info(f"Notification email sent to: {recipient}")
+
+        except Exception as e:
+            logger.error(f"Error preparing or sending notification emails: {e}")
+
```
jbrown-xentity, line 432: We could write a "failing" test here. Give it some real job information that should work, but send it fake/junk smtp code. Then have the test confirm that the expected smtp error comes out of the test. Then, if anything with the job information object changes, we'll catch that because the error code will be different and the test will actually fail.

jbrown-xentity (approved):
Looks pretty good!
I would add a test, see details below. But if we need to move on and come back to this later, that's ok.

harvester/harvest.py
btylerburton: @Jin-Sun-tts here's an example of forcing an error with the CKAN lib: https://github.com/GSA/datagov-harvester/blob/main/tests/integration/harvest/test_harvest_full_flow.py#L50

Jin-Sun-tts: Thank you @jbrown-xentity and @btylerburton. I will merge the code the add the failing test in separate PR.

PR #114, opened 2024-11-21
add fed & non-fed dcatus1.1 schemas

rshewitt:
# Pull Request

Related to [#4870](https://github.com/GSA/data.gov/issues/4870)

## PR TASKS

- [ ] Any changes to docs?


jbrown-xentity (approved):
Looks good!
In the old version, we had a federal and non-federal catalog and single item definitions: https://github.com/GSA/ckanext-datajson/tree/main/ckanext/datajson/pod_schema/non-federal-v1.1
I think in the new version, that is unnecessary. The only difference is the single item definitions. However, how can we have a catalog schema that points to either federal or non-federal, depending upon the use case, without duplicating the catalog JSON schema definition for both use cases? I'm happy to discuss further in person if necessary.

rshewitt:
> Looks good! In the old version, we had a federal and non-federal catalog and single item definitions: https://github.com/GSA/ckanext-datajson/tree/main/ckanext/datajson/pod_schema/non-federal-v1.1 I think in the new version, that is unnecessary. The only difference is the single item definitions. However, how can we have a catalog schema that points to either federal or non-federal, depending upon the use case, without duplicating the catalog JSON schema definition for both use cases? I'm happy to discuss further in person if necessary.

the harvest runner currently only validates using the single item definition. we don't use the catalog schema. are we intending on doing that?

jbrown-xentity:
> > Looks good! In the old version, we had a federal and non-federal catalog and single item definitions: https://github.com/GSA/ckanext-datajson/tree/main/ckanext/datajson/pod_schema/non-federal-v1.1 I think in the new version, that is unnecessary. The only difference is the single item definitions. However, how can we have a catalog schema that points to either federal or non-federal, depending upon the use case, without duplicating the catalog JSON schema definition for both use cases? I'm happy to discuss further in person if necessary.
> 
> the harvest runner currently only validates using the single item definition. we don't use the catalog schema. are we intending on doing that?

Nope. So you're good, good point!

PR #115, opened 2024-11-21
Decouple flask app

btylerburton:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4865

## About

Allows us to run follow-up jobs to sync records which may have previously errored out due network side-effects.

Moves towards ability to run harvest clear and records validation as first-class jobs.

EDIT: This PR is a beast. Make sure to expand the changes to `harvester/harvest.py`. I recommend looking at those in the side-by-side editor as it's a thorough rewrite.

## PR TASKS

- [x] Any changes to docs?


btylerburton:
~DRAFT WIP: not finished yet... but wanted to put this up as we're pausing on harvester work for a sprint and having eyes on this could help clean it up...~

.github/workflows/deploy.yml
rshewitt, line 1: why do we want to turn off deployment?

btylerburton: We don't have apps in staging or prod yet

rshewitt:
a lot has changed here and looks like pytest coverage is [broken](https://github.com/GSA/datagov-harvester/actions/runs/12918107080/job/36025960186?pr=115#annotation:7:31) (and has been broken for a bit). can this be fixed in this PR? low coverage could be fixed in a separate ticket.

btylerburton:
> a lot has changed here and looks like pytest coverage is [broken](https://github.com/GSA/datagov-harvester/actions/runs/12918107080/job/36025960186?pr=115#annotation:7:31) (and has been broken for a bit). can this be fixed in this PR? low coverage could be fixed in a separate ticket.

I didnt realize that coverage action depended on the xml. I can update that.

docs/diagrams/mermaid/dest/job_follow_up_job-1.svg
jbrown-xentity, line 1: We can probably remove Agency Source and MDTranslator from this diagram, they aren't used and make things more busy.

docs/diagrams/mermaid/dest/job_full_harvest-1.svg
jbrown-xentity, line 1: I think there's a missing piece in the LOAD step; that should always save in the DB that the record is loaded into CKAN, right? That way the follow up job doesn't have to analyze, it will just pick up where it left off?

docs/diagrams/mermaid/dest/job_restart_job-1.svg
jbrown-xentity, line 1: Do we have documentation on a restarted job vs a follow-up job? They seem almost identical...

jbrown-xentity (commented):
This is a review of the diagrams.
@btylerburton and I agreed that the historical diagrams don't make sense here, as they are overly thorough and confusing. These will be moved to a google drive folder, with a link for posterity.
The other comments are hopefully helpful to improving and clarifying the diagrams for future users.
@cmhedrickREI I encourage you to look at the new ones, as they show the high level systems, what is talking with what, and when.

.github/workflows/commit.yml
```diff
@@ -51,8 +51,10 @@ jobs:
       - name: Report test coverage
         uses: MishaKav/pytest-coverage-comment@main
         with:
-          pytest-coverage-path: pytest-coverage.txt
-          junitxml-path: pytest.xml
+          multiple-files: |
```
rshewitt, line 53: coverages still don't appear to be [working](https://github.com/GSA/datagov-harvester/actions/runs/12944490684/job/36105474355#step:7:1). it's not a huge deal. can create a separate ticket for it.

app/routes.py
```diff
@@ -465,23 +465,22 @@ def view_harvest_source_data(source_id: str):
     source = db.get_harvest_source(source_id)
     records_count = db.get_harvest_records_by_source(
```
rshewitt, line 466: seems like this would return outdated records as well. is this number supposed to represent the latest set? [db get latest records per source](https://github.com/GSA/datagov-harvester/blob/1e7cd5d35ef51020819183510733df7e682062ee/database/interface.py#L420)

app/routes.py
```diff
@@ -465,23 +465,22 @@ def view_harvest_source_data(source_id: str):
...
         source_id=source.id,
     )
     ckan_records_count = db.get_harvest_records_by_source(
         count=True,
-        skip_pagination=True,
         source_id=source.id,
-        facets=["ckan_id is not null"],
+        facets="ckan_id is not null",
     )
     error_records_count = db.get_harvest_records_by_source(
```
rshewitt, line 475: seems like this would also include outdated erroneous records?

app/routes.py
```diff
@@ -465,23 +465,22 @@ def view_harvest_source_data(source_id: str):
...
         count=True,
-        skip_pagination=True,
         source_id=source.id,
-        facets=["status = 'error'"],
+        facets="status = 'error'",
+    )
```
rshewitt, line 479: getting counts from `db.get_harvest_records_by_source` would include historic records ( e.g. record_count and ckan_records_count = 2 given a create then update ). error_records_count could also return more than expected if a record failed validation, is fixed by the data provider, then fails sync. here's the [get_latest_records](https://github.com/GSA/datagov-harvester/blob/1e7cd5d35ef51020819183510733df7e682062ee/database/interface.py#L420) function. it could be useful in these cases?

harvester/harvest.py
```diff
@@ -104,17 +112,34 @@ def __post_init__(self) -> None:
             self.dataset_schema = open_json(
                 ROOT_DIR / "schemas" / "non-federal_dataset.json"
             )
+        self._validator = Draft202012Validator(self.dataset_schema)
```
rshewitt, line 116: +1. no need to create this every time we validate a record.

harvester/harvest.py
```diff
@@ -197,37 +222,38 @@ def external_records_to_id_hash(self, records: list[dict]) -> None:
...
-
     def prepare_external_data(self) -> None:
         logger.info("retrieving and preparing external records.")
         try:
             job = self.db_interface.get_harvest_job(self.job_id)
-            if self.source_type == "document":
-                if job.job_type == 'clear':
-                    self.external_records_to_id_hash([])
-                else:
+            if job.job_type == "clear":
```
rshewitt: this method is called in `extract`. why does the clear operation need an extraction to occur? putting this here because this prompted my discovery

database/interface.py
```diff
@@ -237,6 +241,17 @@ def get_first_harvest_jobs_by_filter(self, filter):
         )
         return harvest_job
 
+    def get_harvest_jobs_by_source_id(self, source_id):
+        """used by follow-up job helper"""
+        harvest_jobs = (
+            self.db.query(HarvestJob)
+            .filter_by(harvest_source_id=source_id)
+            .order_by(HarvestJob.date_created.desc())
+            .limit(2)
```
rshewitt, line 247: so follow-up jobs are supposed to be responsible for completing unfinished work from a previous harvest job which can be represented by (work completed) - (work to-be-done). “follow_up_job_helper“ expects the job report to have occurred in order to determine the work completed. What kind of record-level work would need to be followed up on? We wouldn’t be following up on transformation or validation error records because we’re using the same data as last time so we’d get the same result.

btylerburton: we would be following up on a broken sync. we've definitely talked about a way to force a sync in isolation. perhaps i should just pare back `follow-up` and `restart` and just make a `sync` job?

.github/workflows/commit.yml
btylerburton: i'm looking into that now. just pushed up a test commit.

app/routes.py
btylerburton: i didn't add this block, just cleaned it up, but i agree we should not be showing "shadow" records that are deleted; this is only useful for the internal harvester when clearing a source. it took me a long time to reconcile exactly what that number meant.

database/interface.py
rshewitt: i think that would be simpler. do we want to restart a job older than what follow-up already does? thinking about where restart could come in where follow-up doesn't

btylerburton: ok, i'm going to pare this back to a simple `sync` job.

harvester/harvest.py
btylerburton: clear works by diffing against an empty incoming source list. i wonder if we can move that insertion up the chain, though. will take a look.

.github/workflows/commit.yml
```diff
@@ -64,12 +77,16 @@ jobs:
...
         uses: actions/checkout@v4
+      - name: Install node.js
+        uses: actions/setup-node@v4
         with:
-          fetch-depth: 1
-      - name: Set up Python ${{ env.PY_VERSION }}
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{ env.PY_VERSION }}
+          node-version: "20"
```
rshewitt: you should be able to use the package.json file and it will discover the engine version

harvester/harvest.py
```diff
@@ -275,173 +300,250 @@ def write_compare_to_db(self) -> dict:
...
+                    "ckan_id": record.ckan_id,
+                    "ckan_name": record.ckan_name,
+                }
+                if action is not None:
+                    db_record = self.db_interface.add_harvest_record(db_record)
+                    record.id = db_record.id
+                records.append(record)
+
+        # set records on new
+        self._records = records
```
rshewitt: _records shouldn't be set directly here. you'll want to add a setter where the "records" property is declared.

harvester/harvest.py
```diff
@@ -275,173 +300,250 @@ def write_compare_to_db(self) -> dict:
...
         }
+        # account for follow-up job's previous status in report count
+        if "previous_job_results" in self.report:
+            for key in self.report["previous_job_results"].keys():
+                job_status[key] = max(
+                    job_status[key] - self.report["previous_job_results"][key], 0
+                )  # don't allow for negative values based on previous results
+
         self.db_interface.update_harvest_job(self.job_id, job_status)
+        self._report = job_status
```
rshewitt, line 418: same here. _report attr shouldn't be set explicitly.

harvester/harvest.py
```diff
@@ -275,173 +300,250 @@ def write_compare_to_db(self) -> dict:
...
+                    msg["Reply-To"] = "no-reply@gsa.gov"
+                    msg["Subject"] = subject
+                    msg.attach(MIMEText(body, "plain"))
+
                     server.sendmail(
                         SMTP_CONFIG["default_sender"], [recipient], msg.as_string()
                     )
                     logger.info(f"Notification email sent to: {recipient}")
+
         except Exception as e:
```
rshewitt, line 466: not sure how frequently this would occur but does only logging this give us the visibility we're looking for? would we want to create a custom exception and add it as a job error?

harvester/harvest.py
```diff
@@ -275,173 +300,250 @@ def write_compare_to_db(self) -> dict:
...
+        logger.info(f"running clear helper for {self.name}")
+        ## get remaining records in db
+        db_records = self.db_interface.get_harvest_records_by_source(
+            paginate=False,
+            source_id=self.id,
+        )
+        records = []
+        for db_record in db_records:
+            records.append(self.make_record_contract(db_record, input_type="orm"))
+        self._records.extend(records)
```
rshewitt: same. update to self.records not self._records

harvester/harvest.py
btylerburton: yes! i'll add a todo and we can take care of this in another ticket.

.github/workflows/commit.yml
btylerburton: next pr should test a new config based on these docs: https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#working-with-lockfiles

harvester/harvest.py
btylerburton: resolved

harvester/harvest.py
btylerburton: resolved

docs/diagrams/mermaid/dest/job_full_harvest-1.svg
btylerburton: docs should all be updated with that info.

docs/diagrams/mermaid/dest/job_restart_job-1.svg
btylerburton: refactored to just a single sync job after a chat with reid.

database/interface.py
```diff
@@ -237,6 +238,17 @@ def get_first_harvest_jobs_by_filter(self, filter):
         )
         return harvest_job
 
+    def get_harvest_jobs_by_source_id(self, source_id):
```
rshewitt, line 241: looks like this function isn't being used anymore

btylerburton: Probably. Lots of cleanup still needed.

harvester/harvest.py
```diff
@@ -275,173 +302,240 @@ def write_compare_to_db(self) -> dict:
...
+                "records_added": job.records_added,
+                "records_updated": job.records_updated,
+                "records_deleted": job.records_deleted,
+                "records_ignored": job.records_ignored,
+                "records_errored": job.records_errored,
+            }
+        }
+        self._report = results
+        new_records = []
+        for job_record in job.records:
```
rshewitt, line 486: we should be skipping records with a status = "success"

btylerburton: we skip those in the `record.sync()`, but it might be good to exclude those here.

btylerburton: updated just in case and no tests failed

rshewitt (approved):
we got a lotta good work over here!

PR #117, opened 2024-12-03
Add test email failure

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4582

Added a test for send email failure


harvester/harvest.py
```diff
@@ -426,9 +426,12 @@ def send_notification_emails(self, results: dict) -> None:
                     server.sendmail(SMTP_CONFIG["default_sender"], [recipient],
                                     msg.as_string())
                     logger.info(f"Notification email sent to: {recipient}")
+                    return "Emails sent successfully"
 
         except Exception as e:
-            logger.error(f"Error preparing or sending notification emails: {e}")
+            error_message = "Error preparing or sending notification emails"
+            logger.error(error_message + f": {e}")
+            return error_message
```
jbrown-xentity: Interesting idea, and this leads to a good test case. But what do we do with this error_message? Is it sent somewhere? I'm trying to picture how we want this raised, and I'm not sure.
Ironically, maybe this is best as an email to datagovhelp@gsa.gov. Something like "Harvest email failure", or something, that contains the error message and/or the email addresses and/or the harvest source.
Maybe the log is fine, and then we just have a new relic view or alert that looks for these log statements and allows us to easily review them. Maybe we can talk at sync today.

Jin-Sun-tts: sure, or another way we may have mock logger to confirm the log message instead of return message.

Jin-Sun-tts: The email address datagovhelp@gsa.gov is already included in the distribution list and will receive the same emails as other users. However, if the email notification connection fails, datagovhelp@gsa.gov will not receive a failure notification email. In such cases, the issue may only be identified through the logs.

harvester/harvest.py
```diff
@@ -438,9 +438,12 @@ def send_notification_emails(self, results: dict) -> None:
                     server.sendmail(SMTP_CONFIG["default_sender"], [recipient],
                                     msg.as_string())
                     logger.info(f"Notification email sent to: {recipient}")
+                    return "Emails sent successfully"
```
FuhuXia: the function return str now, so the function hint `-> None` needs to be `-> str`

FuhuXia (changes requested):

harvester/harvest.py
Jin-Sun-tts: fixed.

tests/integration/harvest/test_harvest_full_flow.py
```diff
@@ -162,3 +161,38 @@ def test_validate_same_title(
...
+        )
+        job_id = harvest_job.id
+        harvest_source = HarvestSource(job_id)
+        harvest_source.notification_emails = source_data_dcatus_single_record[
+                                            "notification_emails"
+                                            ]
+
+        results = {"create": 10, "update": 5, "delete": 3, None: 2}
+        mock_smtp.side_effect = Exception("SMTP connection failed")
+        error_message = harvest_source.send_notification_emails(results)
```
btylerburton: What are we doing with the error message?

harvester/harvest.py
FuhuXia: It feels like it should return None. I don't see the returned string consumed any where.

harvester/harvest.py
```diff
@@ -438,9 +438,12 @@ def send_notification_emails(self, results: dict) -> None:
                     server.sendmail(SMTP_CONFIG["default_sender"], [recipient],
```
FuhuXia, line 438: Not part of this commit, but does function halt if it fails on one recipient? so it fails for all other?

btylerburton: We can put the try within the for loop to account for that.

btylerburton (commented):
We probably need a test of the full flow to determine best how to handle the send mail failure.

harvester/harvest.py
Jin-Sun-tts: just tried, currently it will not fail for all others when there is an invalid email address, but we may try to catch it for the SMTPRecipientsRefused exception if there is need.

harvester/harvest.py
Jin-Sun-tts: Modified the approach to directly test the log error. so removed the returned value

harvester/harvest.py
Jin-Sun-tts: right, just fixed.

tests/integration/harvest/test_harvest_full_flow.py
Jin-Sun-tts: Modified the approach to directly test the log error. so removed the returned value

FuhuXia (approved):

PR #125, opened 2025-01-13
Datasets purge from Ckan and harvestdb

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4852

## About

- Modify the process for deleting or purging datasets from CKAN and the harvest database by utilizing the harvest process.
- Add a job_type field to the harvest_job table to distinguish between regular harvest jobs and clear jobs.


app/templates/view_data.html
```diff
@@ -27,8 +27,10 @@ <h2>{{action}} {{data_type}}</h2>
         <a href="{{ url_for('harvest.trigger_harvest_source', source_id=source_id)}}">
             <button class="btn btn-secondary">Harvest</button>
         </a>
-        <button class="btn btn-danger"
-            onclick="confirmAction('harvest source', 'clear', '/harvest_source/config/clear/{{source_id}}')">Clear</button>
+        <a href="{{ url_for('harvest.trigger_harvest_source', source_id=data.harvest_source.id, job_type='clear') }}"
+            onclick="return confirm('Are you sure you want to clear all datasets?');">
```
btylerburton, line 31: you can run confirmation flow without relying on a JS function! oooh, that's a nice find.

database/interface.py
```diff
@@ -190,113 +187,15 @@ def update_harvest_source(self, source_id, updates):
...
-            )
-            try:
-                ckan.action.bulk_update_delete(
-                    datasets=ckan_ids, org_id=organization_id
-                )
-            except ckanapi.errors.CKANAPIError as api_err:
-                logger.error(f"CKAN API error: {api_err}")
-            except Exception as err:
-                logger.error(f"Error occurred: {err} \n error_type: {type(err)}")
-                return f"Error occurred: {err}"
```
jbrown-xentity, line 274: I'm not quite following; where does this logic move to? Are we losing the ability to delete from CKAN, or is this cleanup of old code?

btylerburton: I believe we just trick the harvester into doing the work for us by pushing an empty list as incoming harvest object.

database/interface.py
```diff
@@ -626,5 +534,6 @@ def get_harvest_records_by_job(self, job_id, facets=[], **kwargs):
         return self.pget_harvest_records(filter=text(filter_string), **kwargs)
 
     def get_harvest_records_by_source(self, source_id, facets=[], **kwargs):
-        filter_string = " AND ".join([f"harvest_source_id = '{source_id}'"] + facets)
+        filter_string = " AND ".join([f"harvest_source_id = '{source_id}'"] 
+                        + ["action != 'delete'"]+ facets)
```
btylerburton, line 541: are the records now being retained as deletes, or is this to cleanup an earlier bug?

database/models.py
```diff
@@ -89,6 +89,7 @@ class HarvestJob(db.Model):
         nullable=False,
         index=True,
     )
+    job_type = db.Column(db.String(20), default="harvest")
```
btylerburton, line 92: this is great. we were going to need this sooner or later.

database/interface.py
```diff
@@ -503,6 +402,15 @@ def update_harvest_record(self, record_id, updates):
             self.db.rollback()
             return None
 
+    def delete_harvest_record(self, ckan_id):
+        records = self.db.query(HarvestRecord).filter_by(ckan_id=ckan_id).all()
```
btylerburton, line 406: should there be cases where we have more than one record here?

btylerburton (dismissed):

database/interface.py
Jin-Sun-tts: Yes, we removed the separate call to the CKAN API for deleting/purging. Instead, we now reuse the current harvest process with an empty list, which effectively purges all datasets from both CKAN and the harvestDB.

database/interface.py
Jin-Sun-tts: I noticed duplicate records with the same ckan_id in the current development environment, but I'm not sure how they ended up there.

database/interface.py
Jin-Sun-tts: no, it is just when a clear/delete operation in progress, any records marked with the 'delete' action should be excluded from the count when refreshing the page.

database/interface.py
btylerburton: Should we long the count temporarily to determine if this issue persists?

Jin-Sun-tts: Sure, I’ll make a note to revisit this later after cleaning the test environment.

Jin-Sun-tts: just added a log message for the duplicate ckan_id cases.

btylerburton (approved):

PR #127, opened 2025-01-21
add harvest terms glossary

rshewitt:
# Pull Request

Related to [#4979](https://github.com/GSA/data.gov/issues/4979)

## About
- picture of feature is on the ticket mentioned above

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/app/static/_scss/components/_glossary.scss
+++ b/app/static/_scss/components/_glossary.scss
@@ -1,0 +1,49 @@
+#glossary {
+  transition: right 0.2s;
+  background-color: #112f4e;
+
+  &[aria-hidden="true"] {
+    right: -20rem;
+  }
+
+  div[aria-hidden="true"] {
+    display: none !important;
+  }
+
+  button {
+    color: white;
+    background: none;
+    border: 0;
+    @include u-text("left");
+  }
+
+  .glossary__definition {
+    @include u-font("sans", "xs");
+    @include u-line-height("sans", 6);
+    @include u-margin-y(2);
+    @include u-width("full");
+
+    a {
+      @include u-text('base-light');
+    }
+  }
+
+  .glossary__term {
+    @include u-border-bottom("1px");
+    @include u-border("base");
+    @include u-margin(0);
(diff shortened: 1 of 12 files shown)


app/static/_scss/components/_glossary.scss
```diff
@@ -0,0 +1,49 @@
+#glossary {
+  transition: right 0.2s;
+  background-color: #112f4e;
```
btylerburton: here's a list of the uswds color tokens. can you find something that works from this list?

https://designsystem.digital.gov/design-tokens/color/theme-tokens/

also, here's how we make use of those in the 11ty project: https://github.com/GSA/datagov-11ty/blob/main/styles/partials/datagovtheme.scss#L102

btylerburton: i believe this should work since we import uswds-core prior to this.

rshewitt: resolved in latest commit

app/static/gulpfile.js
```diff
@@ -1,8 +1,33 @@
...
+function jsTask() {
+    return browserify(`${paths.js.src}`)
+        .transform('babelify', {
+            presets: ['@babel/preset-env'],
+            plugins: ['@babel/plugin-transform-runtime']
+        })
+        .bundle()
+        .pipe(source(paths.js.dest))
+        .pipe(buffer())
+        .pipe(dest("./"));
```
btylerburton, line 25: i believe you're writing twice. once to dest path and once to relative root

app/static/js/index.js
```diff
@@ -0,0 +1,4 @@
+// var SmoothScroll = require("smooth-scroll");
```
btylerburton: you got some undeleted comments here

btylerburton (dismissed):
🚢

app/static/js/index.js
rshewitt: fixed

btylerburton (approved):

PR #128, opened 2025-01-28
Scroll Bars are always displayed on the job error table

cmhedrickREI:
# Pull Request

Related to [4982](https://github.com/GSA/data.gov/issues/4982)
## About
This makes it so that the scroll bars on the Job error table are always displayed so no hover is needed.

Screenshot from 2025-01-28 13-58-14

Note for consideration:
- I know we have a similar table on the Organization Data page, so I made the scss change specific to the Job data Page only. The change could be made so we affect both tables though.

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/app/static/_scss/_uswds-theme-custom-styles.scss
+++ b/app/static/_scss/_uswds-theme-custom-styles.scss
@@ -63,8 +63,6 @@ ul.menu {
   }
 }
 
-div.job-data {
-  .usa-table-container--scrollable {
-    overflow: scroll;
-  }
+.usa-table-container--scrollable {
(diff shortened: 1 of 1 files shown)


rshewitt:
@cmhedrickREI i think for continuity applying it to both is good

cmhedrickREI:
> @cmhedrickREI i think for continuity applying it to both is good

Thank you @rshewitt I think that's a good call, as I much rather have the same elements be uniform. I just pushed another commit to update it

btylerburton (approved):
Looks great!

PR #131, opened 2025-02-03
Added title field to the Harvest Error Record Report page with block view

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4980

## About

- Retrieve the title and identifier for each record with an error and display them in a table.
- Added a block view for each error record, along with a button to toggle between the block view and table view.


rshewitt:
pagination is buggy with switch from table to block view. see video.

https://github.com/user-attachments/assets/911011e3-e433-4fb8-b2f4-21b0550bb2ad

app/templates/view_job_data.html
```diff
@@ -73,41 +73,65 @@ <h2>Job Error Table</h2>
...
+        <h2>Record Error Details</h2>
         {% if not data.record_errors %}
-        No record errors found
+            <p>No record errors found</p>
         {% else %}
-        {% set error_type = "record" %}
-        <a href="{{ url_for('harvest.download_harvest_errors_by_job', job_id=data.harvest_job_id, error_type=error_type)}}">
-            <button class="btn btn-primary">download record errors as .csv</button>
-        </a>
+        <button id="toggleViewBtn" onclick="toggleView()">Switch to Table View</button>
```
rshewitt: - can you apply styling from this? https://designsystem.digital.gov/components/button/ 
- may want to align the buttons. something like this?
Screenshot 2025-02-03 at 12 06 18 PM

app/templates/view_job_data.html
```diff
@@ -73,41 +73,65 @@ <h2>Job Error Table</h2>
...
             {% block record_errors_table %}
             <div id="error_results_pagination">
+                <div id="blockView" class="error-list">
+                    {% for record_error in data.record_errors %}
+                        <div class="error-block">
+                            <h3>{{ record_error.error.id }}</h3>
+                            <p><strong>Identifier:</strong> {{ record_error.identifier }}</p>
+                            <p><strong>Title:</strong> {{ record_error.title }}</p>
+                            <p><strong>Harvest Record ID:</strong>
+                                <a href="{{ url_for('harvest.get_harvest_record', record_id=record_error.error.harvest_record_id) }}">
```
rshewitt, line 96: this returns raw json. not sure if we planned on creating a dedicated page for a single harvest record but i think it breaks continuity. nothing to change. just wanted to mention it.

app/templates/view_job_data.html
```diff
@@ -124,11 +148,80 @@ <h2>Record Error Table</h2>
                 {% if pagination.count > data.record_errors|count %}
                     {% include '/components/pagination/pagination.html' %}
                 {%endif%}
+                <script>
+                    document.addEventListener("DOMContentLoaded", function() {
```
rshewitt: can this be moved to static assets? my frontend isn't the best so if not all good.

app/templates/view_job_data.html
```diff
@@ -124,11 +148,80 @@ <h2>Record Error Table</h2>
...
+                        } else {
+                            blockView.style.display = "block";
+                            tableView.style.display = "none";
+                            toggleBtn.textContent = "Switch to Table View";
+                        }
+                    }
+                </script>
+                
+
+                <style>
```
rshewitt, line 190: same here

app/templates/view_job_data.html
```diff
@@ -124,11 +148,80 @@ <h2>Record Error Table</h2>
...
+                        text-indent: 20px;
+                    }
+                    .error-block a {
+                        color: #0071bc;
+                        text-decoration: none;
+                    }
+                    .error-block a:hover {
+                        text-decoration: underline;
+                    }
+                </style>
```
rshewitt: same here. moved to static assets.

app/templates/view_job_data.html
btylerburton: the listener needs to be set in JS, but we should also be emitting an event whenever we push new updates via HTMX. 
Also, I don't have experience with local storage and Flask, but I'd venture that's worth exploring to keep things clean.

rshewitt:
pytest check appears [broken](https://github.com/GSA/datagov-harvester/actions/runs/13115040315/job/36587186911#step:6:582). green check with failing tests.

app/templates/view_job_data.html
Jin-Sun-tts: removed the toggle for different views as the download function provide user the table view.

app/templates/view_job_data.html
Jin-Sun-tts: good suggestion. Thanks!

app/templates/view_job_data.html
Jin-Sun-tts: removed the toggle function as we have the download option for the table view.

app/templates/view_job_data.html
Jin-Sun-tts: done

Jin-Sun-tts:
pytest errors were [fixed](https://github.com/GSA/datagov-harvester/actions/runs/13162317276/job/36733909685) 

> pytest check appears [broken](https://github.com/GSA/datagov-harvester/actions/runs/13115040315/job/36587186911#step:6:582). green check with failing tests.

database/interface.py
```diff
@@ -345,8 +345,15 @@ def get_harvest_record_errors_by_job(self, job_id: str, **kwargs):
             .filter(HarvestRecord.harvest_job_id == job_id)
             .subquery()
         )
-        query = self.db.query(HarvestRecordError).filter(
-            HarvestRecordError.harvest_record_id.in_(select(subquery))
+        query = (
+            self.db.query(
+                HarvestRecordError,
```
btylerburton, line 363: what are we doing here by passing in these three ORM query items?

tests/integration/harvest/test_transform.py
```diff
@@ -46,9 +46,9 @@ def test_invalid_transform_iso19115_2(
 
         job_errors = interface.get_harvest_record_errors_by_job(harvest_job.id)
         assert len(job_errors) == 1
-        assert job_errors[0].message == expected_error_msg
+        assert job_errors[0][0].message == expected_error_msg
```
btylerburton, line 49: we have an array inside an array. what does that represent?

Jin-Sun-tts: same reason like above, so the first item contains all fields from harvestRecordError, which need to be referenced using two-dimensional indexing.

database/interface.py
Jin-Sun-tts: We added two extra fields, title and identifier from the harvest_record table, in addition to all fields in harvestRecordError, causing the new query to return a tuple instead of a single object.

rshewitt: download to csv isn't working correctly. it's being paginated when it should return all errors @Jin-Sun-tts

tests/integration/harvest/test_transform.py
Jin-Sun-tts: fixed.

database/interface.py
Jin-Sun-tts: fixed.

rshewitt: we have access to the record when querying record errors because of the [relationship](https://github.com/GSA/datagov-harvester/blob/61326ff219a404633b8267f67304ea98f20deda2/database/models.py#L129-L130) in the db. i'm not sure we need to complicate this function further.

app/static/_scss/view-styles.css
btylerburton, line 1: would be preferrable to use USWDS color palette

https://designsystem.digital.gov/design-tokens/color/system-tokens/

btylerburton, line 1: I generally try not to use px for anything but border/border-radius.

https://hackernoon.com/pixels-vs-rem-vs-em-breaking-down-css-sizes

app/templates/base.html
```diff
@@ -5,6 +5,7 @@
     {% block title %}
         <title>{{action}} {{data_type}}</title>
     {% endblock %}
+    <link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='_scss/view-styles.css') }}">
```
btylerburton: Add your scss sheet to this file and it will get bundled
https://github.com/GSA/datagov-harvester/blob/harvest-email-on-errors/app/static/_scss/_uswds-theme-custom-styles.scss#L25

database/interface.py
btylerburton, line 363: why are we returning a tuple when the interface method is `get_harvest_record_errors_by_job`? shouldn't we just be returning a list of errors?

btylerburton: if we need a unified data structure we can use `zip`

Jin-Sun-tts: Instead of a list of errors with information for harvest_record_error, we need an additional field: identifier and title (source_raw) from the harvest_record table.

btylerburton: If we need this data structure for this one specific call, could we call it something else, and could you put some type hinting for what it returns?

app/static/_scss/view-styles.css
Jin-Sun-tts: done.

app/templates/base.html
Jin-Sun-tts: Done.

app/static/_scss/view-styles.css
Jin-Sun-tts: Done

database/interface.py
Jin-Sun-tts: Added a detailed description for clarification.

btylerburton (approved):

PR #132, opened 2025-02-03
Fixes issues with cg deploy

btylerburton:
# Pull Request

Related to [broken development action](https://github.com/GSA/datagov-harvester/actions?query=branch%3Adevelop)

## PR TASKS

- [ ] Any changes to docs?

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -79,13 +79,17 @@ jobs:
       - name: Install node.js
         uses: actions/setup-node@v4
         with:
-          node-version: "package.json"
+          node-version-file: "app/static/package.json"
       - name: Build static assets
         shell: bash
         run: |
-          cd app/static
-          npm ci
-          npm run build
+            cd app/static
+            npm ci
+            npm run build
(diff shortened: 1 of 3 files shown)


.github/workflows/commit.yml
```diff
@@ -99,7 +103,7 @@ jobs:
       - name: deploy DHL
         uses: cloud-gov/cg-cli-tools@main
         with:
-          command: cf push --vars-file vars.development.yml --strategy rolling --no-wait
```
jbrown-xentity, line 102: Just a reminder that this is fine for dev (and maybe even staging), but we probably want to keep around for production...

jbrown-xentity (dismissed):

.github/workflows/commit.yml
btylerburton: I removed this b/c CG build errors out when pushing multiple apps this way. I could fix this by making a multiline command though, which is what I'm going to do. Will take a bit to test.

btylerburton: this is resolved

jbrown-xentity (approved):

PR #133, opened 2025-02-05
Harvest Email Notification Frequency

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#4984

## About
In an effort to give Organizations more control over how many emails they receive this PR adds the functionality of setting the frequency of when they should receive emails which can be set on the `HarvestSource` level. 

- Adds the enum of `notification_frequency` to the `HarvestSource` model
    - Option of either `on_error` or `always`
        - `on_error` will only send the email when there are errors
        - `always` will send the email if errors are detected or if there are none
    - This is a required field
- Update added to the fixture.json
- Updated existing test fixtures to now include the required `notification_frequency` attribute.
- Add tests to assert the `send_notification_emails()` is called or not based on the `notification_frequency`
- Adds migration file to update the db to include `notification_frequency`
- Adds glossary entry for `notification_frequency`

## PR TASKS

- [x] Any changes to docs?


tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
```diff
@@ -10,8 +10,10 @@
 
 class TestHarvestJobFullFlow:
```
rshewitt, line 11: can you add an assertion for the notification email being called regardless of error? could put it in `test_multiple_harvest_jobs` since the harvest source is being updated so you could update the notification frequency as well

rshewitt:
since you're updating the model and the field is displayed in the UI you'll want to update the [glossary](https://github.com/GSA/datagov-harvester/blob/main/app/static/data/glossary.json) with a definition of what the term and values mean.

btylerburton (dismissed):
This looks great. We should create a new enhancement ticket to email when we hit a Critical exception as well.

cmhedrickREI:
@btylerburton Sorry to ask for a review again but I was previously missing the changes @rshewitt asked for and the migration. But now both the changes and migration should now be included!

migrations/env.py
```diff
@@ -1,41 +1,43 @@
...
+
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
 config = context.config
 
 # Interpret the config file for Python logging.
 # This line sets up loggers basically.
 fileConfig(config.config_file_name)
-logger = logging.getLogger("alembic.env")
+logger = logging.getLogger('alembic.env')
```
btylerburton: something is up with your formatter. linter should not sub single for double-quotes.

btylerburton (dismissed):
looks good outside of those formatting issues.

migrations/env.py
cmhedrickREI: I ran the lint command and I think I fixed it. :+1:

btylerburton (dismissed):

btylerburton (approved):

btylerburton:
Going to go ahead and merge this as it's ready and comments have been addressed.

PR #134, opened 2025-02-06
add db clean script

rshewitt:
# Pull Request

Related to [#4774](https://github.com/GSA/data.gov/issues/4774)

- add db-clean script
- add get outdated records function
- convert latest records sql text to orm
- fix deletion bug
- lint

notes
- one fix in this PR is how harvest records are deleted in the [interface](https://github.com/GSA/datagov-harvester/blob/d2cdeb585de8e38b010502b7fbc8608d15cee308/database/interface.py#L422). before this change, harvest records were deleted by just the identifier. this can result in unintended deletions because different harvest sources may have records with the same identifier name ( e.g.  "1", "2", "3" ). the change allows for the deletion by db record ID or by identifier & harvest source ID. also, this PR includes wrapping the deletion with a rollback on exception.
- given a record to delete, the harvester [currently](https://github.com/GSA/datagov-harvester/blob/d2cdeb585de8e38b010502b7fbc8608d15cee308/harvester/harvest.py#L753-L754) will delete all versions of that record within the harvest source. this means, assuming everything goes well, there shouldn't be outdated versions of a record to clean.
- the need for a db clean script isn't necessary assuming the deletions mentioned in the comment above works as intended. a db clean script would be valuable if we wanted to isolate deletions as a separate process (e.g. every saturday) but this means the database would grow in size until the script runs. 
- do we want to perform deletions per job or dedicate a single process for bulk deletions? 
- it's <ins>vital</ins> for deletions to function correctly otherwise how we derive the latest records will give the wrong results.
- i updated the "date_created" default values from `func.now()` to `func.statement_timestamp()` because during testing harvest records occurring in different [jobs](https://github.com/GSA/datagov-harvester/blob/d2cdeb585de8e38b010502b7fbc8608d15cee308/tests/integration/harvest_job_flows/test_harvest_job_full_flow.py#L43) all had the exact same "date_created" value which isn't correct. i confirmed there is a time delta between jobs. this causes how we get the latest records to produce incorrect results breaking a test. this led me to this [SO page](https://stackoverflow.com/questions/13370317/sqlalchemy-default-datetime#:~:text=One%20other%20potential%20gotcha%3A). harvest record creation is committed per record which i thought meant each harvest record creation had its own transaction but i could be wrong. this problem ends up being unique to the test environment because jobs in production aren't going to occur in fast succession like this where the creation dates are so close ( in this case identical ). nevertheless, it was odd behavior I consider needing to be resolved.

## PR TASKS

- [ ] Any changes to docs?


database/interface.py
```diff
@@ -419,57 +419,86 @@ def update_harvest_record(self, record_id, updates):
...
-        )
-        for record in records:
-            self.db.delete(record)
-        self.db.commit()
-        return "Harvest record deleted successfully"
+    def delete_harvest_record(self, identifier=None, ID=None, harvest_source_id=None):
+        try:
+            # "id" is built-in function. don't reassign.
+            # delete all versions of the record within the given harvest source
+            if harvest_source_id is not None and identifier is not None:
```
btylerburton, line 447: this is a good fix

btylerburton: can you add the logger comment on length of records back in?

harvester/harvest.py
```diff
@@ -739,6 +740,8 @@ def sync(self) -> None:
             if self.action == "update":
                 self.ckanify_dcatus()
                 self.update_record()
+            if self.action is not None:
```
btylerburton: doesn't this just recreate the record we just deleted in 736?

scripts/clean-db.py
```diff
@@ -0,0 +1,26 @@
...
+sys.path.insert(1, "/".join(os.path.realpath(__file__).split("/")[0:-2]))
+
+from harvester import HarvesterDBInterface  # noqa E402
+
+
+def main(days):
+    interface = HarvesterDBInterface()
+    outdated_records = interface.get_all_outdated_records(days)
+    for record in outdated_records:
+        interface.delete_harvest_record(ID=record.id)  # this cascades to errors
```
btylerburton, line 14: thanks for the comment. this will change once we hang errors on the job instead of the record.

tests/integration/harvest/test_ckan_load.py
```diff
@@ -83,7 +83,7 @@ def test_sync(
         assert harvest_source.report["records_errored"] == 0
 
         # NOTE: we don't report this status, but it is not in sync b/c deletes aren't counted correctly
-        assert results["status"]["success"] == 7
+        assert results["status"]["success"] == 8
```
btylerburton: doesn't this mean that the record was not actually deleted?

database/models.py
btylerburton, line 1: this looks like a wholesale improvement, but you also might want to look into freezegun for controlling time in tests

harvester/harvest.py
rshewitt: it would update the record as successfully deleted from ckan. it wouldn't recreate the record. now that i'm thinking about it I should add a check in `delete_self_in_db` to only delete if the status == "success"

btylerburton: based on what i've seen, it will create a new record with status:success and action:delete... 
but, so long as our clear job cleans out all deleted records we won't have an issue

tests/integration/harvest/test_ckan_load.py
rshewitt: for this test there's supposed to be 6 creations, 1 update, and 1 delete. i'm thinking that the number of successes should correspond with the number of operations needing to be done, in this case 8. even if a record is labelled for deletion the status of that record should reflect what happened on ckan. after we've successfully deleted the record on ckan we delete the record in the harvester db.

harvester/harvest.py
rshewitt: does `db_interface.update_harvest_record` create a new record? i thought it would update in place?

tests/integration/harvest/test_ckan_load.py
btylerburton: > after we've successfully deleted the record on ckan we delete the record in the harvester db.

this is the part that's not happening. but it isn't a huge issue we just need to be clear about our approach as a team. we can pair on this, and i'll show you what i'm seeing.

harvester/harvest.py
btylerburton: > does db_interface.update_harvest_record create a new record? i thought it would update in place?

yes it creates a new record if the old record has been deleted.

rshewitt: ahhh i see. the records will be deleted after the update ( referring to your 2nd comment ) so there shouldn't be a problem

rshewitt:
# Deletion approaches

How are harvest records deleted?
- by identifier and harvest source id (>=1 records)
- or record db id (1 record)

## Approach 1
1. [action] on ckan 
2. update db record
3. delete if needed

Notes
- stores historic records (c,u,u,u -> 4 records). 
  - db record count per source record: 0 <= rc <= (1+(u*n))
- <ins>removes all instances of the record INCLUDING the delete</ins>

## Approach 2
1. [action] on ckan
2. delete if needed
3. insert record

Notes
- stores historic records (c,u,u,u -> 4 records). 
  - db record count per source record: 1 <= rc <= (1+(u*n))
- <ins>removes all instances of the record EXCEPT a delete</ins>

database/interface.py
```diff
@@ -419,57 +419,90 @@ def update_harvest_record(self, record_id, updates):
...
+                HarvestRecord.harvest_source_id,
+                desc(HarvestRecord.date_created),
+            )
+            .distinct(HarvestRecord.identifier, HarvestRecord.harvest_source_id)
+            .subquery()
+        )
 
-        fields = list(res.keys())
-        records = res.fetchall()
+        sq_alias = aliased(HarvestRecord, subq)
```
rshewitt, line 502: @btylerburton aliasing allows me to do the following
```sql
# this isn't the exact same query for getting the latest but a minimal example of selecting from a subquery based on a condition
select * 
from ( select * from harvest_record as subq ) # here's the alias ( as [name] )
where subq.action != "delete"
```
using an alias for the subquery allows me to refer to it in the surrounding query which is needed in order to properly get the latest set of records for a harvest source. removing records with an action of "delete" within the subquery would produce wrong results.

given the following records...
`id, identifier, action, date_created`
`1, A, "delete", 03-01-2024`
`2, A, "create", 01-01-2024`

- we expect no records to be returned in preparation for the comparison because the latest state of the record shows it has been deleted. so we want to select those records from the subquery where the action != "delete" 
- if we removed "delete[d]" records within the subquery then the "create" record would be returned which we know is incorrect.

harvester/harvest.py
```diff
@@ -752,6 +750,10 @@ def sync(self) -> None:
 
         if self.action == "delete":
             self.delete_self_in_db()
+        # re: elif - we don't want our record to be
+        # re-created immediately after deletion
+        elif self.action is not None:
+            self.update_self_in_db()
```
rshewitt: @btylerburton i reverted back to how you were doing it. i don't think my change was necessary after thinking about it. frankly, i should have put more thought into what you were doing than defaulting to changing things. here's a [writeup](https://github.com/GSA/datagov-harvester/pull/134#issuecomment-2643501802) of both approaches. seems like they're only slightly different.

database/interface.py
```diff
@@ -419,57 +419,90 @@ def update_harvest_record(self, record_id, updates):
...
             return None
 
-    def delete_harvest_record(self, identifier):
-        records = self.db.query(HarvestRecord).filter_by(identifier=identifier).all()
-        if not records:
-            logger.warning(f"Harvest records with identifier {identifier} not found")
-            return
-        logger.info(
-            f"{len(records)} records with idenitifier {identifier}\
+    def delete_harvest_record(self, identifier=None, ID=None, harvest_source_id=None):
```
btylerburton: minor gripe, but can we use record_id here instead of ID? feels a little more pythonic.

database/interface.py
```diff
@@ -419,57 +419,90 @@ def update_harvest_record(self, record_id, updates):
...
-    def get_latest_harvest_records_by_source(self, source_id):
-        # datetimes are returned as datetime objs not strs
-        sql = text(
-            f"""SELECT * FROM (
-                SELECT DISTINCT ON (identifier) *
-                FROM harvest_record
-                WHERE status = 'success' AND harvest_source_id = '{source_id}'
-                ORDER BY identifier, date_created DESC ) sq
-                WHERE sq.action != 'delete';"""
+    def get_all_outdated_records(self, days):
```
btylerburton: can you a default of days? like 365?

database/interface.py
btylerburton: gotcha thanks!

btylerburton (dismissed):

btylerburton (approved):

PR #141, opened 2025-02-11
Redirect Root Route for Datagov Harvest Admin

btylerburton:
# Pull Request

Related to:
- https://github.com/GSA/data.gov/issues/5060
- https://github.com/GSA/data.gov/issues/5080

## About

Also:
- Improves pagination module
- Adds pagination to harvest source summary page
- Drops pagination default to 10 entries

## PR TASKS

- [x] Code well documented
- [x] Files linted


app/routes.py
```diff
@@ -474,28 +475,33 @@ def add_harvest_source():
...
+        "last_job_finished": "N/A",
+        "next_job_scheduled": "N/A",
+    }
+
     # TODO: wire in paginated jobs htmx refresh ui & route
     jobs = db.pget_harvest_jobs(
         paginate=False, facets=f"harvest_source_id = '{source.id}'"
     )
-    next_job = "N/A"
+    if len(jobs):
```
neilmb: **comment (non-blocking)**: I think that the most idiomatic version of this might be `if jobs:`, or maybe `if len(jobs) > 0`, but it's also fine as-is.

app/routes.py
```diff
@@ -474,28 +475,33 @@ def add_harvest_source():
...
+        last_job = jobs[len(jobs) - 1]
```
neilmb: **comment (non-blocking)**: More idioms, negative indexes count from the end so `last_job = jobs[-1]` works the same as what you have here.

app/routes.py
```diff
@@ -876,13 +882,50 @@ def get_harvest_error(error_id: str = None) -> dict:
...
-    return render_template("get_data_sources.html", sources=source, organizations=org)
+@mod.route("/metrics/", methods=["GET"])
+def view_metrics():
+    """Render index page with recent harvest jobs."""
+    # Get current time in UTC
+    current_time = get_datetime()
+    start_time = current_time - timedelta(hours=24)
+
+    # Format timestamps for SQL query
+    time_filter = f"date_created >= '{start_time.isoformat()}'"
```
neilmb: **question**: This makes my hackles raise just a tiny bit because it feels like we are interpolating strings into SQL statements, even though it's completely safe in this instance. With pure curiosity, are we pretty sure that `pget_harvest_jobs` is safe from SQL injection?

neilmb (dismissed):
LGTM, my comments are just bossy quibbles, nice work.

app/routes.py
btylerburton: this is worth having a chat about sanitization and the safety of these functions in general. would appreciate your thoughts on this.

cmhedrickREI (dismissed):
I think this looks good! Thanks for fixing the broken test cases!

app/routes.py
btylerburton: i think i've addressed most of the hackle-raising stuff. it definitely needed some cleanup. as always, thanks for the questions!

cmhedrickREI (approved):
I think this all looks good! :rocket:

PR #146, opened 2025-02-12
Non paginated decorator wrapper

btylerburton:
Adds a new decorator and usage pattern to allow non-paginated interface queries to access the count decorator...

# Pull Request

Related to [LINK TO ISSUE]

## PR TASKS

- [X] Code well documented
- [X] Files linted

### DIFF
--- a/tests/integration/database/test_db.py
+++ b/tests/integration/database/test_db.py
@@ -279,7 +279,27 @@ class TestDatabase:
             job_id,
             count=True,
         )
-        assert count == len(record_data_dcatus)
+        assert count == len(record_data_dcatus) == 10
+
+    def test_endpoint_count_for_non_paginated_methods(
+        self, interface_with_fixture_json, source_data_dcatus, record_data_dcatus
+    ):
+        interface = interface_with_fixture_json
+        count = interface.get_latest_harvest_records_by_source_orm(
+            source_data_dcatus["id"],
+            count=True,
+        )
+        assert (
+            count
+            == len(
+                [
+                    record
+                    for record in record_data_dcatus
+                    if record["status"] == "success"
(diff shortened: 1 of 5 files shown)


tests/integration/database/test_db.py
```diff
@@ -279,7 +279,27 @@ def test_endpoint_count(
...
             count=True,
         )
-        assert count == len(record_data_dcatus)
+        assert count == len(record_data_dcatus) == 10
+
+    def test_endpoint_count_for_non_paginated_methods(
+        self, interface_with_fixture_json, source_data_dcatus, record_data_dcatus
+    ):
+        interface = interface_with_fixture_json
+        count = interface.get_latest_harvest_records_by_source_orm(
```
neilmb, line 288: **question (non-blocking)**: I'm just trying to understand the point of this. It appears that to get the number of records from one of our query methods, you can just use the same query method but pass a `count=True` argument?

That feels strange to me, but it's just vibes. I guess it saves us from having a bunch of `get_count_...` methods or else having to read all the records and then call `len` on the return?

(Is our ORM lazy enough that calling `len` on a query result doesn't read all of the data?)

neilmb (approved):
This code looks fine, but I'm just trying to understand the usage pattern of the `count=True` argument.

tests/integration/database/test_db.py
btylerburton: We needed a memory efficient way to compute count on results that could grow into the 100k+ range. The more I learn about SQLAlchemy the more I see many ways to do this, but I keep coming back to this scaler math count that I've used in the decorator. Definitely appreciate the question and glad to chat more about alternate ways to do this.

PR #150, opened 2025-02-19
Connect HarvestRecordError to HarvestJob

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5072

## About

- Creates a relationship between `HarvestJob` and `HarvestRecordError`
- Removes the required FK on the `HarvestRecordError` connecting to the `HarvestRecord`
- `HarvestRecordError`s will not be deleted on `HarvestRecord` Deletion
- Because we decouple the `HarvestRecordError` and `HarvestRecord` this impacts the UI for the `/harvest_job/{job_id}` job error table could end up with missing info. So we compensate the missing fields with `N/A` to safely load the table.

Screenshots:
Screenshot from 2025-02-26 12-00-42
_(Screenshot of 2 `HarvestRecordError`s being displayed in the "Harvest Record Errors" Section. The one on top is a `HarvestRecordError` where the `HarvestRecord` still exists in the DB. The `HarvestRecordError` on the bottom shows "N/A" from fields that would be provided from the `HarvestRecord` because it has been deleted.)_
## PR TASKS

- [x] Code well documented
- [x] Files linted


tests/conftest.py
```diff
@@ -36,7 +36,7 @@ def default_session_fixture():
         yield
 
 
-@pytest.fixture(scope="session")
```
cmhedrickREI, line 39: Noting that this was removed to help resolve a race condition. It was discussed maybe to revisit test tear down and setup could potentially resolve the race condition.

cmhedrickREI:
@GSA/data-gov-dev-team Because we are now trying to decouple the `HarvestRecord` and `HarvestRecordError` so that way we can preserve the `HarvestRecordError`s we will be put into a position where we potentially will not have a `HarvestRecord` to pull the `HarvestRecord.identifier` as well as the `HarvestRecord.source_raw` for the Record Error Table within the web UI (the table found on `/harvest_job/<job_id>`). The table needs `HarvestRecord.identifier` and `HarvestRecord.source_raw` for the Record Identifier and Title used in the table respectively. 

The proposal that @btylerburton and I have is that we could add the Identifier from the `HarvestRecord` and then the title from the `source_raw` into the `HarvestRecordError` as attributes in order to preserve this information even when we do delete the `HarvestRecord`. Would this be the best way forward to continue to retain the information we would want to see when looking at the Harvest Record Error table that we display on  `/harvest_job/<job_id>`?

app/routes.py
```diff
@@ -735,8 +735,12 @@ def get_harvest_job(job_id=None):
         record_errors_dict = [
             {
                 "error": db._to_dict(row.HarvestRecordError),
-                "identifier": row.identifier,
-                "title": json.loads(row.source_raw).get("title", None),
+                "identifier": row.identifier if row.identifier else "N/A",
```
neilmb: **comment (non-blocking)**: I like the thinking here, but I think that including placeholder text in the return of this Python function is the wrong place to handle this. Python pretty naturally uses `None` to represent missing data and then in the display layer, we should use `if ... is None` to put placeholder text in at that point.

Does that general idea make sense?

app/templates/view_job_data.html
```diff
@@ -121,4 +125,4 @@ <h3>{{ record_error.error.id }}</h3>
 {% endblock %}
 
 {% block scripts %}
-{% endblock %}
+{% endblock %}
```
neilmb: **quibble (non-blocking)**: This is probably an editor setting, but I think that we want newlines at the end of files and even if we don't want to standardize that, this is a whitespace-only change that dilutes the functional changes in the PR.

database/interface.py
```diff
@@ -364,20 +364,14 @@ def get_harvest_record_errors_by_job(self, job_id: str, **kwargs):
...
-            .filter(HarvestRecord.status == "error")
-            .filter(HarvestRecord.harvest_job_id == job_id)
-            .subquery()
-        )
         query = (
             self.db.query(
                 HarvestRecordError, HarvestRecord.identifier, HarvestRecord.source_raw
             )
-            .join(
+            .outerjoin(
```
neilmb, line 371: **comment**: This is lovely, thanks for using the outer join here.

tests/fixtures.json
```diff
@@ -140,43 +140,51 @@
...
         },
         {
             "harvest_record_id": "95021355-bad0-442b-98e9-475ecd849033",
+            "harvest_job_id": "6bce761c-7a39-41c1-ac73-94234c139c76",
             "message": "record is invalid",
             "type": "ValidationException"
         }
     ]
-}
+}
```
neilmb: as above, can just keep the newline at the end of this file

tests/integration/database/test_db.py
```diff
@@ -616,7 +615,78 @@ def test_delete_outdated_records_and_errors(
         assert db_records == len(latest_records_from_db1) + len(latest_records_from_db2)
 
         db_record_errors = interface.pget_harvest_record_errors(count=True)
-        assert db_record_errors == 0
+
+        # It should be expected the 1 error is still present
+        assert db_record_errors == 1
+
+    def test_delete_job(
```
neilmb: **suggestion (non-blocking)**: I think that this test name isn't as specific as it could be. How about 
```suggestion
    def test_deleting_job_deletes_errors(
```

tests/integration/database/test_db.py
```diff
@@ -616,7 +615,78 @@ def test_delete_outdated_records_and_errors(
...
+        associated HarvestRecord is deleted.
+        """
+        interface = interface_with_fixture_json
+        job_id = job_data_dcatus["id"]
+        # Confirm that errors are created
+        error_count = interface.pget_harvest_record_errors(
+            count=True,
+        )
+        assert error_count == len(record_error_data)
+        # Delete harvest job with errors
```
neilmb: **question (non-blocking)**: I think this comment is copy-pasta and doesn't apply inside this test?

tests/integration/database/test_db.py
```diff
@@ -616,7 +615,78 @@ def test_delete_outdated_records_and_errors(
...
+        )
+
+        # Confirm we have existing records and that they equal
+        # the number from our baseline
+        assert len(harvest_job_records) == len(record_data_dcatus)
+        # Delete HarvestRecords
+        for record in harvest_job_records:
+            interface.delete_harvest_record(record_id=record.id)
+
+        # Confirm no more HarvestRecords exist but we do haev HarvestRecordErrors
```
neilmb:
```suggestion
        # Confirm no more HarvestRecords exist but we do have HarvestRecordErrors
```

neilmb (dismissed):
LGTM, comments are all quibbles. 

The only nearly substantive one is about where to put placeholder text values in returned data. I am advocating keeping `None` in the data at the "python" level and then replacing that with our desired placeholder text in the "display" layer. Those layers aren't necessarily as obvious as in some situations, but that's the principle I'm suggesting.

app/routes.py
btylerburton: Seconded. We follow that pattern in a few other view templates. I appreciate the explanation for  how we apply the rule to define a placeholder vs a pythonic representation of `None`.

database/interface.py
```diff
@@ -364,20 +364,14 @@ def get_harvest_record_errors_by_job(self, job_id: str, **kwargs):
...
         """
-        subquery = (
-            self.db.query(HarvestRecord.id)
-            .filter(HarvestRecord.status == "error")
-            .filter(HarvestRecord.harvest_job_id == job_id)
-            .subquery()
-        )
         query = (
             self.db.query(
                 HarvestRecordError, HarvestRecord.identifier, HarvestRecord.source_raw
```
btylerburton, line 369: Question: if the record does not exist does the query still succeed but just return None for the missing HarvestRecord values?

cmhedrickREI: Indeed it does here is an output I created while testing, where I removed a `HarvestRecord`:
```
# example towards the end
query.all()
[(<HarvestRecordError df0dadf8-5edc-4c12-8afd-e76137227249>, 'test_identifier-2', '{"title": "test-1", "identifier": "test-1"}'), (<HarvestRecordError c9e55bbd-c6dc-4819-86eb-dda8a478c375>, 'test_identifier-3', '{"title": "test-2", "identifier": "test-2"}'), (<HarvestRecordError 1c517b06-840b-4925-b307-6d1b6616a24f>, 'test_identifier-4', '{"title": "test-3", "identifier": "test-3"}'), (<HarvestRecordError 5c85ac04-ca14-41cd-82f9-3a2b9fc84aff>, 'test_identifier-5', '{"title": "test-4", "identifier": "test-4"}'), (<HarvestRecordError b75af616-e790-4d03-b85d-1c62d313fa39>, 'test_identifier-6', '{"title": "test-5", "identifier": "test-5"}'), (<HarvestRecordError 8df27bbc-2d0a-4604-a8d6-2861af390e7a>, 'test_identifier-7', '{"title": "test-6", "identifier": "test-6"}'), (<HarvestRecordError 1beaa99d-0e34-479a-905a-9baa85f97a5e>, 'test_identifier-8', '{"title": "test-7", "identifier": "test-7"}'), (<HarvestRecordError 94ed6eda-5ebf-40ae-adac-8952e13906fe>, None, None)]
```

tests/integration/database/test_db.py
cmhedrickREI: :facepalm:

tests/integration/database/test_db.py
cmhedrickREI: Removed the erroneous comment because it was copy and pasted.

tests/fixtures.json
cmhedrickREI: I went in and manually configured the default vscode JSON and html formatter to not delete new lines at EOF. I also re-added the missing lines.

app/routes.py
cmhedrickREI: I applied the feedback. We should be returning `None` now when we don't have the data. In the templates, it will add the `N/A` when it's receives a `None` value.

app/templates/view_job_data.html
```diff
@@ -92,13 +92,18 @@ <h2>Record Error Details</h2>
                     {% for record_error in data.record_errors %}
                     <div class="error-block">
                         <h3>{{ record_error.error.id }}</h3>
-                        <p><strong>Identifier:</strong> {{ record_error.identifier }}</p>
-                        <p><strong>Title:</strong> {{ record_error.title }}</p>
+                        <p><strong>Identifier:</strong> {{ record_error.identifier if record_error.identifier else "N/A"
```
btylerburton, line 95: After some thought, I think it might be **more** meaningful for the user to see N/A when corresponding records have been deleted, instead of hard coding in the info into the error.

btylerburton (approved):
Excellent work, Christopher!

app/templates/view_job_data.html
cmhedrickREI: Thanks @btylerburton, but I may be confused or misunderstanding the comment. The code here uses a ternary in the template that should render a "N/A" to the user when we don't have the info from the `HarvestRecord` that's tied to the `HarvestRecordError` for that section. If we do have the `HarvestRecord` attached to the `HarvestRecordError` it should then be rendering the desired info for that field currently.

btylerburton: Sorry for not responding sooner, I got busy with other tasks. We had talked earlier about hard-coding in that extra information for identifier and title to prevent them from being absent, but having them be N/A is fine, IMO, b/c it would denote to the user that that record no longer exists.

cmhedrickREI: Absolutely no problem, and no worries here! But whoops this is entirely my fault, I understand now. It had fell out of my head to update the ticket or comment as a reply to the original question on this PR on with the decision. On Friday there was a brief discussion and the approach was decided we would just leave out the information that we pull from the `HarvestRecord` when one does not exist. My apologies! :pray:

btylerburton: Aha! That's what I get for being out on Friday 😁

PR #152, opened 2025-02-21
updates to transformer

rshewitt:
# Pull Request

Related to [#4565](https://github.com/GSA/data.gov/issues/4565)

## About

improve transform test
update transform reader mapping
rename files

## PR TASKS

- [x] Code well documented
- [x] Files linted


rshewitt:
an integration test is failing despite the check passing. the exit code seems to be swallowed by something in the original command because running `poetry run pytest ./tests/integration` locally fails and returns 1 exit code. looks like the culprit is the pipe to coverage text file ( i.e. `tee pytest-coverage-integration.txt` )

jbrown-xentity (dismissed):
These test improvements all look good, thanks @rshewitt !

btylerburton:
Nice fixes. Now what's up with that failing functional test?

rshewitt:
> Nice fixes. Now what's up with that failing functional test?

I was working on a load test with transformation on friday on dev but couldn't get the harvest runner app to run on cf so i ended up deleting it ( after trying a handful of things ) thinking it was in a bad state but i forgot we rely on that app guid so that's why the those tests are failing. working on fixing this

btylerburton:
> couldn't get the harvest runner app to run on cf so i ended up deleting it ( after trying a handful of things ) thinking it was in a bad state 

my guess is you ran into the failing token issue that i've been seeing for awhile now?
https://github.com/GSA/datagov-harvester/pull/136

btylerburton:
but now at least this makes sense, and let's pair on replacing that new app_guid string whenever you get the app recreated

btylerburton (approved):

PR #153, opened 2025-02-24
Adds order_by support to paginated methods;

btylerburton:
Adds util for consolidating properties from list of dicts

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5099

## PR TASKS

- [X] Code well documented
- [X] Files linted

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -553,8 +553,8 @@ def view_harvest_source_data(source_id: str):
         if future_jobs:
             summary_data["next_job_scheduled"] = future_jobs[0].date_created
 
-        chart_data = make_jobs_chart_data(
-            db._to_dict(jobs[::-1]),
+        chart_data_values = make_jobs_chart_data(
+            db._to_dict(jobs[::-1]),  # reverse the list order for the chart
             [
                 "date_finished",
                 "records_added",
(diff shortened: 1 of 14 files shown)


app/routes.py
```diff
@@ -547,30 +553,40 @@ def view_harvest_source_data(source_id: str):
...
+            db._to_dict(jobs[::-1]),
+            [
+                "date_finished",
+                "records_added",
+                "records_deleted",
+                "records_errored",
+                "records_ignored",
+            ],
+        )
         chartdata = {
```
neilmb: **comment (non-blocking)**: These names are confusing, `chart_data` and `chartdata`. Could one or both of them indicate their purpose in the name?

database/interface.py
```diff
@@ -628,35 +603,71 @@ def verify_user(self, usr_data):
     #### PAGINATED QUERIES ####
     @count
     @paginate
-    def pget_harvest_jobs(self, facets="", **kwargs):
+    def pget_harvest_jobs(self, facets="", order_by="asc", **kwargs):
         facet_string = query_filter_builder(None, facets)
-        return self.db.query(HarvestJob).filter(text(facet_string))
+        if order_by == "asc":
```
neilmb: **comment (non-blocking)**: This seems like a lot of code repetition in each of these `pget_` methods to handle ordering. Could we use a helper like

```def _order_by(class, order_by_argument):
    if order_by_argument == "asc":
        order_by_val = class.date_created.asc()
    elif order_by_argument == "desc":
        order_by_val = class.date_created.desc()
    return order_by_val
```

tests/integration/database/test_db.py
neilmb, line 1: **question (non-blocking)**: It's not obvious to me that we added a unit test just for the `order_by` argument behavior. Can we just have a simple test that asserts that `order_by="asc"` gives the reverse of the result of `order_by="desc"`?

neilmb (dismissed):
LGTM: just persnickety comments

jbrown-xentity (commented):
Test coverage improvements; woohoo!

cmhedrickREI (approved):
Looks excellent!

PR #154, opened 2025-02-24
added authorization for api call

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4842

Added the authorization to secure RESTFul API calls with environment variable `FLASK_APP_SECRET_KEY`


app/routes.py
```diff
@@ -59,7 +59,14 @@
 def login_required(f):
     @wraps(f)
     def decorated_function(*args, **kwargs):
-        if "user" not in session:
+        provided_token = request.headers.get("Authorization")
+        if request.is_json or provided_token:
+            api_token = os.getenv("FLASK_APP_SECRET_KEY")
+            if not provided_token or provided_token != f"{api_token}":
```
neilmb: **comment (non-blocking)**: I think that it's better to test that `provided_token is None` for the case where `Authorization` header doesn't exist instead of whether or not it's True as a boolean.

app/routes.py
```diff
@@ -59,7 +59,14 @@
...
+                return "error: Unauthorized"
```
neilmb: **question (non-blocking)**: Does this return value do what we need it to? I'm surprised that simply returning a string accomplishes our desired behavior. Does flask translate this into an HTTP response with an appropriate status code?

neilmb (changes requested):
Where are the tests for this newbehavior? I would like to see some new tests in `tests/integration/app/test_login_required.py` that describe the desired behavior for JSON requests that have 1) correct Authorization headers, 2) incorrect authorization headers, and 3) no authorization headers.

app/routes.py
Jin-Sun-tts: good point. I will add status code.

app/routes.py
```diff
@@ -59,7 +59,14 @@
 def login_required(f):
```
btylerburton, line 64: @Jin-Sun-tts can you add this decorator to any method that is not just a simple GET from the DB?

btylerburton (commented):
I think we need some tests, maybe in `test_routes.py`?

app/routes.py
Jin-Sun-tts: Done.

app/routes.py
Jin-Sun-tts: Done

app/routes.py
Jin-Sun-tts: Done

app/routes.py
```diff
@@ -64,6 +64,16 @@
...
     @wraps(f)
     def decorated_function(*args, **kwargs):
+        provided_token = request.headers.get("Authorization")
+        if request.is_json or provided_token is not None:
+            if provided_token is None:
+                return "error: Authorization header missing", 401
+            api_token = os.getenv("FLASK_APP_SECRET_KEY")
+            if provided_token != api_token:
+                return "error: Unauthorized", 401
+            return "message: success authorized", 200
```
btylerburton: **blocking:** I think we need to return the decorated function like in line 70/80 below

app/routes.py
```diff
@@ -618,6 +634,19 @@ def view_harvest_sources():
 @mod.route("/harvest_source/config/edit/<source_id>", methods=["GET", "POST"])
 @login_required
 def edit_harvest_source(source_id: str):
+    if request.is_json:
+        source = db.get_harvest_source(source_id)
```
btylerburton: I don't think we need this check 638-640 b/c the interface does it already

tests/integration/app/test_routes.py
btylerburton, line 1: ❤️ love the tests!

btylerburton (changes requested):
Just one blocking comment. I may be missing something though.

app/routes.py
neilmb: **question**: Since this is used as a decorator for a function `f`, I don't think that we want to return from this wrapper function here. Don't we need to call `f` and return its return value from this function?

neilmb (dismissed):
The tests look good, thank you. I think that in a future ticket we want to develop a more comprehensive set of "login_required" checks for every API route, but for now, these tests looks good.

app/routes.py
Jin-Sun-tts: removed.

app/routes.py
Jin-Sun-tts: just modified and return the function f

Jin-Sun-tts: good catch, fixed.

btylerburton (approved):

PR #156, opened 2025-02-27
Adds harvest reporter class; cleans up report method; removes need fo…

btylerburton:
…r previous_job_results for sync job; fixes delete count;

updates job in process; runs report when harvest encounters critical error; adds new percentage complete to job table

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5073

## PR TASKS

- [X] Code well documented
- [X] Files linted


neilmb (dismissed):
LGTM, but I'm not sure I completely understand the uses of it, but the code looks solid.

README.md
```diff
@@ -177,12 +177,24 @@ This is a python application, chiefy comprised of files in the `harvester` direc
 
 #### database migrations
 
-When there are database DDL changes, use following to do the database update:
+When altering the db during development, you first want to stamp the db before making any changes to the model.
```
btylerburton, line 180: @cmhedrickREI can you confirm these migration steps are correct? thanks.

neilmb (dismissed):

README.md
cmhedrickREI: Yep those look like the exact steps I followed for both of my migrations I made so far! Excellent idea to cover the steps to do so! :smile: 

I do have one minor question, but in the listing of the steps on creating a migrations do we also want to include the step to run:
`flask db upgrade`

Because I believe that after making the migration, the migration is not yet applied until we either restart the app or run the upgrade command, if I remember correctly?

btylerburton: Let me test that.

btylerburton: Yes, looks like it! Will update.

cmhedrickREI: Thanks for confirming and adding it! :raised_hands:

cmhedrickREI (approved):
Looks awesome! :rocket:

PR #157, opened 2025-02-28
add spatial translation

rshewitt:
# Pull Request

Related to [#5084](https://github.com/GSA/data.gov/issues/5084)

## About
- adds spatial translation of dcatus "spatial" field
  - seeds database with locations [data](https://github.com/GSA/datagov-deploy/raw/71936f004be1882a506362670b82c710c64ef796/ansible/roles/software/ec2/ansible/files/locations.sql.gz)
  - adds geojson validation lib and functionality
  - adds location name translation
  - adds string munging for polygon discovery. copy/pasted from [geodatagov](https://github.com/GSA/ckanext-geodatagov/blob/ac752b30fbd916e9a078d732231edb8f81914d9c/ckanext/geodatagov/logic.py#L445). it doesn't do points only polygons.

## PR TASKS

- [x] Code well documented
- [x] Files linted


jbrown-xentity:
I believe this is ready for review. Would like to merge and run this as a team. I'm curious how AWS RDS will handle the db migration (will everything go normally, or will there be errors), and investigate what is possible in regards to loading the csv file onto the DB.

tdlowden:
@jbrown-xentity flagging.

1. Signed commits fail
2. Feel free to assign reviewers when that's fixed
3. What ticket does this tie to?

jbrown-xentity:
> @jbrown-xentity flagging.
> 
✔️  1. Signed commits fail
✔️  2. Feel free to assign reviewers when that's fixed
> 3. What ticket does this tie to?

https://github.com/GSA/data.gov/issues/5084 (it's in the top of the description).

tdlowden:
@jbrown-xentity ah sorry I had missed that. thanks!

database/README.md
```diff
@@ -0,0 +1,12 @@
...
+The data.gov harvester database is defined in the models.py.
+
+There is an complete table dependancy called `locations`. This
+table is built from the `locations.csv` file in this directory.
+
+To load this file into a database, run the following command
+(assuming psql is setup locally, and database is configured
+and connection is setup with variable `DATABASE_URI` in bash)
+
+./load_csv.sh locations.csv locations
```
btylerburton, line 14: Are we loading this file during CI? If so, I don't see it.
And has this been tested against a large harvest source to see if there any concerns with memory issues due to the file being so large?

database/interface.py
```diff
@@ -520,6 +521,22 @@ def get_latest_harvest_records_by_source_orm(
     def get_latest_harvest_records_by_source(self, source_id):
         return self._to_dict(self.get_latest_harvest_records_by_source_orm(source_id))
 
+    def get_geo_from_string(self, location_name):
+        """get a geometry from the locations table using the location name
+        (e.g. California, New York)"""
+        try:
+            location = (
+                self.db.query(func.ST_AsGeoJSON(Locations.the_geom))
```
btylerburton, line 529: can you tell me more about this `ST_AsGeoJSON` function and where it comes from?

database/models.py
```diff
@@ -169,3 +170,13 @@ class HarvestUser(db.Model):
...
     ssoid = db.Column(db.String(200), unique=True, nullable=True)
+
+
+class Locations(db.Model):
+    __tablename__ = "locations"
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String)
+    type = db.Column(db.String)
+    display_name = db.Column(db.String)
+    the_geom = db.Column(Geometry(geometry_type="MULTIPOLYGON"))
```
btylerburton, line 181: Does CKAN support `MULTIPOLYGON`?

database/README.md
jbrown-xentity: I haven't configured this with ci. It's unnecessary in most cases, because it's a one-time load. In catalog we have something that "spins up" the necessary infrastructure: https://github.com/GSA/catalog.data.gov/blob/main/.github/workflows/commit.yml#L67-L73. If we had something like that, I would probably put this here.
As it is the code will not fail. The table will be there, it will just be empty. So our harvest process won't break if the table is blank.

database/interface.py
jbrown-xentity: It's a PostGIS function: https://postgis.net/
It's enabled in the alembic transition: https://github.com/GSA/datagov-harvester/blob/add-spatial-translation/migrations/versions/bd24037e625a_locations_table_addition.py#L21
This should be done on the migration. PostGIS is installed by default on AWS RDS, you just have to enable it: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.PostGIS.html (we don't need all the other extensions that you could enable; this should be fine. If need be I will add more when we push to development and I'm wrong).

database/models.py
jbrown-xentity: It supports any geojson: https://docs.ckan.org/projects/ckanext-spatial/en/latest/spatial-search/#geo-indexing-your-datasets
I'll build in some logic to handle this use case from the docs, should require a multi-polygon:
```
All indexed geometries should fall within the -180, -90, 180, 90 bounds. If you have polygons crossing the antimeridian (i.e. with longituded lower than -180 or bigger than 180) you'll have to split them across the antimeridian.
```

database/README.md
btylerburton: Aha, so the shell script is to load the table info into the DB, which persists it?

database/interface.py
btylerburton: > PostGIS is installed by default on AWS RDS, you just have to enable it

Do we have an example in code of where we've done that for catalog?

database/README.md
jbrown-xentity: Correct. Any new database, once initialized, will need this table populated. This is the way to "seed" it.

database/interface.py
jbrown-xentity: So the installation happened here: https://github.com/GSA/catalog.data.gov/blob/main/tools/locations-table/install-locations-table.sh. That was a self-contained sql command file, which I didn't prefer because that meant it couldn't be managed by alembic and sqlalchemy.
As to the actual query, it's in the geodatagov extension: https://github.com/GSA/ckanext-geodatagov/blob/main/ckanext/geodatagov/logic.py#L526-L538

jbrown-xentity:
@btylerburton I believe this is ready to go, though I'd like to be available when it gets merged so I can debug any db upgrade problems.

I had to write the geojson wrapping around the meridian from scratch. I tried to document a ton of what the functions are doing, but let me know if more is needed.

@GSA/data-gov-dev-team I would love to review this during office hours sometime if helpful.

jbrown-xentity:
I'm approving this because technically I have the power to do so (@rshewitt created this PR). I however would still like someone from the @GSA/data-gov-dev-team to review if possible... I'll plan to merge tomorrow morning and debug any issues with rollout.

jbrown-xentity (approved):

jbrown-xentity:
Never mind, unsigned commits are a problem. Moving to new PR: https://github.com/GSA/datagov-harvester/pull/157

PR #162, opened 2025-03-07
Add ms iis waf support

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5089

## About

- switches from `os.join` to `urljoin` to allow for more flexible url concat
    - with the current code and this change this lets us support nginx, apache, and iis

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -233,7 +233,7 @@ class HarvestSource:
                     download_file(self.url, ".json")["dataset"]
                 )
             if self.source_type == "waf":
-                self.external_records_to_id_hash(download_waf(traverse_waf(self.url)))
+                self.external_records_to_id_hash(download_waf(traverse_waf(url=self.url, files=[])))
         except Exception as e:
             # ruff: noqa: E501
             raise ExtractExternalException(
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -233,7 +233,7 @@ class HarvestSource:
                     download_file(self.url, ".json")["dataset"]
                 )
             if self.source_type == "waf":
-                self.external_records_to_id_hash(download_waf(traverse_waf(url=self.url, files=[])))
+                self.external_records_to_id_hash(download_waf(traverse_waf(self.url)))
         except Exception as e:
             # ruff: noqa: E501
             raise ExtractExternalException(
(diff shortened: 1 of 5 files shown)


harvester/harvest.py
```diff
@@ -233,7 +233,7 @@ def prepare_external_data(self) -> None:
                     download_file(self.url, ".json")["dataset"]
                 )
             if self.source_type == "waf":
-                self.external_records_to_id_hash(download_waf(traverse_waf(self.url)))
+                self.external_records_to_id_hash(download_waf(traverse_waf(url=self.url, files=[])))
```
neilmb: **question (non-blocking)**: Why do you need to add the `files=[]` to this call? That's the default in the function definition so I'm confused why you need to specify it here.

tests/waf-html-examples/ms-iis-waf.html
```diff
@@ -0,0 +1,19 @@
+<html>
```
neilmb, line 2: **suggestion (non-blocking)**: I can't tell by looking where this HTML file was derived from. Could we include a comment either here in this file or in the `test_traverse_waf_ms_iis` about how we got an example of the HTML response from this particular web server?

neilmb (dismissed):
LGTM. Just two quibbles. Very cool fix to find that `urloin` tweak.

tests/waf-html-examples/ms-iis-waf.html
cmhedrickREI: Can do thats a good idea!

harvester/harvest.py
cmhedrickREI: I actually forgot to change that back, and I should bring this up in pairing actually. The quick answer, is that I had done this to get the test to work.

The complicated answer and some speculation, is that the `session` in the test runner isn't clearing out the the `files` in memory. I was seeing artifacts from the other tests where despite just calling `self.external_records_to_id_hash(download_waf(traverse_waf(self.url)))` it would have contents in `files`. So to ensure `files` was cleared out I explicitly call it with `files=[]`.

cmhedrickREI:
@neilmb Thanks again for that article that explained the reason for the stale data! I just made the updates requested :smile:

neilmb (approved):
Perfect, so glad that fixed it.

PR #165, opened 2025-03-11
Detect Mimetype and Other Information from JSON

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5083

## About

- We now are able to detect the `mimetype` from the `downloadURL`
- We also now populate other pieces of information from given the [Resource Mapping](https://github.com/GSA/ckanext-geodatagov/blob/4510c5be2bb9ecc16de8bae082fef4d970f10f55/ckanext/geodatagov/plugin.py#L59-L282) (`name`, `no_real_name`, and `description` _(based on previous criteria)_) 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/utils/ckan_utils.py
+++ b/harvester/utils/ckan_utils.py
@@ -1,8 +1,8 @@
-import re
-import uuid
 import mimetypes
-from typing import Union, Tuple
+import re
 import urllib
+import uuid
+from typing import Tuple, Union
 
 from harvester.harvest import HarvestSource
 
--- a/harvester/utils/ckan_utils.py
+++ b/harvester/utils/ckan_utils.py
@@ -16,6 +16,7 @@ MAX_TAG_LENGTH = 100
 MIN_TAG_LENGTH = 2
 
 # mapping of file formats and their respective names
+# taken from https://github.com/GSA/ckanext-geodatagov/blob/4510c5be2bb9ecc16de8bae082fef4d970f10f55/ckanext/geodatagov/plugin.py#L59-L282
 RESOURCE_MAPPING = {
     # ArcGIS File Types
     "esri rest": ("Esri REST", "Esri REST API Endpoint"),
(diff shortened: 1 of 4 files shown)


harvester/utils/ckan_utils.py
```diff
@@ -12,6 +15,223 @@
 MAX_TAG_LENGTH = 100
 MIN_TAG_LENGTH = 2
 
+# mapping of file formats and their respective names
```
neilmb, line 18: **suggestion (non-blocking)**: This list is almost unbelievably detailed, it must have come from somewhere. Could this comment also describe where the list came from and why we brought it in here?

harvester/utils/ckan_utils.py
```diff
@@ -248,6 +468,116 @@ def get_email_from_str(in_str: str) -> str:
         return res.group(0)
 
 
+def get_filename_and_extension(resource: dict) -> Tuple[str, str]:
+    """
```
neilmb, line 473: **question (non-blocking)**: If we nabbed this code from somewhere else, should we make a note of it in this docstring?

harvester/utils/ckan_utils.py
```diff
@@ -248,6 +468,116 @@ def get_email_from_str(in_str: str) -> str:
...
+    last_part = split[-1]
+    ending = last_part.split(".")[-1].lower()
+    if len(ending) in [2, 3, 4] and len(last_part) > 4 and len(split) > 1:
+        return last_part, ending
+    return "", ""
+
+
+def change_resource_details(resource: dict) -> None:
+    """
+    Pull the provided file name, format, and description.
```
neilmb, line 494: **suggestion (non-blocking)**: This function works by mutating its dict argument. I think that's significant to be worth mentioning in the docstring. And if this code came in from somewhere else, we should take note of it.

neilmb (dismissed):
LGTM, quibbles, mostly looking for better attribution on the code coming in from CKAN

harvester/utils/ckan_utils.py
cmhedrickREI: That's a good idea, especially since I'm unsure as well as to _why_ we have one so detailed. I'm assuming it's an exact mapping for all supported extensions for CKAN. The recommendation to include it came from the sketch in the ticket pointing to the map's origins: https://github.com/GSA/ckanext-geodatagov/blob/4510c5be2bb9ecc16de8bae082fef4d970f10f55/ckanext/geodatagov/plugin.py#L59-L282
I'll add that link to the comment.

neilmb (approved):

PR #166, opened 2025-03-12
Setup playwright and add basic coverage

btylerburton:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4811

## PR TASKS

- [X] Code well documented
- [X] Files linted


Makefile
```diff
@@ -29,23 +29,33 @@ install-static: ## Installs static assets
...
+test-integration: ## Runs integration tests.
 	poetry run pytest --local-badge-output-dir tests/badges/integration/ --cov-report term-missing --junitxml=pytest-integration.xml --cov=harvester ./tests/integration | tee pytest-coverage-integration.txt
 
-test-functional: ## Runs integration tests. Compatible with dev environment / `make up`
+test-functional: ## Runs functional tests.
 	poetry run pytest --local-badge-output-dir tests/badges/functional/ --noconftest --cov-report term-missing --junitxml=pytest-functional.xml --cov=harvester ./tests/functional | tee pytest-coverage-functional.txt
 
-test: up test-unit test-integration ## Runs all tests. Compatible with dev environment / `make up`
+test-playwright: ## Runs playwright tests.
+	poetry run pytest --local-badge-output-dir tests/badges/playwright/ --cov-report term-missing --junitxml=pytest-playwright.xml --cov=app ./tests/playwright | tee pytest-coverage-playwright.txt
```
neilmb, line 42: **question (non-blocking)**: I know we had a recent issue with tee swallowing error codes. Do we need to use something like `pipefail` here?

neilmb (approved):
This is so nice to see and the existing tests provide nice examples of how to test HTML and JSON responses.

Makefile
btylerburton: calling `SHELL=/bin/bash -o pipefail` at the top of the file cascades for the whole file apparently.

PR #170, opened 2025-03-14
Fix bug for old-spatial being changed.

jbrown-xentity:
old-spatial metadata value was being changed and causing old-spatial and spatial to match. This adds a test case and fixes that bug.

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5084

## About

Weird thing about this. We found that when the DCAT-US file was harvested onto catalog-next, there seemed to be a bug around old-spatial. The old-spatial and spatial fields had the same value (they shouldn't in this test case, old-spatial should read "United States"). It seemed obvious that the field was getting manipulated in the wrong way.

However, when creating a "failing" test (to reproduce this error), I couldn't reproduce it. Tests immediately passed. I tried to make the safe changes I would make had everything worked as I would have expected. And of course I added the tests.

This should be tested out against development, and rerun the [test harvest](https://datagov-harvest-admin-dev.app.cloud.gov/harvest_source/41f2a5f0-cacb-4b41-9f49-07cfcc5287b7) (after clearing) to see if the problem is resolved, or something deeper is going on.

image

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/harvester/utils/ckan_utils.py
+++ b/harvester/utils/ckan_utils.py
@@ -424,8 +424,8 @@ def create_ckan_extras(
                 }
             )
         elif extra == "spatial":
-            data["value"] = translate_spatial(metadata["spatial"])
             output.append({"key": "old-spatial", "value": metadata["spatial"]})
+            data["value"] = translate_spatial(metadata["spatial"])
         else:
             if isinstance(val, list):  # TODO: confirm this is what we want.
                 val = val[0]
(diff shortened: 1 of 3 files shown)


rshewitt:
the reason why this is happening is because we're translating the spatial value twice. first in the flask app from "United States" to the geojson str then again in geodatagov from geojson str into the same thing. the solution is to disable this functionality in geodatagov.

btylerburton (approved):

PR #171, opened 2025-03-18
Add force update job

btylerburton:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5096

## About

- Adds force-update job and tests
- Migrates buttons to form on org and harvest-source detail pages
- Adds test for CKAN export based on mock
- Adds new delete job CLI command
- Updates logic to block org delete until all harvest sources are deleted
- updates CF manifest to bump disk_quota mem after deploy fail
- isolates app creation from db seed in conftest, speeding up test runs

## PR TASKS

- [X] Code well documented
- [X] Files linted

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,9 +44,9 @@ test-functional: ## Runs functional tests.
 test-playwright: ## Runs playwright tests.
 	poetry run pytest --local-badge-output-dir tests/badges/playwright/ --cov-report term-missing --junitxml=pytest-playwright.xml --cov=app ./tests/playwright | tee pytest-coverage-playwright.txt
 
-test: up test-unit test-integration ## Runs all local tests
+test: clean up test-unit test-integration ## Runs all local tests
 
-test-e2e-ci: up load-test-data test-playwright test-functional clean ## All e2e/expensive tests. Run on PR into main.
+test-e2e-ci: re-up test-playwright test-functional clean ## All e2e/expensive tests. Run on PR into main.
 
 test-ci: up test-unit test-integration clean ## All simulated tests using only db and required test resources. Run on commit.
 
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,9 @@ install-static: ## Installs static assets
 	npm install; \
 	npm run build
 
+load-test-data: ## Loads fixture test data
+	docker compose exec app flask testdata load_test_data
+
 test-unit: ## Runs unit tests.
 	poetry run pytest  --local-badge-output-dir tests/badges/unit/ --cov-report term-missing --junitxml=pytest-unit.xml --cov=harvester ./tests/unit | tee pytest-coverage-unit.txt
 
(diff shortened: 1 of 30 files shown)


app/routes.py
```diff
@@ -442,19 +473,20 @@ def edit_organization(org_id):
 
 
 ### Delete Org
-@mod.route("/organization/config/delete/<org_id>", methods=["POST"])
+@mod.route("/organization/config/delete/<org_id>", methods=["GET"])
```
jbrown-xentity: Is there a reason this isn't just `/organization/<org_id>`, and using the `DELETE` method?

app/routes.py
```diff
@@ -698,13 +773,19 @@ def edit_harvest_source(source_id: str):
 
 
 # Delete Source
-@mod.route("/harvest_source/config/delete/<source_id>", methods=["POST"])
+@mod.route("/harvest_source/config/delete/<source_id>", methods=["GET"])
```
jbrown-xentity: Same here; why isn't this a DELETE method?

jbrown-xentity (dismissed):
Lot of changes in this PR not for the specified ticket.
All look good, 2 minor quibbles about URL methods and paths that should probably be discussed and addressed (if necessary) in a follow up PR.

app/routes.py
btylerburton: We invoke that from a route triggered by an HTML form POST:

https://github.com/GSA/datagov-harvester/pull/171/files#diff-f67826701212aab477be0634a23fdcd7ffdfe748b8ce35eb27b8f690d334c732R408

Outside of using FE AJAX, I don't believe we can use DELETE method. If I'm wrong, let me know and I'm glad to update.

app/routes.py
btylerburton: same as above

app/routes.py
btylerburton: TBH, the real enhancement is probably to remove the redirected GET method altogether and just run the DB delete operation via the original POST.

neilmb: Just piling on: since we are in a python method handler in that form post, I don't see any reason to use an HTTTP redirect to perform the deletion. If we want to design an API that includes a delete method, I think that using a DELETE verb on `/organization/<id>` is totally natural.

app/routes.py
```diff
@@ -285,6 +291,18 @@ def cli_remove_harvest_source(id):
         print("Failed to delete harvest source")
 
 
+## Harvet Job Management
+@job.cli.command("delete")
+@click.argument("id")
+def cli_remove_harvest_job(id):
+    """Remove a harvest source with a given id."""
```
neilmb: **comment (non-blocking)**: I think that this docstring is inaccurate because this deletes a harvest job and not a harvest source. Copy-pasta maybe?

neilmb (dismissed):
LGTM, one tiny fix to a docstring.

app/routes.py
btylerburton: only the finest aldente 🍝

Makefile
```diff
@@ -29,6 +29,9 @@ install-static: ## Installs static assets
 	npm install; \
 	npm run build
 
+load-test-data: ## Loads fixture test data
+	docker compose exec app flask testdata load_test_data
```
cmhedrickREI, line 33: I really appreciate this. I say more `make` commands the merrier

cmhedrickREI (dismissed):
Looks excellent!

Makefile
btylerburton: you'll love `make re-up` then. 😄

cmhedrickREI (approved):
Looks good to me!

PR #173, opened 2025-03-19
Update spatial schema

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5055

## About

- Adds validation schema for Json objects within `spatial` for federal schemas (`dcatus1.1: federal`)
    - Anticipated json objects matching the `type` (string) and `coordinates` (array of arrays of numbers _[list of lists of floats]_)
- Adds tests to test for the supported formats for spatial.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/tests/integration/harvest/test_ckan_load.py
+++ b/tests/integration/harvest/test_ckan_load.py
@@ -228,7 +228,6 @@ class TestCKANLoad:
         record = internal_compare_data["records"][0]
         # force in "spatial" attr as string
         record["spatial"] = "United States"
-        harvest_source.validator.is_valid(record)
         assert harvest_source.validator.is_valid(record)
 
     def test_dcatus1_1_federal_validator_success_spatial_object(
--- a/tests/integration/harvest/test_ckan_load.py
+++ b/tests/integration/harvest/test_ckan_load.py
@@ -1,5 +1,7 @@
 from itertools import groupby
 from unittest.mock import patch
+import json
+import pytest
 
 from deepdiff import DeepDiff
 
@@ -7,6 +9,7 @@ from harvester.harvest import HarvestSource, Record
 from harvester.utils.ckan_utils import create_ckan_resources
 from harvester.utils.general_utils import dataset_to_hash, sort_dataset
 
+from jsonschema.exceptions import ValidationError
 # ruff: noqa: E501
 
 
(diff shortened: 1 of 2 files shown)


tests/integration/harvest/test_ckan_load.py
```diff
@@ -207,3 +209,94 @@ def test_create_ckan_resources(self, dol_distribution_json):
...
+
+        # Set up the harvest source
+        harvest_source = HarvestSource(harvest_job.id)
+        # Confirm the correct schema type is used in the example
+        assert harvest_source.schema_type == "dcatus1.1: federal"
+        record = internal_compare_data["records"][0]
+        # force in "spatial" attr as string
+        record["spatial"] = "United States"
+        try:
+            harvest_source.validator.validate(record)
```
rshewitt: a valid record shouldn't throw an exception so couldn't you just validate the record then `assert True` afterwards? this'll throw an assertion error if it's invalid but this test is supposed to evaluate a valid record. same goes for `test_dcatus1_1_federal_validator_success_spatial_object`

btylerburton: you could also patch ValidationError and confirm call_count on the mock is 0

cmhedrickREI: > you could also patch ValidationError and confirm call_count on the mock is 0

@btylerburton I'll take on @rshewitt's approach, I think I was overthinking the test when I did the try and except. The exception would cause the test to fail and we should see the error in the output. But I like the `assert True` after the `validate` because I think it makes the test seem more explicit that the test is over after `validate` is successfully called.

tests/integration/harvest/test_ckan_load.py
```diff
@@ -207,3 +209,88 @@ def test_create_ckan_resources(self, dol_distribution_json):
...
+        dcatus1.1: federal as their schema and contains a "spatial"
+        attribute that isn't a string or json object that meets
+        the defined criteria, fails validation.
+        """
+        interface.add_organization(organization_data)
+        interface.add_harvest_source(source_data_dcatus)
+        harvest_job = interface.add_harvest_job(job_data_dcatus)
+
+        harvest_source = HarvestSource(harvest_job.id)
+        assert harvest_source.schema_type == "dcatus1.1: federal"
```
neilmb, line 269: **question (non-blocking)**: These three tests have the same setup. Is it worth refactoring everything up to the `harvest_source` into a pytest fixture? It's right on the boundary of a rule-of-three, so I can see it either way.

tests/integration/harvest/test_ckan_load.py
```diff
@@ -207,3 +209,88 @@ def test_create_ckan_resources(self, dol_distribution_json):
...
+        harvest_job = interface.add_harvest_job(job_data_dcatus)
+
+        # Set up the harvest source
+        harvest_source = HarvestSource(harvest_job.id)
+        # Confirm the correct schema type is used in the example
+        assert harvest_source.schema_type == "dcatus1.1: federal"
+        record = internal_compare_data["records"][0]
+        # force in "spatial" attr as string
+        record["spatial"] = "United States"
+        harvest_source.validator.validate(record)
```
btylerburton: i was poking around in the validator code and it has a method `is_valid` to return a boolean. that might work better here even though it doesn't reflect what we do in code.

neilmb (dismissed):
LGTM. One honest question about whether to refactor some slight repetition or not. I trust your judgement on what direction to go with it.

tests/integration/harvest/test_ckan_load.py
btylerburton: https://github.com/python-jsonschema/jsonschema/blob/dd1cd339bbcca809970cbbbf01136d865b82e281/jsonschema/protocols.py#L155

tests/integration/harvest/test_ckan_load.py
btylerburton: There should be a pre-existing seeded fixture already if that works:
https://github.com/GSA/datagov-harvester/blob/main/tests/conftest.py#L408

btylerburton (dismissed):

tests/integration/harvest/test_ckan_load.py
cmhedrickREI: Sorry @neilmb and @btylerburton but that was really good feedback so I wanted to update the test cases with the example Tyler listed because I think it's worth keeping it DRY as we can. The new commit may have invalidated the approvals :sweat_smile:

tests/integration/harvest/test_ckan_load.py
```diff
@@ -223,26 +221,19 @@ def test_dcatus1_1_federal_validator_success_spatial_string(
...
-        harvest_source = HarvestSource(harvest_job.id)
+        harvest_source = HarvestSource(job_data_dcatus["id"])
         # Confirm the correct schema type is used in the example
         assert harvest_source.schema_type == "dcatus1.1: federal"
         record = internal_compare_data["records"][0]
         # force in "spatial" attr as string
         record["spatial"] = "United States"
-        harvest_source.validator.validate(record)
-        assert True
+        harvest_source.validator.is_valid(record)
```
neilmb: **question (non-blocking)**: Is this line now redundant? Do we know if the `validate` method does anything different than the `is_valid` call?

neilmb (dismissed):
LGTM, thanks for the DRYing.

tests/integration/harvest/test_ckan_load.py
cmhedrickREI: doh! I forgot to remove those lines from those 2 tests. I only meant to keep `validate` and `is_valid` for the final test because I wanted the assertion for the Exception being raised was done

neilmb (approved):

PR #176, opened 2025-03-25
Test login on routes

btylerburton:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5143

## PR TASKS

- [X] Code well documented
- [X] Files linted

### DIFF
--- a/tests/integration/app/test_routes.py
+++ b/tests/integration/app/test_routes.py
@@ -3,6 +3,11 @@ import re
 
 # ruff: noqa: F401
 
+LOCATION_ENUMS = {
+    "LOGIN": "/login",
+    "NONE": None,
+}
+
 
 class TestDynamicRouteTable:
     """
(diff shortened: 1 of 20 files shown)


tests/integration/app/test_routes.py
```diff
@@ -1,54 +1,141 @@
...
+            "main.callback",
+        ]
+        # provide a simple map for MOST other routes
+        simple_assertion_map = {"200": None, "302": "/login"}
+        # provide a special assertions map for routes which do something special
+        special_assertion_map = {
+            "main.index": {"status_code": 302, "location": "/organizations/"},
+        }
+        for route in client.application.url_map.iter_rules():
+            if route.endpoint not in whitelisted_routes:
```
neilmb: **suggestion (non-blocking)**: Complete quibble, but instead of putting the whole body inside an `if ... not...` you could use `if route.endpoint in whitelisted_routes: continue` to skip the rest of the loop body in those cases.

tests/integration/app/test_routes.py
```diff
@@ -1,54 +1,141 @@
...
+                replacements = [
+                    ("<org_id>", organization_data["id"]),
+                    ("<source_id>", source_data_dcatus["id"]),
+                    ("<job_id>", job_data_dcatus["id"]),
+                    ("<error_type>", "record"),
+                    ("<record_id>", record_data_dcatus[0]["id"]),
+                    ("<error_id>", record_error_data[0]["id"]),
+                ]
+                cleaned_route_rule = route.rule
+                for old, new in replacements:
```
neilmb: **suggestion (non-blocking)**: Probably not worth the squeeze, but there is a clever solution here involving a single call to `re.sub` https://stackoverflow.com/a/765835

neilmb (dismissed):
LGTM, just pedantry.

tests/integration/app/test_routes.py
btylerburton: good pedantry... i like it...

neilmb (approved):

tests/integration/app/test_routes.py
btylerburton: i really enjoy seeing the technical solution, but looks less readable...

PR #178, opened 2025-03-26
Harvest Source Evaluation Script

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5132

## About

Adds a flask command that runs a script to evaluate Harvest Sources. The script should gather the following information:

1. Source ID
2. Source Title
3. Source Type (waf/datajson/waf collection)
4. Org ID
5. Org Title
6. Org State (Active or Inactive)
7. Status Code (of a request to the data source)
8. Metadata Type (ISO-X, DCATUS, etc)
9. Last Modified
10. Server Type

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/app/commands/evaluate_sources.py
+++ b/app/commands/evaluate_sources.py
@@ -1,18 +1,22 @@
-import requests
 import csv
-from typing import Dict, List
 import logging
-from requests.models import Response
 import xml.etree.ElementTree as ET
-from harvester.utils.general_utils import traverse_waf
-from requests.exceptions import Timeout
+from typing import Dict, List
 
+import requests
+from requests.exceptions import JSONDecodeError, Timeout
+from requests.models import Response
+
+from harvester.utils.general_utils import traverse_waf
 
 logger = logging.getLogger(__name__)
 
 CATALOG_SOURCE_URL = "https://catalog.data.gov/api/action/package_search"
 HEADERS = {
-    "User-Agent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/134.0.0.0 Safari/537.36"
+    "User-Agent": (
+        "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 "
+        "(KHTML, like Gecko) Chrome/91.0 Safari/537.36"
+    )
 }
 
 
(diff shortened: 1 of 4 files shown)


harvester/utils/general_utils.py
```diff
@@ -337,3 +337,19 @@ def process_job_complete_percentage(job):
...
+    Original code from https://github.com/ckan/ckanext-spatial/blob/master/ckanext/spatial/harvesters/waf.py#L277
+    """
+    if not server or "apache" in server.lower():
+        return "apache"
+    if "nginx" in server.lower():
+        return "nginx"
+    if "Microsoft-IIS" in server:
+        return "iis"
+    else:
+        return "other"
```
jbrown-xentity: Why not return the raw value of `server` here?

jbrown-xentity (dismissed):
LGTM, one tiny non-blocking quibble!

app/commands/evaluate_sources.py
```diff
@@ -0,0 +1,270 @@
...
+                                url=xml_file_url, headers=HEADERS, timeout=10
+                            )
+                            if xml_response and xml_response.ok:
+                                if xml_response.headers.get("Last-Modified"):
+                                    timestamps.append(
+                                        xml_response.headers.get("Last-Modified")
+                                    )
+                                metadata_type = determine_metadata_type(xml_response)
+                                # if None, it's a type we don't currently work with.
+                                if metadata_type is not None:
```
btylerburton: Should you be returning `None` from `determine_metadata_type` instead of `N/A` to fix this check?

tests/integration/app/test_evaluate_sources.py
```diff
@@ -0,0 +1,334 @@
...
+        "mock_writer_class": mock_writer_class,
+        "mock_writer_instance": mock_writer_instance,
+    }
+    # Stop patches
+    open_patch.stop()
+    dictwriter_patch.stop()
+
+
+@pytest.fixture
+def mock_catalog_request():
```
btylerburton: where do we use this fixture?

btylerburton (dismissed):
If you can think of a good failure scenario, that would be a nice test to add. Otherwise, looks good!

harvester/utils/general_utils.py
cmhedrickREI: @jbrown-xentity I think that's a good question. I had originally left it as is from the original code, but looking deeper at the original code it seems that we don't really rely much on "other". So I'll make an update to return the raw value, because I feel it adds a good amount of value.

jbrown-xentity (approved):
This looks good to me!

PR #179, opened 2025-03-28
add tag splitting

rshewitt:
# Pull Request

Related to [#5140](https://github.com/GSA/data.gov/issues/5140)

## About

add tag splitting so we get more tags from keywords

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/harvester/utils/ckan_utils.py
+++ b/harvester/utils/ckan_utils.py
@@ -20,6 +20,16 @@ MIN_TAG_LENGTH = 2
 
 db = HarvesterDBInterface()
 
+
+def trim_tag(tag):
+    # deal with something like
+    # EARTH   SCIENCE > ATMOSPHERE > ATMOSPHERIC ELECTRICITY > ATMOSPHERIC CONDUCTIVITY
+    # Truncate individual keywords to 100 characters since DB fields is varchar 100
+    trimmed = re.split(r"[;,>]", tag)
+    trimmed = [t.lower().strip() for t in trimmed]
+    return set([" ".join(t.split())[:100] for t in trimmed if t != ""])
+
+
 # mapping of file formats and their respective names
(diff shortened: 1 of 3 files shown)


harvester/utils/ckan_utils.py
```diff
@@ -456,8 +466,10 @@ def create_ckan_tags(keywords: list[str]) -> list:
     output = []
 
     for keyword in keywords:
-        output.append({"name": munge_tag(keyword)})
-
+        for tag in trim_tag(keyword):
+            data = {"name": munge_tag(tag)}
+            if data not in output:
+                output.append(data)
     return output
```
rshewitt, line 473: looks like tags are sorted on dataset pages. idk if that's done in the template or if we want to do it here. either way i don't want to do it more than once.

tests/conftest.py
```diff
@@ -835,3 +835,14 @@ def mock_get(url, *args, **kwargs):
 
     # Apply the patch using monkeypatch
     monkeypatch.setattr(requests, "get", mock_get)
+
+
+@pytest.fixture
+def dcatus_keywords():
+    return [
+        "EARTH SCIENCE > BIOSPHERE > ECOSYSTEMS > MARINE ECOSYSTEMS > COASTAL",
```
neilmb: **question (non-blocking)**: You put special handling in `trim_tag` to condense multiple spaces in a single tag into one. Do you want to make one of these into a test case for that special handling?
```suggestion
        "EARTH         SCIENCE > BIOSPHERE > ECOSYSTEMS > MARINE ECOSYSTEMS > COASTAL",
```

neilmb (dismissed):
Looks fine to me, small query, but the test case is nice.

tests/conftest.py
rshewitt: i was making lint happy (line too long) and got carried away with the uniformity of the spacing and forgot about that. thanks for this!

neilmb (approved):

PR #184, opened 2025-04-08
Check duplicate identifier

Jin-Sun-tts:
Related to (https://github.com/GSA/data.gov/issues/5039)

Added validation to detect and report duplicate identifiers in harvested external records.


tests/integration/harvest/test_ckan_load.py
```diff
@@ -274,3 +274,25 @@ def test_dcatus1_1_federal_validator_fails(
...
+        interface.add_harvest_source(source_data_dcatus)
+        job = interface.add_harvest_job(job_data_dcatus)
+
+        harvest_source = HarvestSource(job.id)
+        harvest_source.prepare_external_data()
+        record = list(harvest_source.external_records.values())[0].metadata
+
+        # Attempt to load the same record again (simulate duplicate)
+        with pytest.raises(ExtractExternalException) as exc_info:
+            harvest_source.external_records_to_id_hash([record])
```
rshewitt: we're never going to be processing external records twice in 1 harvest so this isn't a realistic test case. do you mind creating a dcatus fixture which actually contains duplicate identifiers? then you can just wrap your `pytest.raises` around line 290.

Jin-Sun-tts: sure, Done.

rshewitt (approved):
looks good! thanks for that change @Jin-Sun-tts

PR #190, opened 2025-04-09
only get json when expected

rshewitt:
# Pull Request

related to [5176](https://github.com/GSA/data.gov/issues/5176) 
@btylerburton this fixes the issue with the job throwing an exception on record transformation

## About

mdtranslator crashed because of out-of-memory. this obviously caused an issue with transformation in the harvest runner which led me to finding out about this.  we should only be calling `.json()` if the response was okay.

## PR TASKS

- [ ] Code well documented
- [x] Files linted

### DIFF
--- a/tests/integration/harvest/test_transform.py
+++ b/tests/integration/harvest/test_transform.py
@@ -1,3 +1,5 @@
+from unittest.mock import patch
+
 import pytest
 from deepdiff import DeepDiff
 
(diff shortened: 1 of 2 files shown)


jbrown-xentity (dismissed):
Looks good.

tests/integration/harvest/test_transform.py
```diff
@@ -82,3 +84,32 @@ def test_valid_transform_iso19115_docs(
...
+        harvest_source = HarvestSource(harvest_job.id)
+        harvest_source.extract()
+        harvest_source.compare()
+
+        for record in harvest_source.records:
+            if record.identifier == "http://localhost:80/iso_2_waf/invalid_iso2.xml":
+                test_record = record
+
+        with patch("requests.post") as mock_get:
+            mock_get.return_value.status_code = 404
```
rshewitt, line 108: i scaled mdtranslator to 0 instances, posted to it, got a 404. i assume this would be the case if the app crashed/was down?

rshewitt: ran `cf stop mdtranslator` and confirmed i get a 404. figure it'd be the same if it crashed.

jbrown-xentity (approved):
Looks like good improvements, with more clarity for when issues arise.

PR #192, opened 2025-04-14
#4964 Script to migrate harvest sources from CKAN to H2.0

neilmb:
# Pull Request

This adds a script to migrate harvest sources from CKAN into Harvester. To assist in the process, it adds a JSON output for the `/organization/<organization_id>` route using a `to_dict` method from our model's base class.

## About

Also includes some logging fixes in `database/interface.py` to make debugging easier inside the running harvest-admin.

After this is merged, someone should run the script against the actual dev harvest admin instance.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


app/static/_scss/styles.scss
```diff
@@ -1,10 +1,3 @@
-@use "uswds-core" with (
-    $theme-show-notifications: false,
-    $theme-color-base-light: "gray-cool-5",
-    $theme-color-primary: "blue-70v",
-    $theme-color-primary-vivid: "blue-60v",
-    $theme-color-primary-dark: "blue-warm-70v",
-);
-
+@forward "uswds-theme";
```
btylerburton: was there another `uswds-theme` file you meant to include?

database/models.py
```diff
@@ -10,6 +10,9 @@ class Base(DeclarativeBase):
     __abstract__ = True  # Indicates that this class should not be created as a table
     id = Column(String(36), primary_key=True, default=lambda: str(uuid.uuid4()))
 
+    def to_dict(self):
```
btylerburton, line 13: that's a really nice helper method! we can probably clean up the other one in the interface in favor of this one eventually.

scripts/ckan_migration/migrate_harvest_sources.py
```diff
@@ -0,0 +1,219 @@
...
+    def _source_type(source):
+        if source["source_type"].startswith("waf"):
+            return "waf"
+        return "document"
+
+    def _schema_type(source, org):
+        # TODO: can we actually determine this now?
+        organization_types = [
+            extra["value"]
+            for extra in org.get("extras", [])
```
btylerburton: i don't see where you're getting the org extras. am i missing the `organization_show` call?

database/interface.py
```diff
@@ -148,7 +148,7 @@ def update_organization(self, org_id, updates):
                 if hasattr(org, key):
                     setattr(org, key, value)
                 else:
-                    print(f"Warning: non-existing field '{key}' in organization")
+                    logger.warning("Warning: non-existing field '%s' in organization", key)
```
cmhedrickREI: Replacing all the `print`s with actual `logger` makes me happy :) Thanks for doing this.

app/static/_scss/styles.scss
neilmb: These are unrelated changes, let me get them out of this PR...

scripts/ckan_migration/migrate_harvest_sources.py
neilmb: I forgot, adding that now, also putting in the notification_emails which come from that too.

app/routes.py
```diff
@@ -472,7 +479,7 @@ def view_organization(org_id: str):
             "view_org_data.html",
             data=data,
             form=form,
-        )
+        ), (200 if org is not None else 404)
```
btylerburton, line 482: **💭 thought:** ‏I'm curious what the UI behavior is here once we send a status code. I initially coded an if/else into the template but this might be a better approach...

database/interface.py
```diff
@@ -80,20 +81,11 @@ def _impl(self, *args, **kwargs):
...
-                pool_timeout=60,
-                pool_recycle=1800,
-            )
-            session_factory = sessionmaker(bind=engine, autoflush=True)
-            self.db = scoped_session(session_factory)
-        else:
-            self.db = session
+        # Flask-SQLAlchemy provides a request-scoped database session
+        # so use it here
+        self.db = db.session
```
btylerburton, line 88: **👏 praise:** ‏If this works long-term, I will be very happy! 😂

scripts/ckan_migration/migrate_harvest_sources.py
```diff
@@ -0,0 +1,278 @@
...
+        "frequency": source["frequency"].lower(),
+        "source_type": mapper["source_type"](source),
+        "schema_type": mapper["schema_type"](source, org),
+        "notification_emails": mapper["notification_emails"](source, org),
+        # no equivalent in CKAN, choose a sane default
+        "notification_frequency": "always",
+    }
+
+
+@cache
```
btylerburton, line 179: **👏 praise:** great use of cache here!‏

btylerburton (approved):

app/routes.py
neilmb: I actually like the if/else in the template, but I thought that this status code goes well with the "custom" error page produced by the template. I'm also happy to take this out since I think I only depend on the JSON response giving a non-200 status code, but this seemed more consistent with that.

btylerburton: i think it's a good improvement. we'll see what it serves in the UI and make a small improvement to the 404-template if necessary.

PR #194, opened 2025-04-15
add dcatus email prefix to pass validation

rshewitt:
# Pull Request

Related to [5190](https://github.com/GSA/data.gov/issues/5190)

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


tests/integration/harvest/test_ckan_load.py
```diff
@@ -123,7 +123,7 @@ def test_ckanify_dcatus(
             "author": None,
             "author_email": None,
             "maintainer": "Harold W. Hild",
-            "maintainer_email": "hhild@CFTC.GOV",
+            "maintainer_email": "mailto:hhild@CFTC.GOV",
```
btylerburton, line 126: If I'm not mistaken, one issue I see with this is some sources already provide a `mailto:` prefix.

rshewitt: we only pull the email right [now](https://github.com/GSA/datagov-harvester/blob/2fad23a53debd15bb7968f77f3ff92ba4c006028/harvester/utils/ckan_utils.py#L491-L494) so even if they pass `mailto:` we should still be good?

harvester/utils/ckan_utils.py
btylerburton, line 1: https://github.com/GSA/datagov-harvester/pull/194/files#diff-439db4eb35a7e886e2da6be8766f5e454b5b77b5751975e677aebe1e6fee5a27R647-R648

do these need to be updated?

harvester/utils/ckan_utils.py
```diff
@@ -663,6 +663,7 @@ def simple_transform(metadata: dict, owner_org: str) -> dict:
             for k2, v2 in v.items():
                 if k2 == "hasEmail":
                     v2 = get_email_from_str(v2)
+                    v2 = "mailto:" + v2
```
btylerburton, line 670: Is this `mailto:` validation constraint derived from the example here: https://resources.data.gov/resources/dcat-us/#contactPoint-hasEmail?

btylerburton (approved):
Looks good overall!

I'm concerned that appending to datasets with data quality issues will incur more data quality issues, but we won't know until we see that happen, I suppose... 🤷🏼

harvester/utils/ckan_utils.py
rshewitt: yeah should swap maintainer with author.

harvester/utils/ckan_utils.py
rshewitt: it's one of the regex patterns in the schema. either 
`"^mailto:[\\w\\_\\~\\!\\$\\&\\'\\(\\)\\*\\+\\,\\;\\=\\:.-]+@[\\w.-]+\\.[\\w.-]+?$"` 
or 
`"^(\\[\\[REDACTED).*?(\\]\\])$"`

probably none of the datasets from [onms](https://data.noaa.gov/waf/NOAA/nos/onms/iso/xml/) will pass validation because of this (i.e. none of the emails have `mailto:`). i'll add a helper function to evaluate the email and add "mailto:" to make it validate

PR #196, opened 2025-04-16
Add script to compare dataset counts

Jin-Sun-tts:
Related: https://github.com/GSA/data.gov/issues/4758

Add script to compares the number of datasets per organization between catalog and catalog-next

### DIFF
--- a/scripts/eval-harvest-state/compare_datasets_count.py
+++ b/scripts/eval-harvest-state/compare_datasets_count.py
@@ -40,6 +40,15 @@ DEFAULT_OUTPUT_JSON_NEXT = "catalog_next_orgs.json"
 DEFAULT_OUTPUT_CSV = "catalog_orgs_comparison.csv"
 
 def fetch_all_organizations(base_url):
+    """
+    Fetch all organizations and their dataset counts from a catalog.
+
+    Args:
+        base_url (str): Base URL of the catalog.
+
+    Returns:
+        list: A list of organization dictionaries with dataset counts.
+    """
     organizations = []
     limit = 25
     offset = 0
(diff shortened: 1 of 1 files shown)


scripts/eval-harvest-state/compare_datasets_count.py
```diff
@@ -0,0 +1,165 @@
...
+
+def save_orgs_to_json(orgs, filename):
+    with open(filename, "w", encoding="utf-8") as jsonfile:
+        json.dump(orgs, jsonfile, indent=2)
+
+def load_orgs(json_path):
+    with open(json_path, "r", encoding="utf-8") as file:
+        return json.load(file)
+
+def compare_orgs(catalog_orgs, catalog_next_orgs, output_csv_path):
```
neilmb, line 107: **suggestion (non-blocking)**: Even with a good name like this, a docstring would be nice, especially one that explains any return values or side-effects like files that might be written.

scripts/eval-harvest-state/compare_datasets_count.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            raise Exception(f"Failed to fetch organizations from {base_url}")
+
+    return organizations
+
+def save_orgs_to_json(orgs, filename):
+    with open(filename, "w", encoding="utf-8") as jsonfile:
+        json.dump(orgs, jsonfile, indent=2)
+
+def load_orgs(json_path):
```
neilmb: **suggestion (non-blocking)**: Docstring? It can make things easier to read if you re-use argument names. The previous function has "filename", could this be called "filename" as well?

scripts/eval-harvest-state/compare_datasets_count.py
```diff
@@ -0,0 +1,165 @@
...
+            if not batch:
+                break
+            organizations.extend(batch)
+            offset += limit
+        else:
+            raise Exception(f"Failed to fetch organizations from {base_url}")
+
+    return organizations
+
+def save_orgs_to_json(orgs, filename):
```
neilmb, line 80: **suggestion (non-blocking)**: docstring

neilmb (dismissed):
LGTM but I would like to see more docstrings.

scripts/eval-harvest-state/compare_datasets_count.py
Jin-Sun-tts: Done

scripts/eval-harvest-state/compare_datasets_count.py
Jin-Sun-tts: Done

scripts/eval-harvest-state/compare_datasets_count.py
Jin-Sun-tts: Done

Jin-Sun-tts:
> LGTM but I would like to see more docstrings.

Added more Docstrings. Thank you @neilmb

neilmb (approved):
Very nice docstrings now, thank you.

PR #198, opened 2025-04-17
Create orgs in catalog-next as well

neilmb:
# Pull Request

GSA/data.gov#4964 also needs the new organizations that are being migrated into harvester to be created in catalog-next. This extends the migration script to do that as well with arguments for the second target URL and a second API key.

Also added a README file to hopefully make the actual using of the script more understandable.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


scripts/ckan_migration/README.md
btylerburton, line 1: **👏 praise:** hoooray for docs!‏

scripts/ckan_migration/migrate_harvest_sources.py
```diff
@@ -262,11 +325,18 @@ def migrate_source(api_token, limit=0, debug=False, harvester_url=None):
     if not CONFIG.harvester_url.endswith("/"):
```
btylerburton, line 325: should we add the same default override for this as you have for `CONFIG.new_catalog_url`?

scripts/ckan_migration/migrate_harvest_sources.py
```diff
@@ -237,19 +273,46 @@ def upload_sources(sources):
     markers.
     """
     click.echo("Uploading sources...")
-    with ThreadPoolExecutor() as executor:
+    with ThreadPoolExecutor(max_workers=CONFIG.workers) as executor:
```
btylerburton, line 276: what happens if `CONFIG.workers` is not set?

neilmb: It's set unconditionally in the main script with a default of `None` when the argument isn't given which gives the default behavior in `ThreadPoolExecutor`.

scripts/ckan_migration/migrate_harvest_sources.py
neilmb: It's there, it just off the top of this diff view

scripts/ckan_migration/migrate_harvest_sources.py
btylerburton: I see it now!

btylerburton (approved):
🚢

PR #200, opened 2025-04-17
Refactors db_session creation

btylerburton:
- Refactors db_session creation so harvest process creates its own engine outside flask context;
- updates stuck jobs when cancelled even when they're not currently running; 
- marks failed translations as invalid and gates them from creating secondary validation errors; 
- adds new full flow test for WAF collection

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5193

## PR TASKS

- [X] Code well documented
- [X] Files linted

### DIFF
--- a/database/interface.py
+++ b/database/interface.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from datetime import datetime, timezone
 from functools import wraps
 
@@ -21,7 +20,6 @@ from .models import (
     db,
 )
 
-DATABASE_URI = os.getenv("DATABASE_URI")
 PAGINATE_ENTRIES_PER_PAGE = 10
 PAGINATE_START_PAGE = 0
 
@@ -81,11 +79,14 @@ def count_wrapper(fn):
 
 
 class HarvesterDBInterface:
-
     def __init__(self, session=None):
-        # Flask-SQLAlchemy provides a request-scoped database session
-        # so use it here
-        self.db = db.session
+        if session:
+            ## For the Harvest Runner we create our own session and pass it in
+            self.db = session
+        else:
+            # Flask-SQLAlchemy provides a request-scoped database session
+            # so use it here
+            self.db = db.session
 
     @staticmethod
     def _to_dict(obj):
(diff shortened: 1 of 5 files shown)


cmhedrickREI (dismissed):

jbrown-xentity (dismissed):
All seem like good improvements.
Someday, when we are less busy, it would be nice to have documentation on how the db interface and sessions work, since we have a single code base that supports multiple application types (harvest admin, harvest UI, harvest runner, etc).

database/interface.py
```diff
@@ -81,11 +79,14 @@ def _impl(self, *args, **kwargs):
 
 
 class HarvesterDBInterface:
-
     def __init__(self, session=None):
-        # Flask-SQLAlchemy provides a request-scoped database session
-        # so use it here
-        self.db = db.session
+        if session:
```
neilmb: **suggestion (non-blocking)**: When you intend to test whether or not an argument was passed, prefer `if session is not None:` because passed arguments might inadvertently be "falsy" and fail this check.

harvester/lib/load_manager.py
```diff
@@ -82,7 +82,8 @@ def stop_job(self, job_id, job_type="harvest"):
             return f"No task with job_id: {job_id}"
 
         if job_task[0][1] != "RUNNING":
-            return f"Harvest job {job_id} is not running"
+            updated_job = interface.update_harvest_job(job_id, {"status": "complete"})
```
neilmb, line 85: **suggestion (non-blocking)**: Since we immediately return after this, the `updated_job` return value is never used in a log message like it is in the later `update_harvest_job` call.
```suggestion
            interface.update_harvest_job(job_id, {"status": "complete"})
```

neilmb (dismissed):
LGTM, just style quibbles.

harvester/lib/load_manager.py
btylerburton: good catch. i updated it to use the return val.

rshewitt (approved):
all good additions!

PR #203, opened 2025-04-24
Use internal routes for MDTranslator

neilmb:
# Pull Request

#5179 asks for an access-control solution for our MDTranslator service. I set up an internal route `mdtranslator-development.apps.internal` for MDTranslator and removed the old `mdtranslator.app.cloud.gov` route. Then, adjusting the `MDTRANSLATOR_URL` environment variable is all that is needed to use the new route.

## About

The staging version of this is aspirational: that route does not exist and nor does the application. When we start deploying in new environments, we have to do the equivalent of

```bash
$ cf map-route mdtranslator apps.internal --hostname mdtranslator-staging
$  cf add-network-policy datagov-harvest-admin mdtranslator
$  cf add-network-policy datagov-harvest-runner mdtranslator
```

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/vars.development.yml
+++ b/vars.development.yml
@@ -10,7 +10,7 @@ CKAN_API_URL: https://catalog-next-dev-admin-datagov.app.cloud.gov
 
 HARVEST_RUNNER_APP_GUID: 40dba74b-4916-447e-914b-77525de5e608
 
-MDTRANSLATOR_URL: https://mdtranslator.app.cloud.gov/translates
+MDTRANSLATOR_URL: http://mdtranslator-development.apps.internal:8080/translates
 
 # Login.gov
 CLIENT_ID: urn:gov:gsa:openidconnect.profiles:sp:sso:gsa:datagov-dev-harvest-admin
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
Better to have it private, so other people can't slow down are harvester...
Looks good!

vars.development.yml
```diff
@@ -10,7 +10,7 @@ CKAN_API_URL: https://catalog-next-dev-admin-datagov.app.cloud.gov
 
 HARVEST_RUNNER_APP_GUID: 40dba74b-4916-447e-914b-77525de5e608
 
-MDTRANSLATOR_URL: https://mdtranslator.app.cloud.gov/translates
+MDTRANSLATOR_URL: http://mdtranslator-development.apps.internal:8080/translates
```
btylerburton, line 13: Do you know if this affects how we invoke the service?

```
mdt_url = os.getenv("MDTRANSLATOR_URL")
resp = requests.post(mdt_url, json=data)
```
https://github.com/GSA/datagov-harvester/blob/parellize-ckan-network-calls/harvester/harvest.py#L798-L799

neilmb: Requests transparently handles the URL with port apparently, because it is working right now in development with this exact setting.

btylerburton: That's what I like to hear!

PR #204, opened 2025-04-24
Parallelize CKAN network calls

btylerburton:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5199

## PR TASKS

- [X] Code well documented
- [X] Files linted


harvester/__init__.py
```diff
@@ -15,13 +16,7 @@
...
 
 # create a scopedsession for our harvest runner
-engine = create_engine(
-    DATABASE_URI,
-    pool_size=10,
-    max_overflow=20,
-    pool_timeout=60,
-    pool_recycle=1800,
-)
+engine = create_engine(DATABASE_URI, poolclass=NullPool)
```
neilmb: Suggestion (non-blocking): Maybe a comment here linking to their recommendation to use a NullPool for this situation?

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
 
 ROOT_DIR = Path(__file__).parents[1]
 
+# harvest worker count
+harvest_worker_sync_count = int(os.getenv("HARVEST_WORKER_SYNC_COUNT"))
```
neilmb: Suggestion (non-blocking): If the env variable is missing, this will give an error I think. Can we use a safe default as the second argument to getenv?
```suggestion
harvest_worker_sync_count = int(os.getenv("HARVEST_WORKER_SYNC_COUNT", 1))
```

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
...
+
+
+class CKANSyncTool:
```
neilmb: **suggestion (non-blocking)**: There's something that feels strange about using a class here in this file just to package related functions that never save data on any instance. I don't have such a huge problem with it, but isntead I might have used a python module `ckan_sync.py` with these five functions in it and in this file just `import ckan_sync` and use `ckan_sync.sync(record)` in the futures call.

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
...
+            record: An instance of HarvestRecord (Record) class
+
+        Returns:
+            True (bool): on record success
+
+        Raises:
+            DCATUSToCKANException: Any DCAT-US to CKAN metadata transform issue
+            SynchronizeException: Any ckan sync issue
+        """
+        if record.status == "success" or record.status == "error":
```
neilmb: **question (non-blocking)**: This check is reasonable here, but can we save ourselves a few function calls by filtering the records we pass in here using `...for record in records if record.status not in ("succes", "error")]`

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
...
+                repr(e), record.harvest_source.job_id, record.id
+            )
+
+        # begin ckanpi sync proper
+        start = get_datetime()
+        try:
+            if record.action == "delete":
+                self.delete_record(record)
+            elif record.action == "create":
+                self.ckanify_record(record)
```
neilmb: question (non-blocking)**: Didn't we just do this  in the pre-sync step? Is there some reason we need to do it again?

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
...
+                res = self.create_record(record)
+                record.ckan_id = res["id"]
+                record.ckan_name = record.ckanified_metadata["name"]
+            elif record.action == "update":
+                self.ckanify_record(record)
```
neilmb: **question**: As above, I think we already did this in the pre-sync phase?

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
...
+
+        # # update harvest job
+        record.harvest_source.db_interface.update_harvest_job(
+            record.harvest_source.job_id, record.harvest_source.reporter.report()
+        )
+
+        return True
+
+    def create_record(self, record, retry=False) -> dict:
+        from harvester.utils.ckan_utils import add_uuid_to_package_name
```
neilmb: **question (non-blocking)**: It feels weird to be importing inside the inner function here. Is this import particularly expensive or this function called so infrequently that we don't want to just import it at the top of the file?

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
...
+
+    def update_record(self, record) -> dict:
+        updated_metadata = {
+            **record.ckanified_metadata,
+            **{"id": record.ckan_id, "name": record.ckan_name},
+        }
+        return ckan.action.package_update(**updated_metadata)
+
+    def delete_record(self, record) -> None:
+        return ckan.action.dataset_purge(**{"id": record.ckan_id})
```
neilmb: suggestion (non-blocking)**: I think that with just one keyword argument, it's just as easy to not use the dict unwrapping:
```suggestion
        return ckan.action.dataset_purge(id=record.ckan_id)
```

harvester/harvest.py
```diff
@@ -57,6 +58,126 @@
...
+
+    def ckanify_record(self, record) -> None:
+        from harvester.utils.ckan_utils import ckanify_dcatus
```
neilmb: **question (non-blocking)**: As above, is there a reason to be doing this import inside a function that it going to be called frequently instead of at the top of the file?

harvester/harvest.py
```diff
@@ -352,23 +473,43 @@ def validate(self) -> None:
...
+            try:
+                future.result()
             except (
                 DCATUSToCKANException,
                 SynchronizeException,
             ):
                 self.reporter.update("errored")
 
+        with ThreadPoolExecutor(max_workers=harvest_worker_sync_count) as executor:
+            futures = [
```
neilmb: **suggestion (non-blocking)**: We don't actually need to store the futures here, we can call `add_done_callback` on the return value of `submit` inside the loop
```suggestion
            [executor.submit(ckan_sync.sync, record).add_done_callback(error_callback) for record in self.records]
```
(eh, the suggestion isn't totally right, sorry)

neilmb (dismissed):
This is great work. I've got more than my usual curmudgeonly comments, but they are  all stylistic, so your call on what to incorporate.

harvester/harvest.py
btylerburton: Thank you!

harvester/harvest.py
btylerburton: Thanks for the reminder. That was my initial plan and then I got too busy cleaning up the tests.

harvester/harvest.py
btylerburton: I deleted the pre_sync and combined them back together.

harvester/harvest.py
btylerburton: Same comment as above.

harvester/harvest.py
btylerburton: This formerly had a naming conflict collision, but now that I've renamed the method the import should be able to be lifted.

harvester/harvest.py
btylerburton: Oh nice. I'll take a look at that.

harvester/harvest.py
btylerburton: Aha, had to take a look at my own code to see that I had hoisted this up above the actual sync task. Thanks for reminding me of my own good intentions. It's been a long decade since last night. 😂 / 😭

harvester/harvest.py
btylerburton: Same oversight 🙈

harvester/harvest.py
btylerburton: 🐐 moving this class to ckan_utils made this problem go away

harvester/harvest.py
btylerburton: 🐐 moving this class to ckan_utils made this problem go away

neilmb (approved):

PR #205, opened 2025-04-29
add iso dcatus schema based off non-federal

rshewitt:
# Pull Request

Related to [#5215](https://github.com/GSA/data.gov/issues/5215)

## About
- add separate dcatus schema for transformed iso docs

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


tests/integration/harvest/test_extract.py
```diff
@@ -40,3 +39,18 @@ def test_extract_dcatus(
...
+        source_data_waf_iso19115_2,
+        job_data_waf_iso19115_2,
+    ):
+        interface.add_organization(organization_data)
+        interface.add_harvest_source(source_data_waf_iso19115_2)
+        harvest_job = interface.add_harvest_job(job_data_waf_iso19115_2)
+
+        harvest_source = HarvestSource(harvest_job.id)
+
+        assert str(harvest_source.schema_file).endswith("iso-non-federal_dataset.json")
```
neilmb, line 56: **question (non-blocking)**: Can we also make a test where a dataset either passes or fails to pass the new validation schema? Are there particular places where the schema we use for ISO-transformed catalogs differs from the other schemas, and if so, should we make test cases to ensure that we don't break those with future changes?

neilmb (dismissed):
The code looks fine, but there's a larger question about if and how we want to test the behavior of these different schemas. I just don't know what our current status or system is for testing schemas.

tests/integration/harvest/test_extract.py
rshewitt: Up to this point the only tests we have on validation are for dcatus fed & non-fed (not transformed) and the one included in this PR. The only difference between the fed & non-fed schemas is the federal schema requires `bureauCode` and `programCode`. currently, there's no difference between the iso_non-fed & non-fed but the intention is to loosen the rules for iso_non-fed as-needed (and within reason) in order to get records onto ckan. i can include 2 tests in this PR for passing/non-passing transformed ISO docs.

harvester/harvest.py
```diff
@@ -102,14 +102,18 @@ def __post_init__(self) -> None:
...
+        self.schemas_root = ROOT_DIR / "schemas"
+
         if self.schema_type == "dcatus1.1: federal":
-            self.dataset_schema = open_json(
-                ROOT_DIR / "schemas" / "federal_dataset.json"
-            )
+            self.schema_file = self.schemas_root / "federal_dataset.json"
+        elif self.schema_type == "dcatus1.1: non-federal":
+            self.schema_file = self.schemas_root / "non-federal_dataset.json"
+        elif self.schema_type.startswith("iso19115"):
```
btylerburton, line 111: We say that we support CSDGM here: https://github.com/GSA/datagov-harvester/blob/main/database/models.py#L63

Should we remove that?

harvester/harvest.py
```diff
@@ -102,14 +102,18 @@ def __post_init__(self) -> None:
...
         self.get_source_info_from_job_id(self.job_id)
 
+        self.schemas_root = ROOT_DIR / "schemas"
+
         if self.schema_type == "dcatus1.1: federal":
-            self.dataset_schema = open_json(
-                ROOT_DIR / "schemas" / "federal_dataset.json"
-            )
+            self.schema_file = self.schemas_root / "federal_dataset.json"
+        elif self.schema_type == "dcatus1.1: non-federal":
```
btylerburton: We can remove this elif if it's falling into `non-federal` in the else

tests/integration/harvest/test_extract.py
btylerburton: +1 for more validation tests. I think a good guideline is to explicitly assert any parts of the loosened spec that differ from non-federal.

harvester/harvest.py
rshewitt: i think csdgm removal should be its own ticket. i can create that. have we heard back from csdgm data providers?

harvester/harvest.py
rshewitt: if i do that i feel the need to write a comment indicating which schema's we support fall into the "else" category and at that point is it much different than explicitly checking self.schema_type == "dcatus1.1: non-federal"? i suppose raising an exception on the else might be good

btylerburton: Fair point. How about you add a INFO log with the source ID to call out that we're not hitting on any of the expected conditionals. Do we even need an else block then, is it a helpful catch-all?

rshewitt: If we're not hitting any of the conditions i would consider this a critical error because we're unable to validate the records. that's what prompted me to think raising an exception in the else block might be good.

btylerburton: Oh, sure! That makes sense. I didn't exactly follow your logic before.

harvester/harvest.py
```diff
@@ -106,12 +106,13 @@ def __post_init__(self) -> None:
...
             self.schema_file = self.schemas_root / "federal_dataset.json"
-        elif self.schema_type == "dcatus1.1: non-federal":
+        elif self.schema_type in ["dcatus1.1: non-federal", "csdgm"]:
             self.schema_file = self.schemas_root / "non-federal_dataset.json"
         elif self.schema_type.startswith("iso19115"):
             self.schema_file = self.schemas_root / "iso-non-federal_dataset.json"
         else:
-            self.schema_file = self.schemas_root / "non-federal_dataset.json"
+            # this can't happen because we apply an enum in our model but just in case.
+            raise Exception
```
btylerburton, line 120: Where does this bubble up to? Should we populate the exception with any job specific info? 
I'd also recommend adding a logger.error here, too.

tests/conftest.py
btylerburton, line 1: oooh nice this fixes some issues with tests i'm seeing break now that i have the new mdtranslator image locally.

tests/integration/harvest/test_validate.py
```diff
@@ -0,0 +1,87 @@
...
+
+        harvest_source = HarvestSource(harvest_job.id)
+        harvest_source.prepare_external_data()
+
+        iso2_name = "http://localhost:80/iso_2_waf/valid_iso1.xml"
+        iso2_test_record = harvest_source.external_records[iso2_name]
+        iso2_test_record.transform()
+
+        # validator throws an exception when the dataset is invalid
+        with pytest.raises(ValidationException):
```
btylerburton: #

**🎯 suggestion (non-blocking):** It would be nice to also assert against the message of the validation error.

harvester/harvest.py
rshewitt: i suppose it would bubble up to the `SystemExit` exception in the sort-of main function we have at the bottom of the script? I can add a log message.

btylerburton: If that's the case then we will also want to update the job with `status:error`

rshewitt: done

btylerburton (approved):
Looks good. Great to see those validation messages!

PR #206, opened 2025-04-29
Harvest UI updates

btylerburton:
# Pull Request

- Makes JSON faceting function apply to jobs, job_errors and record_errors
- Renames some methods for consistency
- Adds links to harvest source in a job UI, and org in a source UI
- Updates metrics page with new job id column

## PR TASKS

- [X] Code well documented
- [X] Files linted


app/routes.py
```diff
@@ -1150,6 +1124,43 @@ def view_metrics():
...
+            page=request.args.get("page", type=convert_to_int),
+            per_page=request.args.get("per_page", type=convert_to_int),
+            paginate=request.args.get("paginate", type=is_it_true),
+            count=request.args.get("count", type=is_it_true),
+            order_by=request.args.get("order_by"),
+            facets=facets,
+        )
+        if not res:
+            return f"No {model} found for this query", 404
+        elif isinstance(res, int):
```
neilmb, line 1157: **question (non-blocking)**: I can't figure out here why `res` would just be a number. Can you maybe include a comment in here what this case could result from and why we want to return a very short text response rather than anything JSON?

app/templates/metrics_dashboard.html
```diff
@@ -53,24 +58,6 @@ <h2>Recent Harvest Jobs</h2>
...
-                                        <a href="{{ url_for('main.view_harvest_job', job_id=job.id) }}"
-                                            class="usa-button usa-button--outline">
-                                            View
-                                        </a>
-                                    </li>
-                                    {% if job.status == 'in_progress' %}
-                                    <li class="usa-button-group__item">
-                                        <a href="{{ url_for('main.cancel_harvest_job', job_id=job.id) }}"
-                                            class="usa-button usa-button--secondary">
-                                            Cancel
```
neilmb, line 68: **question (non-blocking)**: I haven't used this enough to know if this will be fine for actual admin use, but is the intention now to click through to the job to cancel?

docs/routes.md
neilmb, line 1: **question (non-blocking)**: We changed some of the API routes for `edit`, do we want to keep updating this documentation to match?

tests/integration/app/test_routes.py
```diff
@@ -254,3 +255,39 @@ def test_get_missing_organization(
         )
         assert res.status_code == 404
         assert res.is_json
+
+    @pytest.mark.parametrize(
+        "route,status_code,response",
+        [
+            (
+                "/harvest_records/?harvest_source_id=2f2652de-91df-4c63-8b53-bfced20b276b",
```
neilmb, line 263: **question (non-blocking)**: I can't figure out where the quantities for the various types of records come from here since I don't see them in a mock right here. Can we have a comment to point at where these particular "magic numbers" came from?

neilmb (dismissed):
Beautiful work here. My only questions are around additional comments and documentation.

app/templates/metrics_dashboard.html
btylerburton: Yes. Once things are wired in correctly we shouldn't be cancelling too much. We can easily add it back in as long as we wrap it in a "is-user-logged-in" check similar to how it is done on the jobs page.

docs/routes.md
btylerburton: yes. thanks for the reminder.

app/routes.py
btylerburton: We can pass in `count=True` and the db will response with just an int. Historically we've just responded with a string, but I think I'll make it a real JSON response.

tests/integration/app/test_routes.py
btylerburton: They're seeded by the fixture.json. A comment to that effect can't hurt though.

btylerburton (commented):
Thanks for the great feedback, Neil! I'll make those changes now.

app/routes.py
```diff
@@ -415,7 +416,7 @@ def add_organization():
 
 
 @main.route("/organizations/", methods=["GET"])
```
btylerburton: I didn't initially pull in `/organizations/` and `/harvest_sources/` into the json_builder_query method as they were already in use by our UI list, but the more I think about it the more I think it's a valid change.

So I'm planning to remap: 

`/organizations/` -> `/organization_list/`
`/harvest_sources/` -> `/harvest_source_list/`

`/organizations/` & `/harvest_sources/` -> will now be used in json_builder_query

neilmb (approved):

PR #207, opened 2025-05-02
Log duplicate identifier

Jin-Sun-tts:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5039

introduces a new non-critical exception, DuplicateIdentifierException, to handle duplicate harvest record identifiers gracefully during processing. Instead of halting the harvest job, duplicates are now:
- Logged at the record level in the harvest_record_error table.
- Associated with a minimal placeholder harvest_record, created solely to satisfy foreign key constraints.
- Allowed to continue processing, ensuring only valid, unique records are added to the harvest.


neilmb (approved):
Looks great, thanks for the very clear test cases.

PR #223, opened 2025-05-14
Notes on scheduling many harvests

neilmb:
# Pull Request

We learned something about how to use our own API to schedule many harvest jobs, so I wanted to write down what we learned.

### DIFF
--- a/docs/ops/README.md
+++ b/docs/ops/README.md
@@ -1,0 +1,32 @@
+# Operations notes
+
+This is miscellaneous notes on operating the Harvester system.
+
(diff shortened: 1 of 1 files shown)


cmhedrickREI (approved):
excellent documentation :tada:

btylerburton:
Great documentation! You might also consider throttling back that concurrent harvesting count to a much smaller number (maybe 5?) if it's causing issues: https://github.com/GSA/datagov-harvester/blob/main/harvester/lib/load_manager.py#L14

PR #224, opened 2025-05-15
Fix `source_raw` return for `harvest_record/{record_id}/raw`

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5216

## About

- Makes it so the `harvest_record/{record_id}/raw` API endpoint still returns JSON for DCAT, and original XML for XML sources.

## PR TASKS

- [x] Code well documented
- [x] Files linted


app/routes.py
```diff
@@ -1017,10 +1017,11 @@ def get_harvest_record_raw(record_id=None):
     record = db.get_harvest_record(record_id)
     if record:
         try:
+            # if this fails, it's not JSON, but possibly XML
             source_raw_json = json.loads(record.source_raw)
             return source_raw_json, 200
         except json.JSONDecodeError:
-            return {"error": "Invalid JSON format in source_raw"}, 500
+            return record.source_raw, 200
```
neilmb: **question (non-blocking)**: It looks like this will just return the raw string as the response body. That can be surprising for API callers who are parsing the responses as if they are JSON objects (like we do with errors). Do we want to wrap this in a JSON object with a single key?
```suggestion
            return {"raw": record.source_raw}, 200
```

It's still surprising because it'll have a different schema than the previous `source_raw_json` response, but at least it'll be a parseable JSON object.

tests/integration/app/test_routes.py
```diff
@@ -315,3 +321,65 @@ def test_json_builder_query(
...
+        harvest_source.compare_sources()
+        harvest_source.write_compare_to_db()
+
+        iso2_name = "http://localhost:80/iso_2_waf/valid_iso2.xml"
+        iso2_test_record = harvest_source.external_records[iso2_name]
+        response = client.get(f"/harvest_record/{iso2_test_record.id}/raw")
+
+        assert response.status_code == 200
+        assert response.text == iso2_test_record.source_raw
+        assert response.json is None
```
neilmb: If the above suggestion is taken, then this needs to change to match

```suggestion
        assert response.json["raw"] == iso2_test_record.source_raw
```

neilmb (dismissed):
The tests are so beautiful! One question with no obvious answer, but the code looks good either way.

app/routes.py
cmhedrickREI: I think this is a good idea and we should follow this route. That way we keep everything as JSON responses so that way anyone who hits the endpoint they get the expected result.

neilmb (approved):
Tests pass, looks great. I think that unifying the return schema under the same "raw" key looks right. We don't have any consumers of this yet, so changing the structure of the response for when we have data.json shouldn't be a problem.

PR #230, opened 2025-05-19
Dont fail the test job if coverage upload fails

neilmb:
# Pull Request

By design, Dependabot doesn't have a Github token with write-access so it can't upload test coverage results on its PRs. We don't need to fail the job just because we can't upload the test results. This changes the workflow so that success on the coverage upload is not required.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
This is true, this probably shouldn't be blocking. It's a nice to have.

PR #231, opened 2025-05-19
Set max runner tasks from environment variable

neilmb:
# Pull Request

GSA/data.gov#5255 asks for the maximum number of harvest runner tasks to be configurable in an environment variable. This adds that code and sets baseline values in our manifest and vars files.


cmhedrickREI (approved):
Looks good and straightforward to me :+1:

jbrown-xentity (approved):
This looks good to me!

PR #232, opened 2025-05-19
configure playwright

rshewitt:
sleep after clean; install just chrome, its sys depends, and only its shell

# Pull Request

Related to [#5241](https://github.com/GSA/data.gov/issues/5241)

## About
- add sleep after clean to (hopefully) avoid race conditions in DDL after cleanup
- only use chromium shell for e2e tests to slim down memory usage in action. chromium is the default [browser](https://playwright.dev/python/docs/test-runners#cli-arguments) for playwright in the python implementation. 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


neilmb (dismissed):
Looks reasonable, hope it works!

.github/workflows/e2e.yml
```diff
@@ -48,7 +48,7 @@ jobs:
         poetry env use ${{ env.PY_VERSION }}
         poetry install
     - name: Install Playwright Browsers
-      run: poetry run playwright install
+      run: poetry run playwright install --with-deps --only-shell chromium
```
rshewitt, line 51: do we want to run our tests against other browsers? (e.g. webkit & firefox). we haven't done this so far which i why I'm only installing chromium here but maybe it was assumed that all installed browsers would run?

Makefile
```diff
@@ -74,6 +74,7 @@ down: ## Tears down the flask and harvester containers
 clean: ## Cleans docker images
 	docker compose down -v --remove-orphans
 	docker compose -p harvest-app down -v --remove-orphans
+	sleep 5
```
btylerburton, line 79: have you tried adding a custom `sleep5` command and then shimming that in between `up` and `load-test-data`

neilmb (approved):
Glad to see these pass, let's merge it and move on.

PR #238, opened 2025-05-22
Count needs to be an integer even coming from environment

neilmb:
# Pull Request

Fixes a bug that James found that has been killing harvest runner.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
Looks safe. If we pass bad arguments to this, shame on us.

PR #240, opened 2025-05-22
Return RAW of the source

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5216

## About

- Returns the actual raw of the file as the respective JSON or XML

## PR TASKS

- [x] Code well documented
- [x] Files linted


neilmb (approved):
Thanks I hate it, but it is apparently the right thing to do.

PR #241, opened 2025-05-25
Add arch for cleanup and retry

jbrown-xentity:
Adds logic for "retry" job (picking up where left off). Adds logic for "cleanup" jobs/tasks (updating DB if job is orphaned, errors, etc).

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5261

## About

This is initially for discussion. We should evaluate if the logic here makes sense.
This attempts to codify how to manage cleanup in our harvest system. How can we retry if a job fails mid-flight out of our control? At what point is it safe to retry? At what point should we mark a job as "failed" and stop retrying?

Note: we are considering re-vamping our understanding of some of our processes. Originally we determined that the "compare" would have to run to completion before taking on the next steps (ie we have to know all the objects and if they need to be added, updated, or deleted before starting). However we can consider an "incremental" approach. If the evaluation of the records existence in a WAF is much less than the pull of the complete record, then we can get the records to be deleted early without getting the full content. Once we have that list, every object we process through comparison that will need to be added or updated is something discreet that we can do; whether we get the full list or not is irrelevant. So only the delete list of the compare has to run to completion before a job is safe to "retry". Then each object is saved in the DB as it is "compared" and then a retry can come back and process those records if memory or other interruption causes a failure.

## PR TASKS

- [x] Code well documented
- [ ] ~Files linted~


jbrown-xentity:
@GSA/data-gov-dev-team can anyone take a look at these and validate if they are ready to merge?

cmhedrickREI (approved):
This looks correct to me!

PR #248, opened 2025-06-02
Deployment scripts

neilmb:
# Pull Request

This sets up the Github actions to manage the staging and prod environments of the Harvest admin app. It also includes a big change getting rid of the separate `datagov-harvest-runner` app and renaming the admin app to just `datagov-harvest`, including a URL change to `datagov-harvest-dev.app.cloud.gov`. As part of the change, we no longer have to keep around a `GUUID` for the runner app, we can just look up the GUUID for the current running CF application.

## About

This is joint work that started with @btylerburton and then was mostly finished by @Jin-Sun-tts and I just pushed it the last step over the line.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


.github/workflows/deploy.yml
```diff
@@ -93,59 +100,56 @@ jobs:
...
+        command: [
+          "cf add-network-policy datagov-harvest mdtranslator --protocol tcp --port 8080"
+        ]
+        environ: ["staging", "prod"]
+    steps:
+      - name: proxy to web
+        uses: cloud-gov/cg-cli-tools@main
         with:
-          filename: .github/deploy_failure.md
-          update_existing: true
```
jbrown-xentity: What's the intention behind removing this feature? We have used this in the past to be notified and automatically create issues for triage...

bin/scale_calculate.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+MAX_INSTANCES=9
+MIN_INSTANCES=5
+SCALE_STEP=2
+
+CPU_BUSY_THRESHOLD=300
+CPU_IDLE_THRESHOLD=220
+
+app_status=$(cf app catalog-web)
```
jbrown-xentity: This is for scaling harvest admin, right? Do we want catalog-web hardcoded here?

jbrown-xentity (dismissed):
I like these improvements! I'm going to approve, but I think we should address the comments listed here...

.github/workflows/deploy.yml
neilmb: I'll put it back, not sure why someone took it out.

bin/scale_calculate.sh
neilmb: Yes, this script is called from the Restart job's `check-and-renew` script but we need to change the app name here. Good catch.

neilmb:
@jbrown-xentity re-review after my fixes when you have a chance?

jbrown-xentity (approved):
LGTM, thanks for the changes!

PR #251, opened 2025-06-03
Create secrets service too if it doesn't exist

neilmb:
# Pull Request

One more deployment fix, another service needed to be created.

## About

This creates the secrets service, but does not add any of the secrets to it. Manually running `cf uups -p ...` will still be required.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
It may not have the necessary info, but at least things won't crash!

Issue #249, opened 2025-06-03
Deployment Failure


Jin-Sun-tts:
Workflow with Issue: 2 - Deploy
Job Failed: deploy-staging
Last Commit: 4ebe8e5a048b054b5348cf1af4f8cb2957eb7b20
Number of times run: 2
Last run by: jbrown-xentity
Github Action Run: https://github.com/GSA/datagov-harvester/actions/runs/16321622992

FuhuXia:
resolved.

PR #252, opened 2025-06-05
Fix app for staging and prod

neilmb:
# Pull Request

This is a bunch of changes to get the staging and prod harvest apps up and running (see #249). 

The largest change is around database migrations. We had two problems: we were doing `db.create_all()` on every run of the app to force SQLAlchemy to create the database schema (which doesn't work with column changes) and we had not been keeping our model changes synced up with the files in our migrations directory. This changes both things. First, we stop running `create_all` and in every invocation (docker and cloud.gov) we call `flask db upgrade` before we start our flask app. This is safe because if all migrations have been applied that command does nothing, and if they have not applied, then they need to be before the app can start. Second, I re-started our chain of migrations with a single migration file that creates our existing schema. Now is the best time to do this as we head towards staging and prod databases that we will want to keep around, but it will mean some work to get things running right on development with the current database, or dropping the current dev database and starting over.

There are also some simplifications in `.profile` because we no longer have multiple app names&mdash;the only app is called `datagov-harvest`&mdash;and a recently-introduced bug was fixed in the `cf_handler.py` with using the `v3` CF API's method for getting the tasks of a specific app.

## About

This is a big change to some of our infrstructure but 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


cmhedrickREI (approved):
Looks excellent! Also thank you for making the base migration :+1:

PR #253, opened 2025-06-05
Set organization_type when we import harvest sources / organizations

neilmb:
# Pull Request

GSA/data.gov#5259 asks that our migration from the current catalog into the new harvester should also migrate the organization's `organization_type` field which tells us whether it is federal, state, etc. In order to keep the harvester database as the source of truth, we add an `organization_type` field to the `Organization` model. 

## About

The migration script does not update existing organizations, so the actual import will need to be done into an empty catalog-next and datagov-harvester database.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


cmhedrickREI (dismissed):
Looks excellent!

neilmb:
@cmhedrickREI Re-review after I fixed the test failures?

database/models.py
```diff
@@ -30,6 +30,7 @@ class Organization(db.Model):
 
     name = db.Column(db.String, nullable=False, index=True)
     logo = db.Column(db.String)
+    organization_type = db.Column(db.String)
```
jbrown-xentity: Do we maybe want to specify the "current" types, so as to give what typical usage of this is? Or some other type of definition?

jbrown-xentity (dismissed):
LGTM!

cmhedrickREI (dismissed):

neilmb:
@jbrown-xentity Now with an enumerated set of possible values, re-review?

database/models.py
neilmb: Now enumerated in the latest commit.

cmhedrickREI (approved):
I think the using an Enum was an excellent enhancement to this PR :+1:

PR #254, opened 2025-06-06
Bugfix/schedule next job

jbrown-xentity:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5266

## About

Attempts Failing tests first to replicate issues, and then addressing the fix.
This should keep us from creating more than 1 harvest job per source into the future (but manual jobs can still be created).
It also keeps jobs from being started if another job in the same source is starting.

## PR TASKS

- [x] Code well documented
- [x] Files linted


cmhedrickREI (dismissed):
looks excellent :+1:

harvester/lib/load_manager.py
```diff
@@ -51,20 +51,42 @@ def start(self):
...
-        """task manager start interface,
-        takes a job_id"""
-        task_contract = {
-            "command": f"python harvester/harvest.py {job_id} {job_type}",
-            "task_id": f"harvest-job-{job_id}-{job_type}",
-        }
+        try:
+            """Check if a job is already running for this source."""
+            harvest_job = interface.get_harvest_job(job_id)
+            jobs_in_progress = interface.pget_harvest_jobs(
```
neilmb, line 69: **suggestion (non-blocking)**: There is a wrapper for this query method where if you include `count=True` as a keyword argument, it will return only the number of results, instead of having to fetch all of the rows. (See https://github.com/GSA/datagov-harvester/blob/main/database/interface.py#L57).

It would more efficient I think to use that here.

harvester/lib/load_manager.py
```diff
@@ -51,20 +51,42 @@ def start(self):
...
+                facets=f"harvest_source_id = '{harvest_job.harvest_source_id}',\
+                    status = 'in_progress'",
+                paginate=False,
+            )
+            if len(jobs_in_progress):
+                return f"Can't trigger harvest. Job {jobs_in_progress[0].id} already in progress."  # noqa E501
```
neilmb: **question (non-blocking)**: This is an internal method rather than an API method. I'm not sure that returning a string with a message explaining what happened is useful in that case.

Do we want to raise an exception if it fails? Or return an error type that can be programmatically checked? And document the possible return values in the docstring?

harvester/lib/load_manager.py
```diff
@@ -112,18 +134,27 @@ def schedule_next_job(self, source_id):
...
+            logger.info("No job scheduled for manual source.")
+            return "No job scheduled for manual source."
+
+        # check if there is a job already scheduled in the future
+        future_jobs = interface.get_new_harvest_jobs_by_source_in_future(source_id)
+        if len(future_jobs) > 0:
+            message = f"Job already scheduled for source {source_id} at \
+            {future_jobs[0].date_created}."
+            logger.info(message)
+            return message
```
neilmb, line 157: **comment (non-blocking)**: As above, for an internal method, I don't think that returning a string explanation is the most helpful return value.

neilmb (dismissed):
This is fine, but I think that the error message returns aren't optimal. I would suggest using various exceptions that can be caught in the callers, or specific error codes that can be checked programatically by the callers.

harvester/lib/load_manager.py
jbrown-xentity: Good call. Limiting it to 1, as that's all I need to know if a new job should be created or not.

harvester/lib/load_manager.py
jbrown-xentity: Throwing an Exception here makes the most sense, as it's handled by the rest of the code. Good call.

harvester/lib/load_manager.py
jbrown-xentity: This message is "flashed" for the user when the source is newly created and the first job scheduled, or when changing the frequency of a source. This is still relevant I think. If we want to have a broader discussion about our formality when passing errors, messages, and objects between functions I think that's a great addition, but probably out of scope for this ticket as it's sprinkled throughout.

harvester/lib/load_manager.py
jbrown-xentity: After review, I backed off throwing an exception.
The try except block in this function does a retry, and then eventually changes the job to created "now" (in theory this should already be in the past, as we only pickup jobs that are past their start date). However since this is to block any ability to start a job when another one is ongoing, it would have the unintended affect of changing a job scheduled for the future to "now". I only caught this because our tests check for it.
I considered changing the test to match the behavior, but it's not how I would expect the function to perform (if for some reason a "start_job" is run for a job scheduled for the future, I wouldn't expect it to change the job schedule).
I considered changing the behavior, but decided that may have unintended consequences. I believe this was put in place because we wanted a "back-off" retry; if it failed to create a task we wanted it to try again later.
I think this is the best solution, but would like @neilmb to weigh in.

cmhedrickREI (approved):
Everything looks good!

PR #257, opened 2025-06-09
Create services should wait while the DB is being created

neilmb:
# Pull Request

In pairing we discovered that our development deploy from scratch failed because the services had not been created when the app tried to start. This makes the database service creation happen last and wait until it succeeds. All of the other service creations should succeed by then too.


jbrown-xentity (approved):
LGTM

PR #259, opened 2025-06-10
Fix for JSON objects in CKAN extras

neilmb:
# Pull Request

GSA/data.gov#5124 found that CKAN cannot create a Solr index for packages with extras whose values are JSON objects or lists. We fixed it in ckanext-datajson, but we also have to fix it in the harvester. This applies the same fix as GSA/ckanext-datajson#161 by serializing extras values that are `dict` or `list` type. There were a couple changes to the test suite to match this behavior.

## About

Because this matches the current behavior of Catalog even more than it used to, I think that the change is worthwhile, although the inclusion of multiple bureauCodes, programCodes, and keywords in these extras seems like it could be a big change for how datasets appear in Catalog-next.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


tests/integration/app/test_routes.py
```diff
@@ -1,6 +1,7 @@
 import json
 import os
 import re
+from unittest.mock import patch
```
cmhedrickREI: Not blocking, just  a minor question: Are there any un-committed tests that were meant to be added? This patch seems to be extraneous.

neilmb: It was extraneous. I'm honestly not sure where it came from. I just made this branch from `develop` and it was there compared to `main`. Linting didn't complain, so maybe `patch` is actually used in this file? Let me learn more...

neilmb: Resolved while I was fixing the tests.

cmhedrickREI (approved):
Thanks for investigating the import!

PR #260, opened 2025-06-12
use gevent instead of gthread

neilmb:
# Pull Request

In GSA/data.gov#5279 we were seeing problems with SSL errors with our database connections. We had changed our Gunicorn workers from `gevent` to `gthread` which appears to have had a conflict with how we set up our database connections. This PR moves us back to `gevent` workers.

## About

The `gevent` workers do give warnings on startup about "late monkey-patching". This is related to running the app under NewRelic. NewRelic support says that these warnings are false positives and no problems will be caused by doing it this way <https://support.newrelic.com/s/hubtopic/aAX8W0000008aE0/seeing-monkeypatching-error-while-running-newrelicadmin-with-python-37-geven>. 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvest-admin-start.sh
+++ b/harvest-admin-start.sh
@@ -9,4 +9,4 @@ if [ "$CF_INSTANCE_INDEX" = "0" -o -z "$CF_INSTANCE_INDEX" ]; then
     flask db upgrade 
 fi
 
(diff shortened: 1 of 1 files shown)


jbrown-xentity (approved):
Let's test this out!

jbrown-xentity (approved):
I'm seeing somewhat inconsistent results. On a particular harvest source page, I found 3 error pages in 50 refreshes. In using the metrics page, I've refreshed over a hundred times and not seen a single error. So it's possible this fixed it (I know this is already pushed to dev). Let's deploy and verify in production...

PR #261, opened 2025-06-13
Improve Harvester Logging and Exception Handling

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5265

## About

- Adds some additional logging
- Adds new critical exception: `SendNotificationException`
    - Should be raised, but not block the connection exit towards the end of the harvest
    - Should be raised on connection errors, and other errors.
    - Grabs job id and logs the error
    - Adds a test that tests a connection error
- Additional Fixes:
    - Noticed that `"notification_emails"` was set in some of the conftests as just strings when they should be lists.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


harvester/exceptions.py
```diff
@@ -47,6 +47,10 @@ class CompareException(HarvestCriticalException):
     pass
 
 
+class SendNotificationException(HarvestCriticalException):
+    pass
```
jbrown-xentity, line 51: We should probably rename the `HarvestCriticalException` base class, if they aren't all critical. Maybe just `HarvestJobException`. And then rename the ones that are "Critical" to have that in the name.

jbrown-xentity (approved):
LGTM! We could use some renaming after this change, but not critical (see what I did there? 😆 )

harvester/exceptions.py
cmhedrickREI: I agree with this! Also thank you for the bit of humor on a monday morning :laughing:

PR #264, opened 2025-06-16
Login fixes

jbrown-xentity:
# Pull Request

Related to recent e2e playwright test failures

## About

Also fixes a weird bug; if your local computer time is slightly ahead of actual time, it can cause a failure to login from login.gov. This sets it back 100 seconds, to make sure that error doesn't occur.
It also updates the certificate in login.gov, as it had expired.

## PR TASKS

- [x] Code well documented
- [x] Files linted (in final commit, yay for checklists!)


cmhedrickREI (approved):
Approved! Also thanks for making those linting changes too! :+1:

PR #266, opened 2025-06-16
Upgrade psycopg

jbrown-xentity:
# Pull Request

We had a number of bugs related to the DB connection:
- Random page failures (somewhere between 1-5 and 1-20)
- Authentication failures unless we used `sync` worker, which isn't actually parallel.

This has been deployed to dev, and is working well from a UI perspective. I want to investigate more the harvester, as there are some stalled jobs. It may be unrelated.

## About

[psycopg3](https://www.psycopg.org/psycopg3/docs/advanced/async.html#async) has out of the box async support, which fits our use case.

Utilizes basic setup for psycopg3 (also known as simply psycopg): https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#module-sqlalchemy.dialects.postgresql.psycopg

## PR TASKS

- [ ] ~Code well documented~
- [ ] ~Files linted~


cmhedrickREI (commented):
lgtm!

neilmb (dismissed):

jbrown-xentity:
This needs more changes. We are getting a recursion error when trying to login in the development environment. Debugging...

neilmb (approved):

PR #267, opened 2025-06-17
Tweaks necessary to make importing work

neilmb:
# Pull Request

To make the import to Catalog-next work, I had to tweak the JSON that we sent to the `organization_create` call.

## About

I think that this gets us the data we need in Catalog-next but doesn't bring over very much else for organizations from Catalog to Catalog-next which I think is reasonable.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM!

PR #269, opened 2025-06-18
waf optimization

rshewitt:
- change to per-record processing across the board (i.e. for datajson and waf)
- add waf datetime filtering
- only pull minimum fields from latest internal records
- fix affected tests

# Pull Request

Related to [5261](https://github.com/GSA/data.gov/issues/5261)

## About
-the major change introduced by this work is how we process records. prior to this change the harvest runner gathered all the external documents and determined all the work needing to be done (i.e. compare for all records). this approach couldn't scale with our largest harvest source IOOS which is was prompted this optimization.

## PR TASKS

- [ ] Code well documented
- [x] Files linted


harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
                 setattr(self, attr, getattr(source_data, attr))
         except Exception as e:
             raise ExtractInternalException(
-                f"failed to extract source info from {job_id}. exiting",
+                f"failed to extract source info from {job_id}. exiting :: {repr(e)}",
                 self.job_id,
             )
 
-    def internal_records_to_id_hash(self, records: List[dict]) -> None:
+    def internal_records_to_class(self, records: List[dict]) -> None:
```
neilmb: **question (non-blocking)**: This method has only side-effects taking the given list of records and storing them on a dict in the instance after some transformation, but the name doesn't make that obvious and there is no docstring explaining it. 

If we're going to rename the method, could we name it something more descriptive? Maybe `store_records_as_internal`? And a docstring explaining what it does?

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+                _date_finished=record["date_finished"],
             )
 
-    def get_record_identifier(self, record: dict) -> str:
-        record_id = "identifier" if self.schema_type.startswith("dcatus") else "url"
-
-        if record_id not in record:
-            raise Exception
+    def write_duplicate_to_db(self, identifier: str) -> None:
+        # Create a minimal harvest_record for error tracking purposes only
```
neilmb, line 211: **suggestion (non-blocking)**: In python, we use docstrings instead of comments to explain the purpose of a method. This name doesn't appear to be appropriate for what the "docstring" says the method does?

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
-        if record_id == "":
-            raise Exception
+        raise DuplicateIdentifierException(
+            f"Duplicate identifier '{identifier}' found for source: {self.name}",
+            self.job_id,
+            harvest_record_id,
+        )
 
-        return record_id
+    def filter_duplicate_identifiers(self) -> None:
```
neilmb, line 229: **suggestion (non-blocking)**: Docstring for this method, please.

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+        # first record all the duplicates in db
+        indices = find_duplicates_by_index(self.external_records)
+        for idx in indices:
+            try:
+                self.write_duplicate_to_db(self.external_records[idx]["identifier"])
```
neilmb, line 237: **question (non-blocking)**: It looks like this method raises an exception unconditionally. Why run it inside a try/except block if we know it's going to raise an exception?

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+            except DuplicateIdentifierException:
+                del self.external_records[idx]
+                self.reporter.update("errored")
+                continue
```
neilmb: **suggestion (non-blocking)**: I think that this `continue` is unnecessary since we are already at the end of the loop body.

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
 
-    def external_records_to_id_hash(self, records: List[dict]) -> None:
-        # ruff: noqa: F841
+    def filter_waf_files_by_datetime(self) -> list:
```
neilmb: **suggestion (non-blocking)**: Docstring please and I don't see anything that makes this method about "waf files". It looks like it just operates on self.external_records regardless of how they got there?

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+                identifier = data["identifier"]
+                internal_record = self.internal_records.get(identifier, None)
+                if internal_record is not None:
+                    # date_finished should never be None since we
+                    # only grab successful internal records which are datetime stamped
+                    if data["modified_date"] > internal_record.date_finished:
+                        records.append(data)  # update
+                else:
+                    records.append(data)  # create
+        except:  # noqa E722
```
neilmb: **question (blocking)**: This bare except seems really unnecessary. The body of the try  doesn't do that much, it just gets some properties from `data` and `internal_record` and appends to some lists. 

Can we guard the manipulations inside the try so they always succeed? Or limit the except here to some exceptions that we know could occur?

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+            # if anything happens just use the original data
+            records = self.external_records
+        finally:
+            self.external_records = records
+
+    def determine_internal_deletions(self) -> None:
```
neilmb, line 265: **suggestion (non-blocking)**: Docstring please.

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+        external_ids = set([record["identifier"] for record in self.external_records])
+        internal_ids = set(self.internal_records.keys())
+        self.deletions = internal_ids - external_ids
+
+    def setup_internal_records_to_be_deleted(self) -> any:
```
neilmb: **question (non-blocking)**: This seems like a strange name for a method that is a generator. Maybe just `internal_records_to_be_deleted`? Or `iter_internal_records_to_be_deleted?` It also seems surprising that this would write to the database. Can we make a docstring and be sure to call out that side-effect?

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+        for identifier in self.deletions:
+            internal_record = self.internal_records[identifier]
+            internal_record.action = "delete"
+            internal_record.write_compare_to_db()
+            yield internal_record
+
+    def setup_records_to_be_processed(self) -> any:
```
neilmb: **question (non-blocking)**: As above, this name seems very strange for a method that is a generator. Maybe `external_records_to_process`? And it handles the actual downloading which feels surprising. Can you add a dosctring that explains that this is where that happens?

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
-                    f"{self.name} {self.url} failed to convert to id:hash :: {repr(e)}",
+                self.reporter.update("errored")
+                raise ExternalRecordToClass(
+                    f"{self.name} {self.url} failed to prepare record for harvest :: {repr(e)}",
                     self.job_id,
                 )
 
-    def prepare_external_data(self) -> None:
-        logger.info("retrieving and preparing external records.")
+    def acquire_minimum_external_data(self) -> list:
```
neilmb, line 320: **suggestion (non-blocking)**: Docstring. This method does lots of interesting things for different types of sources. It would be good to know what they are.

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
+                self.external_records = traverse_waf(self.url)
         except Exception as e:
             # ruff: noqa: E501
             raise ExtractExternalException(
                 f"{self.name} {self.url} failed to extract harvest source. exiting :: {repr(e)}",
                 self.job_id,
             )
 
-    def prepare_internal_data(self) -> None:
+    def acquire_minimum_internal_data(self) -> None:
```
neilmb, line 350: **suggestion (non-blocking)**: Docstring. This method does some important things and it would be nice to know what those are.

harvester/harvest.py
```diff
@@ -188,236 +178,161 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
...
-            if record.status == "error" or record.status is None:
-                continue
-
-            if record.action == "delete":
-                record.delete_self_in_db()
-            elif record.action is not None:
-                record.update_self_in_db()
+    def update_harvest_job(self, status: str):
+        self.db_interface.update_harvest_job(
+            self.job_id, {"status": status, "date_finished": get_datetime()}
```
neilmb, line 400: **question (non-blocking)**: Setting the `date_finished` property seems surprising to me. How about a docstring that explains why it does that and maybe a name that makes that clear like `finish_job_with_status`?

harvester/harvest.py
```diff
@@ -488,38 +403,21 @@ def send_notification_emails(self, job_results: dict) -> None:
...
-        self.records.extend(records)
         logger.warning(f"{len(db_records)} uncleared incoming db records")
-        for record in db_records:
+        for db_record in db_records:
+            record = self.make_record_contract(db_record)
             self.db_interface.delete_harvest_record(
                 identifier=record.identifier, harvest_source_id=self.id
             )
+            self.reporter.update("delete")
+            del record
```
neilmb: **comment (non-blocking)**: This is quite un-pythonic, we would normally just let the instance created by `make_record_contract` get garbage-collected when it goes out of scope. If you need to keep this, can we put a comment before this line explaining how we know that it is necessary.

harvester/harvest.py
```diff
@@ -692,9 +575,71 @@ def status(self, value) -> None:
...
+                self.action = "create"
+
+            if self.action is not None:
+                self.write_compare_to_db()
+            else:
+                self.harvest_source.reporter.update(None)
+
+            self.harvest_source.reporter.total += 1
+
+        except Exception as e:
```
neilmb: **question (non-blocking)**: This seems like another really strange except block. The try isn't doing too much. Why would we expect anything there to fail and if it did fail, wouldn't we want to see the error because something was wrong, instead of raising an exception that we are just going to ignore in the `harvest()` method? If this is really what we want, then we need a docstring that explains that we are doing it and why.

harvester/harvest.py
```diff
@@ -692,9 +575,71 @@ def status(self, value) -> None:
             raise ValueError("status must be a string")
         self._status = value
 
+    def harvest(self) -> None:
```
neilmb, line 643: **suggestion (non-blocking)**: Docstring. This does lots and lots of things and seems to be the main method of the `Record` class and yet there's nothing here to say so.

harvester/harvest.py
```diff
@@ -796,38 +758,25 @@ def update_self_in_db(self) -> bool:
...
-    if job_type in ["harvest", "clear", "sync", "force_harvest"]:
-        harvest_source.sync()
+    if job_type == "validate":
+        harvest_source.acquire_minimum_external_data()
+        for record in harvest_source.setup_records_to_be_processed():
+            if harvest_source.schema_type.startswith("iso19115"):
+                record.transform()
+            try:
+                record.validate()
+            except:  # noqa: E722
```
neilmb, line 845: **comment (blocking)**: This bare except seems completely wrong. Why would we not want to see any exceptions ever from the `validate` method?

harvester/utils/general_utils.py
```diff
@@ -94,8 +95,68 @@ def download_file(url: str, file_type: str) -> Union[str, dict]:
...
+        "harvest_source_id": record.harvest_source.id,
+        "source_hash": record.metadata_hash,
+        "source_raw": record.source_raw,
+        "action": record.action,
+        "ckan_id": record.ckan_id,
+        "ckan_name": record.ckan_name,
+    }
+
+
+def find_duplicates_by_index(records: list):
```
neilmb: **suggestion (non-blocking)**: Docstring. This doesn't "find" the duplicates, it filters them out. Maybe call it `remove_duplicates`? And what does `by_index` mean? Isn't it filtered by `identifier`?

harvester/utils/general_utils.py
```diff
@@ -94,8 +95,68 @@ def download_file(url: str, file_type: str) -> Union[str, dict]:
...
+    # retains the first instance of a duplicate.
+    seen = set()
+    output = []
+    for i in range(len(records)):
+        identifier = records[i]["identifier"]
+        if identifier in seen:
```
neilmb, line 133: **comment (blocking)**: This seems wrong. Shouldn't we append the record to `output` if it is NOT in `seen`? This function should have a unit test that caught this.

harvester/utils/general_utils.py
```diff
@@ -94,8 +95,68 @@ def download_file(url: str, file_type: str) -> Union[str, dict]:
...
+            output.append(i)
+        seen.add(identifier)
+    output.sort(reverse=True)  # avoid index shifting
```
neilmb, line 136: **comment (non-blocking)**: This seems very surprising. The elements of `output` are dictionaries and sorting by them seems totally unhelpful. Why not keep the original order that came from `records`? If this is important, then it should be described in the docstring and there should be a unit test for it.

harvester/utils/general_utils.py
```diff
@@ -94,8 +95,68 @@ def download_file(url: str, file_type: str) -> Union[str, dict]:
...
+
+    return output
+
+
+def get_waf_datetimes(soup: BeautifulSoup, files: list) -> list:
```
neilmb: **suggestion (non-blocking)**: Docstring. It appears that `files` is never used except to check the number of found timestamps. Could the second argument be changed to `expected_length` and then we wouldn't have to compute `len` on the same list three separate times?

harvester/utils/general_utils.py
```diff
@@ -105,44 +166,46 @@ def traverse_waf(
...
-                and anchor["href"] not in filters  # and it's not on our exclusion list?
-            ):
-                folders.append(urljoin(url, anchor["href"]))
-
-            if anchor["href"].endswith(file_ext):
-                # TODO: just download the file here
-                # instead of storing them and returning at the end.
-                files.append(urljoin(url, anchor["href"]))
+    if not res.ok:
+        raise Exception
```
neilmb: **suggestion (non-blocking)**: The plain `Exception` here hides all information about why the request failed. Suggest replacing this with `res.raise_for_status()`

tests/integration/harvest/test_compare.py
```diff
@@ -32,69 +31,31 @@ def test_compare(
             interface.add_harvest_record(data)
 
         harvest_source = HarvestSource(internal_compare_data["job_id"])
-        harvest_source.extract()
-        harvest_source.compare()
 
-        harvest_job_records = interface.get_harvest_records_by_job(
-            job_data_dcatus["id"], paginate=False
+        harvest_source.acquire_minimum_internal_data()
```
neilmb: **question (non-blocking)**: This looks suspiciously like the body of `Record.harvest()`. Should we just be calling that method instead of repeating all of its code here?

tests/integration/harvest/test_exception_handling.py
```diff
@@ -192,22 +203,36 @@ def test_dcatus_to_ckan_exception(
...
 
-        test_record = [x for x in harvest_source.records if x.identifier == "cftc-dc1"][
-            0
-        ]
+        external_records_to_process = harvest_source.setup_records_to_be_processed()
 
-        interface_record = interface.get_harvest_record(test_record.id)
-        interface_errors = interface.get_harvest_record_errors_by_record(test_record.id)
+        # 7 records
+        records = [record for record in external_records_to_process]
```
neilmb: **suggestion (non-blocking)**: This is the exact same thing as `list(external_records_to_process)` which could just go into the previous line.

tests/integration/harvest/test_exception_handling.py
```diff
@@ -223,20 +248,30 @@ def test_ckan_sync_exception(
...
```
neilmb: **suggestion (non-blocking)**: As above, use `list(iterator)` to consume it easily.

tests/integration/harvest/test_exception_handling.py
```diff
@@ -266,11 +301,23 @@ def test_validate_nested_exception_handling(
...
+        external_records_to_process = harvest_source.setup_records_to_be_processed()
+
+        records = [record for record in external_records_to_process]
+
+        for test_record in records:
+            test_record.compare()
+            test_record.validate()
+            try:
+                test_record.sync()
+            except:  # noqa: E722
```
neilmb: **question (non-blocking)**: Why would we be swallowing exceptions in a test where we control the test data?

tests/integration/harvest/test_exception_handling.py
```diff
@@ -336,16 +390,23 @@ def test_ckan_sync_500_error(
         interface.add_harvest_source(source_data_dcatus)
         harvest_job = interface.add_harvest_job(job_data_dcatus)
         job_id = harvest_job.id
+
         harvest_source = HarvestSource(harvest_job.id)
-        harvest_source.extract()
-        harvest_source.compare()
+        harvest_source.acquire_minimum_internal_data()
```
neilmb: **comment (question))**: I have seen this code pattern enough now that I think it should be refactored out into either a method on `harvest_source`, or at least a helper function in this test file.

tests/integration/harvest/test_validate.py
```diff
@@ -17,9 +17,11 @@ def test_validate_dcatus_federal(
         harvest_job = interface.add_harvest_job(job_data_dcatus)
 
         harvest_source = HarvestSource(harvest_job.id)
-        harvest_source.prepare_external_data()
+        harvest_source.acquire_minimum_external_data()
+        external_records_to_process = harvest_source.setup_records_to_be_processed()
 
-        test_record = harvest_source.external_records["cftc-dc1"]
+        # "cftc-dc1" is always the first one
+        test_record = [record for record in external_records_to_process][0]
```
neilmb: **suggestion (non-blocking)**: No need to consume an entire iterator to get the first item in it. Use `next(external_records_to_process)` to just get the first one.

tests/integration/harvest/test_validate.py
```diff
@@ -36,9 +38,11 @@ def test_validate_dcatus_non_federal(
         harvest_job = interface.add_harvest_job(job_data_dcatus_non_federal)
 
         harvest_source = HarvestSource(harvest_job.id)
-        harvest_source.prepare_external_data()
+        harvest_source.acquire_minimum_external_data()
+        external_records_to_process = harvest_source.setup_records_to_be_processed()
 
-        test_record = harvest_source.external_records["cftc-dc1"]
+        # "cftc-dc1" is always the first one
+        test_record = [record for record in external_records_to_process][0]
```
neilmb: **suggestion (non-blocking)**: As above, use `next()`.

tests/integration/harvest/test_validate.py
```diff
@@ -55,12 +59,11 @@ def test_invalid_license_uri_dcatus_non_federal(
...
-        harvest_source.prepare_external_data()
+        harvest_source.acquire_minimum_external_data()
+        external_records_to_process = harvest_source.setup_records_to_be_processed()
 
-        test_id = (
-            "https://www.arcgis.com/home/item.html?id=99731bb0369848169d98f31ce83fb0e2"
-        )
-        test_record = harvest_source.external_records[test_id]
+        # "cftc-dc1" is always the first one
+        test_record = [record for record in external_records_to_process][0]
```
neilmb: As above.

neilmb (changes requested):
I have many comments, but the main ones are about docstrings and exception handling. Generally, I believe that exceptions are meant to be seen, not hidden away unless we are very, very sure that we need to hide them. If we do need to, then we should be very, very explicit about when we do it and why.

cmhedrickREI (dismissed):
LGTM excellent work and nice changes

harvester/harvest.py
rshewitt: i isolated this logic to its own function and kept @Jin-Sun-tts comments. i'll add a docstring!

harvester/harvest.py
rshewitt: our custom exceptions are responsible for logging errors and writing errors to the db. the purpose is to reduce repetitive code on exception. keeping to this pattern we raise `DuplicateIdentifierException`. the parent function anticipates it and moves forward when discovered. i could hoist that exception to `run_full_harvest` since we're doing exception handling there. would you prefer that?

tests/integration/harvest/test_exception_handling.py
rshewitt: in this case we should be good to not wrap `sync` in a try/except since the anticipated exceptions are handled within that function and not raised. this isn't the case for `validate`, just mentioning it since it's right above sync, since it raises an exception on failed validation. I definitely think we could better standardize our exceptions and i'm totally open for any ideas on how we can improve them across the board!

harvester/harvest.py
rshewitt: i'll add a docstring! the function is only called when the source type is `waf`. i can clarify that in the docstring.

harvester/harvest.py
rshewitt: totally. i wrapped this in a try/except as a safeguard to generally handle anything unexpected but i agree that the body doesn't do much and it's probably safe to not wrap it with try/except.

harvester/harvest.py
rshewitt: i like `iter_internal_records_to_be_deleted`. i'll add a docstring! this db write pattern is intended to occur immediately after identifying what work needs to be done. in this case the deletion of the record. same thing happens in `compare`.

harvester/harvest.py
rshewitt: i'll remove the `del`. i was in memory-saving mode and threw that in there just to be make sure but you're right.

harvester/harvest.py
rshewitt: this is another case of covering our bases where if something unexpected goes wrong then it's handled without knowing an exact circumstance. i'll remove it and add docstring.

harvester/harvest.py
rshewitt: i quickly wrote this because we're intending on removing "validate' jobs anyways. should I still update this?

harvester/utils/general_utils.py
rshewitt: the function returns the element position/index of duplicates within the list. it doesn't remove anything from the input list. the output of the function is eventually used to delete the duplicate elements while retaining the first instance of duplicated records. i could name it `find_duplicates_by_elem_position`? 

```python
data = [ { "identifier": "a" }, { "identifier": "a" }, { "identifier": "b" }, { "identifier": "b" }]
find_duplicates_by_index(data) # [ 3, 1 ]
```

harvester/utils/general_utils.py
rshewitt: i'll add a unit test! if we've already "seen" the record identifier then it's a duplicate.

harvester/utils/general_utils.py
rshewitt: `output` is a list of integers representing element positions of duplicates records. this list is then used to record duplicate record errors in the db and to remove from `self.external_records`. sorting it in reverse (.sort edits in place) places the largest numbers first. this is necessary to avoid index shifting when you're deleting from a list. without this take the following scenario...
```python
positions = [ 1, 3 ]
data = [ 0, 0, 1, 1 ] 

# deleting data[1] can shift the data since we're mutating the list. what happens when we try to del data[3] on [ 0, 1, 1]?
```

neilmb: > `output` is a list of integers representing element positions of duplicates records. this list is then used to record duplicate record errors in the db and to remove from `self.external_records`. sorting it in reverse (.sort edits in place) places the largest numbers first. this is necessary to avoid index shifting when you're deleting from a list. without this take the following scenario...
> 
> ```python
> positions = [ 1, 3 ]
> data = [ 0, 0, 1, 1 ] 
> 
> # deleting data[1] can shift the data since we're mutating the list. what happens when we try to del data[3] on [ 0, 1, 1]?
> ```

You have great docstrings in your PR comments. This explanation will be perfect in the code.

Maybe `find_indexes_for_duplicates`?

harvester/harvest.py
neilmb: No, I just didn't know that we had chosen to use exceptions to do this.

harvester/harvest.py
neilmb: This is fine for now.

neilmb (dismissed):

neilmb:
@rshewitt This is ready to go now I think. When you merge it, let's get it running on `development` too.

neilmb (approved):

PR #278, opened 2025-06-27
Fix and test showing only recent failed jobs

neilmb:
# Pull Request

The intention was to only have the failed jobs from the last 24 hours shown on the metrics page, but if you don't test for it, it doesn't exist, and we were putting all the jobs on there instead. This PR adds the appropriate filter and a test case for the behavior.

## About

It's surprisingly easy to write fixtured app tests in Pytest without involving Playwright.

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -1111,8 +1111,9 @@ def view_metrics():
             per_page=pagination.per_page,
             order_by="desc",
         )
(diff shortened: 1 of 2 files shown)


jbrown-xentity (dismissed):
LGTM

neilmb:
@jbrown-xentity re-approve after fixing a merge conflict?

jbrown-xentity (approved):
LGTM

PR #281, opened 2025-06-30
Fix metrics pagination by importing inside block

neilmb:
# Pull Request

I made a bug on the metrics page with pagination by using a Jinja macro inside of the block that our HTMX pagination code renders. Rendering just the block fails because the macro hasn't been imported there. This repeats the import inside the block so that the HTMX call can succeed.

## About

The template fragment rendering extension we use for HTMX means that we have to keep track of these imports manually. If we use more refactoring of Jinja macros, we might see more bugs like this.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM

PR #282, opened 2025-06-30
reduce bulky exception messages

rshewitt:
- truncate validation message & add test
- increase task memory
- remove full object represention from ckan responses
- fix tests

# Pull Request

Related to [5287](https://github.com/GSA/data.gov/issues/5287)

## About
- these changes are intended to reduce the size of bulk exceptions which can contribute to increased memory usage of datagov-harvest tasks. 

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/harvester/lib/cf_handler.py
+++ b/harvester/lib/cf_handler.py
@@ -42,7 +42,7 @@ class CFHandler:
     def start_task(self, command, task_id):
         """Start a task on the currently running app's GUUID."""
         self.setup()
-        TASK_MEMORY = os.getenv("HARVEST_RUNNER_TASK_MEM", "1536")
+        TASK_MEMORY = os.getenv("HARVEST_RUNNER_TASK_MEM", "2056")
         TASK_DISK = os.getenv("HARVEST_RUNNER_TASK_DISK", "4096")
         return self.task_mgr.create(
             self._app_guuid(),
--- a/harvester/utils/general_utils.py
+++ b/harvester/utils/general_utils.py
@@ -21,6 +21,28 @@ logger = logging.getLogger()
 FREQUENCY_ENUM = {"daily": 1, "weekly": 7, "biweekly": 14, "monthly": 30}
 
 
+def truncate_validation_message(e: Exception) -> str:
+    """
+    truncates the dcatus validation message to 150 characters using
+    " ...[truncated]... " as the substituted value within the
+    message.
+    """
+    msg = e.message
+    # limit the message size to be 150 characters
+    if len(msg) > 150:
+        msg = msg[:100] + " ...[truncated]... " + msg[-50:]
+
+    elem = ""
+    try:
+        if e.schema_path[0] == "properties":
+            elem = e.schema_path[1]
+            elem = "'" + elem + "':"
+    except Exception:
+        pass
+
+    return elem + msg
+
+
 def prepare_transform_msg(transform_data):
     # ruff: noqa: E731
     mask_info = lambda s: "WARNING" in s or "ERROR" in s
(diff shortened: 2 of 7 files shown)


harvester/lib/cf_handler.py
```diff
@@ -42,7 +42,7 @@ def setup(self):
     def start_task(self, command, task_id):
         """Start a task on the currently running app's GUUID."""
         self.setup()
-        TASK_MEMORY = os.getenv("HARVEST_RUNNER_TASK_MEM", "1536")
+        TASK_MEMORY = os.getenv("HARVEST_RUNNER_TASK_MEM", "2056")
```
neilmb, line 45: **question (non-blocking)**: This PR was supposed to be reducing memory usage, so why do we need to increase the default memory allocation of every task here?

harvester/utils/general_utils.py
```diff
@@ -21,6 +21,28 @@
...
+    # limit the message size to be 150 characters
+    if len(msg) > 150:
+        msg = msg[:100] + " ...[truncated]... " + msg[-50:]
+
+    elem = ""
+    try:
+        if e.schema_path[0] == "properties":
+            elem = e.schema_path[1]
+            elem = "'" + elem + "':"
+    except Exception:
```
neilmb, line 40: **suggestion (blocking)**: No bare exceptions, please. I don't know where you got this habit from, but this isn't good python (e.g. https://peps.python.org/pep-0760/). If an attribute might not exist, then use `except AttributeError`. If there might not be two components in the schema path, use `except IndexError`. If both, use `except (IndexError, AttributeError)`

tests/integration/app/test_load_manager.py
```diff
@@ -316,7 +316,7 @@ def test_assert_env_var_changes_task_size(
         load_manager = LoadManager()
         load_manager.trigger_manual_job(source_data_dcatus["id"])
         start_task_mock = CFCMock.return_value.v3.tasks.create
-        assert start_task_mock.call_args.kwargs["memory_in_mb"] == "1536"
+        assert start_task_mock.call_args.kwargs["memory_in_mb"] == "2056"
```
neilmb, line 319: **comment (non-blocking)**: If we revert the default memory change above, this will need to change too.

neilmb (changes requested):
This is a nice improvement and there's just the one glaring problem with an unnecessary bare except.

harvester/lib/cf_handler.py
rshewitt: Prior to this specific update, i ran IOOS and the stable memory usage jumped from 1 GB, see [prior run](https://github.com/GSA/data.gov/issues/5261#issuecomment-3001317145), to 1.5 GB (see picture). I didn't spend time figuring out why there was a 0.5 GB increase between runs (maybe there's variability between runs. i [observed](https://github.com/GSA/data.gov/issues/5287#issuecomment-3020214517) 1.5 GB where IOOS was the only running task but simultaneous tasks don't appear to interfere with each other otherwise why is the app memory usage and out-of-memory event almost identical to the task memory allocation?) but because of this 0.5 GB increase i was concerned the job would terminate before seeing results from shortening validation messages. interestingly, the delta in memory usage between stable and peak (leading up to out of memory) between runs where 1 GB is the stable usage and 1.5. GB is the stable usage is 0.6 GB. I didn't need to submit this as a draft because i don't intend to increase the task memory, which is confusing from a review perspective, but i wanted visibility of what i was doing instead of just deploying it from my local. 
Screenshot 2025-07-01 at 9 12 02 AM

harvester/utils/general_utils.py
rshewitt: i copy/pasted this from [prod](https://github.com/GSA/ckanext-datajson/blob/105aaf01f70d053fabeefb61de9dff49f3e707bf/ckanext/datajson/datajson.py#L395-L409) in an effort to get something out quick but i agree this should be updated to handle exact scenarios rather than broad catch-alls. i'll update it when i submit the PR for review.

PR #283, opened 2025-07-01
Use the ORM to replace some IDs with names

neilmb:
# Pull Request

Using backrefs marked as joined, we can replace some of the opaque IDs in the web app with the corresponding names:

- On the Harvest Source page, we use the organization name in the link to the organization (as well as formatting the notification emails list).
- On the Harvest Job page, we use the harvest source name in the link to the harvest source
- On the metrics page, we link to harvest sources by name instead of ID.

## About

To accomplish this efficiently, we have changed a couple of the `backref`s in our ORM models to be `lazy="joined"` which means that a query for the child object (source, or job, or error) automatically joins the parent table (organization, source, or job respectively) to be able to get the parent properties without a separate database query. In general, this is much more efficient than going back to the database N+1 times for each parent property we need.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -727,7 +727,7 @@ def view_harvest_source(source_id: str):
         }
         source = db.get_harvest_source(source_id)
         data = {
-            "source": db._to_dict(source),
+            "source": source,
             "summary_data": summary_data,
             "jobs": jobs,
             "chart_data": chart_data,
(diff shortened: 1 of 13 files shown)


jbrown-xentity (approved):
Definitely makes this more user friendly, nice!

PR #286, opened 2025-07-07
QA script fixes

neilmb:
# Pull Request

The issue GSA/data.gov#5251 asks for a couple improvements to our QA script to put output in other directories and do better to match datasets between catalog-next and catalog. This PR refactors the QA script into individual files with isolated output directories. In `datasets.py` it does more "best-effort" attempts to find a matching dataset in catalog using `identifier` and sometimes `guid`.

## About

This probably was unnecessarily invasive to the original script, but it switched to `click` for command line handling and added CLI options for sample size and random seed so that we can repeat QA runs of various sizes if we need to.

I acknowledge that I didn't add any unit tests since they didn't exist, just running the script manually to test its behavior.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


scripts/lib/qa/datasets.py
```diff
@@ -0,0 +1,178 @@
...
+        except StopIteration:
+            return None
+
+    def get_num_datasets(self):
+        """Return the number of datasets in this catalog."""
+        res = requests.get(f"{self.base_url}/api/action/package_search")
+        res.raise_for_status()
+        return res.json()["result"]["count"]
+
+    def fetch_matching_dataset(self, other):
```
rshewitt, line 67: +1 for adding more options for finding the matching dataset. hopefully this results in fewer non-matching datasets

scripts/lib/qa/datasets.py
```diff
@@ -0,0 +1,178 @@
...
+            res = requests.get(
+                f"{self.base_url}/api/action/package_search"
+                f"?start={start}&rows={self.sample_size}&sort=id%20asc"
+            )
+            res.raise_for_status()
+            self.datasets = res.json()["result"]["results"]
+        else:
+            # datasets from other catalog were specified
+            click.echo("Getting matching datasets from catalog")
+            with click.progressbar(self.other_datasets) as bar:
```
rshewitt, line 127: definitely more satisfying than a log message (or nothing at all!)

scripts/lib/qa/harvest_sources.py
```diff
@@ -0,0 +1,174 @@
...
+    for hs_name, hs_data in catalog_harvest_sources.sources.items():
+        # we don't harvest these things anymore
+        if hs_data["source_type"] in ["csw", "arcgis", "geoportal"]:
+            continue
+
+        harvest_source_data_next = catalog_next_harvest_sources.sources.get(
+            hs_name, catalog_harvest_sources.source_template
+        )
+
+        compare_data = [
```
rshewitt, line 125: are we still intending on comparing the organization of the harvest sources? if so, there's 6 fields in the rows and 7 in the header

scripts/lib/qa/utils.py
```diff
@@ -0,0 +1,40 @@
...
+CATALOG_PROD_BASE_URL = "https://catalog.data.gov"
+CATALOG_NEXT_BASE_URL = "https://catalog-next-dev-datagov.app.cloud.gov"
+
+
+class OutputBase:
+    """Output methods and directory data."""
+
+    def __init__(self, output_dir=None, clear_on_start=True):
+        if output_dir is None:
+            self.output_dir = Path()
```
rshewitt, line 15: looks like you're passing in `output_dir` in the comparison functions but did you mean to say `output_dir = Path()` here? otherwise you're assigning self.output_dir = None in the default case. I know this isn't part of the proper harvester and we don't have tests for a script like this but wanted to mention this.

scripts/lib/qa/harvest_sources.py
neilmb: Good call, I had been in the process of changing this and lost track of it.

scripts/lib/qa/utils.py
neilmb: This uses `None` as a semaphore, a sign that no output_dir was specified, in which case we need to do something different (not empty the current directory).

scripts/lib/qa/utils.py
```diff
@@ -0,0 +1,40 @@
...
+            # no clean up if no output_dir is specified
+            return
+
+        self.output_dir = Path(output_dir)
```
rshewitt, line 19: this was the line (19) I was referring to in the above comment. i swear i clicked this line but idk. what i'm saying is that `Path(None)` is going to throw an exception.

neilmb: If output_dir is `None` then the previous `if` statement has already returned and we won't get here.

rshewitt: omg yes i see that

rshewitt (approved):

PR #287, opened 2025-07-08
waf memory optimization

rshewitt:
- minimize needed fields from internal records
- loop through external records via while/pop
- bug fix
- fix db test

# Pull Request

Related to [#5287](https://github.com/GSA/data.gov/issues/5287)
Related to [#5261](https://github.com/GSA/data.gov/issues/5261)

## About
- IOOS has routinely caused tasks to run out of memory which we want to avoid. this PR includes fixes which, hopefully, allow large WAF harvest sources to no longer run out of memory when running. 
- [latest IOOS job](https://datagov-harvest-dev.app.cloud.gov/harvest_job/d026ab05-b35a-44f7-b19e-55b6411cfc56) ran to completion without any gradual memory increase. the plateaus you see are caused by other harvest sources being processed at the same time.
```terminal
716   harvest-job-d026ab05-b35a-44f7-b19e-55b6411cfc56-harvest   SUCCEEDED   Mon, 07 Jul 2025 16:44:47 UTC   python harvester/harvest.py d026ab05-b35a-44f7-b19e-55b6411cfc56 harvest
```

Screenshot 2025-07-08 at 8 14 17 AM

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


database/interface.py
```diff
@@ -561,25 +561,30 @@ def get_latest_harvest_records_by_source_orm(
...
-            self.db.query(
-                HarvestRecord.identifier,
-                HarvestRecord.source_hash,
-                HarvestRecord.ckan_id,
-                HarvestRecord.ckan_name,
-                HarvestRecord.date_created,
-                HarvestRecord.id,
-                HarvestRecord.action,
-            )
+            self.db.query(HarvestRecord)
```
jbrown-xentity, line 564: Won't this include the full content, which will be a memory load? Why stop being specific about the fields we want?

jbrown-xentity (approved):
One quick question...

database/interface.py
neilmb: I think that that the self-join in the subquery brings in all the fields anyways and what is most important is that we don't return them out of this function into the longer-lived `Record` objects. In here, I suspect they will get garbage-collected.

neilmb (approved):

database/interface.py
rshewitt: I thought limiting fields in the subquery here would show up in the output but it doesn't. I should have known it wasn't working because at the time of that commit i didn't need to update `test_get_latest_harvest_records` which meant all fields were being read. Specifying fields in the return statement works which is reflected in the test.

PR #288, opened 2025-07-08
Send email notifications when finding failed jobs

neilmb:
# Pull Request

In GSA/data.gov#5301 we need the O+M team to know when we find new failed jobs. This PR sends an email to the same "recipient" from our SMTP_CONFIG environment setup that already gets emails for successful and failed jobs.

## About

I refactored some functionality from `harvester/harvest.py` into `harvester/utils/general_utils.py` so it could be re-used and prevent circular imports.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
This looks good from my end!

PR #289, opened 2025-07-08
Details on commands for scheduling harvest jobs

neilmb:
# Pull Request

When I was scheduling all of the harvest jobs for our sources in GSA/data.gov#5284 I was using shell commands that aren't quite ready to be a Bash script, but I wanted to document what I was doing.

## About

Maybe this could eventually be completely automated, but for now, I think it's more useful as commands that can be pasted into a terminal and each command can have its effects inspected.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,28 @@
+# Migration scripts
+
+Migrating from our production CKAN into our new system takes lots of automated
+data transformation. The scripts in this directory help with that.
+
+## Scheduling Harvest Sources for Harvesting
+
+Part of our QA process is to schedule a harvest job for every one of our
+harvest sources. Here's some commands to pull the list of harvest sources. (We
+should make a JSON form of this list page.)
+
+```bash
+$ curl https://datagov-harvest-dev.app.cloud.gov/harvest_source_list/ > harvest_source_list.html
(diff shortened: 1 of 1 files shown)


rshewitt (approved):

rshewitt:
nice documentation!

scripts/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Part of our QA process is to schedule a harvest job for every one of our
+harvest sources. Here's some commands to pull the list of harvest sources. (We
+should make a JSON form of this list page.)
+
+```bash
+$ curl https://datagov-harvest-dev.app.cloud.gov/harvest_source_list/ > harvest_source_list.html
+$ grep href=\"/harvest_source/ harvest_source_list.html | sed -e s/\<a// -e s/href=\"// -e s/\"// > harvest_source_paths.txt
+$ sed -e "s:^.*/harvest_source/::" < harvest_source_paths.txt > harvest_source_ids.txt
+$ cf env datagov-harvest | grep FLASK_APP_SECRET_KEY
```
rshewitt, line 16: what's the flask app secret key for?

neilmb: That environment variable is used to store the required value of `API_TOKEN`: https://github.com/GSA/datagov-harvester/blob/main/app/routes.py#L85

PR #291, opened 2025-07-09
Check for more work at the end of a harvest task

neilmb:
# Pull Request

To make harvesting more efficient, at the end of a harvest task, we ask the load manager to schedule at most one new task if there are more jobs to do. We add a new `check_from_task` argument to the `LoadManager._start_new_jobs` function that changes the math to account for the task that is running but shortly won't be and also limits to starting only one replacement task. The limit of one new job isn't strictly necessary, but it limits the chances of a race condition where two tasks that are finishing try to start the same harvest job.

## About

There are similar tests now in `tests/integration/app/test_load_manager` and `tests/integration/harvest_job_flows/tests_harvest_job_full_flow` because the `check_for_more_work` wrapper is currently a very thin wrapper on `_start_new_jobs` but it won't necessarily always be that way, so I wanted to test that function separately.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/lib/load_manager.py
+++ b/harvester/lib/load_manager.py
@@ -31,8 +31,6 @@ class LoadManager:
                 f"err {e} :: CFHandler is not configured correctly. \
                 Check your env vars."
             )
-        self.jobs = []
-        self.running_tasks = []
 
     def _handle_failed_job(self, job):
         """Handle a HarvestJob that failed.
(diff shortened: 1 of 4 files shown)


harvester/lib/load_manager.py
```diff
@@ -83,21 +81,33 @@ def _clean_old_jobs(self):
...
+                f"{running_tasks} running tasks >= max tasks count ({MAX_TASKS_COUNT})."  # noqa E501
             )
             return
         else:
-            slots = MAX_TASKS_COUNT - self.running_tasks
+            slots = MAX_TASKS_COUNT - running_tasks
+
+        if check_from_task:
+            # from a task only do 1 at most
+            slots = 1 if slots > 0 else 0
```
rshewitt, line 105: why not fill all the slots instead of just 1?

jbrown-xentity (approved):
This looks good to me. I also agree that keeping these from running into each other is a good idea.
I'm still nervous about this in practice, I think we could double-run a job and cause some issues, but I don't have any way around that in the given system. Should be a longer discussion/architecture/planning meeting.

harvester/lib/load_manager.py
neilmb: There's an explanation in the PR description, it is (I think) safer against multiple jobs ending at the same time trying to schedule the same job to run twice.

PR #293, opened 2025-07-11
Fix bug with nonexistent db interface method

neilmb:
# Pull Request

This error was hidden by our voracious exception swallowing in the harvest process. There is no `db_interface.finish_with_status` method. This switches the local method to using the interface's existing `db_interface.update_harvest_job` method instead.

## About

There are more errors like this that we will need to find from harvest jobs that complete "successfully" but bury these exceptions on what exactly went wrong. I'm hard-pressed to figure out how to test this.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -398,7 +398,7 @@ class HarvestSource:
         date_finished. date_finished is set here because this function is
         either called on critical exception or the job has completed harvesting.
         """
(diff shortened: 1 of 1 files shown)


jbrown-xentity (approved):
Yeah, that should be the direct command. Not sure how that snuck in there.

PR #299, opened 2025-07-11
Create logic for stopping task if duplicate

jbrown-xentity:
Makes sure these 2 use cases are handled:
- Case where there is already another job in progress (stops and marks as error)
- Case where another task is already running this job (stops and logs an error)

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5309

## PR TASKS

- [x] Code well documented
- [x] Tests written, run and passed (not yet, not sure why)
- [x] Files linted


neilmb (approved):
So good, the tests really help the confidence on this kind of deep change.

PR #307, opened 2025-07-15
Use non-federal DCAT schema for all harvest sources

neilmb:
# Pull Request

Related to GSA/data.gov#5306 which asks to use the less restrictive non-federal DCAT schema for validation of all DCAT harvest sources.

## About

This will only be relevant the next time we do an import into a harvester environment, likely staging.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
Good comments as to why, and leaving the old logic in place. Helpful in the future.

PR #308, opened 2025-07-15
Use placeholders for missing DCAT-US 1.1 required fields

neilmb:
# Pull Request

Related to GSA/data.gov#5323 we want to add some of the DCAT-US required elements with generic placeholder values for WAF/ISO sources that don't include them (because they weren't included in the ISO standards).

## About

I think that the placeholders are all in one place and they should be extensible enough if we change our minds in the future about what the values should be or we want to add placeholders to more fields.

We also added a potential call back into the database to get the harvest source's organization data to use in the `publisher` field.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -15,6 +15,8 @@ from requests.exceptions import HTTPError, Timeout
 sys.path.insert(1, "/".join(os.path.realpath(__file__).split("/")[0:-2]))
 
 
+from database.models import HarvestSource as HarvestSourceORM
+
 # ruff: noqa: E402
 from harvester import SMTP_CONFIG, HarvesterDBInterface, db_interface
 from harvester.exceptions import (
@@ -48,7 +50,6 @@ from harvester.utils.general_utils import (
     sort_dataset,
     traverse_waf,
 )
-from database.models import HarvestSource as HarvestSourceORM
 
 # requests data
 session = create_retry_session()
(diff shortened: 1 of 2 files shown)


harvester/harvest.py
```diff
@@ -647,6 +652,7 @@ def harvest(self) -> None:
                 return
             if self.harvest_source.schema_type.startswith("iso19115"):
                 self.transform()
+                self.fill_placeholders()
```
rshewitt, line 656: we decided we only want to fill placeholders for ISO? the reasoning being we're trying to "throw a bone" to the ISO sources but we're being more strict for dcatus providers because there's no transformation?

tests/integration/harvest/test_validate.py
```diff
@@ -4,6 +4,22 @@
 from harvester.harvest import HarvestSource
 
 
+@pytest.fixture
+def valid_iso_2_record(
```
rshewitt, line 8: +1 for turning this into a fixture

harvester/harvest.py
tdlowden: @rshewitt correct. dcatus providers should still be held to the schema reqs more strictly

jbrown-xentity (dismissed):
Great tests, LGTM!

FuhuXia (approved):

PR #311, opened 2025-07-17
Catch exceptions in load manager

neilmb:
We had a problem recently where a CloudFoundry bug was leading to exceptions from the CloudFoundry client in `cf_handler.get_all_app_tasks`. This adds exception handling and tests for that particular error situation. BUT this error was also preventing our web app from starting up, so I also added a very broad try-clause around `load_manager.start()` so that any errors in that phase wouldn't stop the web app from coming up.

### DIFF
--- a/harvester/lib/cf_handler.py
+++ b/harvester/lib/cf_handler.py
@@ -5,6 +5,7 @@ import os
 import re
 
 from cloudfoundry_client.client import CloudFoundryClient
+from cloudfoundry_client.errors import InvalidStatusCode
 
 logger = logging.getLogger("harvest_admin")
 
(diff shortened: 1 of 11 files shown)


harvester/lib/cf_handler.py
```diff
@@ -71,9 +72,16 @@ def get_all_app_tasks(self, app_guuid=None):
...
-        )
+        try:
+            result_iter = self.client.v3.apps._pagination(
+                self.client.v3.apps.get(app_guuid, "tasks")
+            )
+        except InvalidStatusCode:
+            logger.warning(
+                "Failed to get app tasks from CF, task information is not accurate"
+            )
+            return []
```
jbrown-xentity: So if we return an empty array here, won't this then spin up a full set of jobs (say 3)? And on next startup, if the Task list is still down, it will spin up 3 more?
Just to be clear, we have https://github.com/GSA/datagov-harvester/blob/main/harvester/lib/load_manager.py#L91C38-L91C59, which gets from num_running_app_tasks, which calls get_running_app_tasks, which calls this function. We need a way to propagate that up in a way that allows us to recognize we shouldn't spin up new tasks, I would think.

jbrown-xentity (commented):
The tests look good, but I have a concern about the default case, returning an empty array...

harvester/lib/cf_handler.py
neilmb: Sigh, yes that could happen. I re-plumbed everything up from `get_all_app_tasks` using `None` as a sentinel value for when the API isn't available. Please re-review.

jbrown-xentity (approved):
Looks great! Sorry that cascaded into so many locations...

PR #312, opened 2025-07-18
improve validation messages

rshewitt:
# Pull Request

Related to [#5206](https://github.com/GSA/data.gov/issues/5206)

## About
- improve validation messages by digging into the exact reason(s) for failure. this particularly improves errors with nested content (e.g. distributions) where the original error would be vague to the effect of "there's something wrong with distributions" which isn't helpful at all. this instead will return exactly what's wrong by providing the offending distribution and field (for example).
- update to the ISO19115_1 fixture to reflect update to the email in contactPoint processing in transformation app (lowercase) 

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/harvester/utils/general_utils.py
+++ b/harvester/utils/general_utils.py
@@ -5,6 +5,7 @@ import logging
 import os
 import re
 import smtplib
+from collections import defaultdict
 from datetime import datetime, timedelta, timezone
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
@@ -15,6 +16,7 @@ import geojson_validator
 import requests
 import sansjson
 from bs4 import BeautifulSoup
+from jsonschema.exceptions import ValidationError
 from requests.adapters import HTTPAdapter
 from urllib3.util.retry import Retry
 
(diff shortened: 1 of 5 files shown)


neilmb (dismissed):
This looks fine as far as it goes, but I don't see that `assemble_validation_messages` is used anywhere. It might be a useful utility function, but if we are going to include it in a PR, should we call it somewhere besides the unit test?

tests/integration/harvest/test_validate.py
```diff
@@ -113,13 +119,23 @@ def test_multiple_invalid(
...
             e[0] for e in interface.get_harvest_record_errors_by_job(harvest_job.id)
         ]
-        assert len(errors) == 3
-        assert errors[-1].message.startswith(
-            "<ValidationError: '\"<p align=\\'center\\'"
-        )
-        assert errors[-1].message.endswith(
-            "is not valid under any of the given schemas'>"
-        )
+        assert len(errors) == 8
```
rshewitt: before this change there were 3 errors but with this improvement we're given the exact rules for why it failed (the full strings aren't listed for brevity)
1. contactEmail 
2.  distribution 
3. license.

harvester/utils/general_utils.py
```diff
@@ -496,3 +497,53 @@ def send_email_to_recipients(recipients, subject, body):
...
+def assemble_validation_errors(validation_errors: list, messages=None) -> list:
+    """
+    given a list of errors, follow each one recursively through its context
+    and get the simplest cause for error. errors with lists or dicts (other than empty)
+    will often return the entire object followed by 'is not valid under any of the given schemas'
+    which isn't helpful.
+    """
+
+    if messages is None:
+        messages = []
```
jbrown-xentity: Probably could just use an empty array in the function definition as the default if not passed...

tests/integration/harvest/test_validate.py
jbrown-xentity: So there will be a distinct error for each rule, not each key? That's a little confusing... Can we somehow wrap these into the key? I wouldn't expect there to be multiple errors if their contactEmail is bad. I would expect one, and the error to say "this doesn't match our possible schemas/values/format, possible formats are ...".

jbrown-xentity (commented):
Getting there. I like that you put this as a util function; we may want to report these errors in some type of validation/test format in the future (and it will be handy that it's not just embedded in the harvest logic).

tests/integration/harvest/test_validate.py
rshewitt: what's important is how this information is presented to the data providers
- first, if they're all bundled up.
Screenshot 2025-07-23 at 8 36 29 AM


- second, if they're split individually
Screenshot 2025-07-23 at 8 32 15 AM

harvester/utils/general_utils.py
rshewitt: i've steered away from using mutable types as defaults because they can produce weird results if the function is called using the default which isn't how we're doing it but it's considered bad practice.

tests/integration/harvest/test_validate.py
neilmb: I like the first presentation much better. I feel like the cognitive load of the second presentation scales wrong because of the repeated "subject" in each sentence. I don't see that there will be an easy programmatic way in a template to group three separate errors into the first display.

rshewitt: we could treat the bundled message as 1 validation error instead of 3. the template could split by `:` then `,` to bolden each format? we can identify the formats for bundling because they're either in quotes or "non-empty" in the case for a list

jbrown-xentity: I think it would make more sense if we bundle when reporting the error, not after the fact. If accessURL is a string like "file:///some-local-file", I would expect to bundle up the types of things it can be as in the first example.
Can we look at the beginning of the string to try to bundle these into a single record with all the possible types of values it can be? Then we're creating a valuable message.
If it's a different field then it's a different error with a different message.

rshewitt: i think we're saying the same thing. using the original example we would have 3 validation record errors in the db. each one would bundle the acceptable formats.
1. contactEmail does not match any of the acceptable formats: ^mailto:[\\\\\\\\w\\\\\\\\_\\\\\\\\~\\\\\\\\!\\\\\\\\$\\\\\\\\&\\\\\\\\\\\'\\\\\\\\(\\\\\\\\)\\\\\\\\*\\\\\\\\+\\\\\\\\,\\\\\\\\;\\\\\\\\=\\\\\\\\:.-]+@[\\\\\\\\w.-]+\\\\\\\\.[\\\\\\\\w.-]+?$, ^(\\\\\\\\[\\\\\\\\[REDACTED).*?(\\\\\\\\]\\\\\\\\])$
2. distribution ( and so on...)
3. license ( and so on...)

tests/integration/harvest/test_validate.py
```diff
@@ -83,12 +83,12 @@ def test_invalid_license_uri_dcatus_non_federal(
...
-            "<ValidationError: '\"<p align=\\'center\\'"
-        )
-        assert errors[0].message.endswith(
-            "is not valid under any of the given schemas'>"
-        )
+
+        assert len(errors) == 1
+
+        # ruff: noqa E501
+        expected = "<ValidationError: \"$.license, 'center' does not match any of the acceptable formats: 'uri', 'null', '^(\\\\\\\\[\\\\\\\\[REDACTED).*?(\\\\\\\\]\\\\\\\\])$'\">"
```
rshewitt, line 90: the regex only pulled "center" from the html value used in the license. getting the entire html content will most likely complicate the regex used to find the invalid data. updating it to look for either single or double quotes reads "<p align=\\\'" instead of "center"

rshewitt: regexes always grow more complicated as more edge cases are handled which makes them increasingly harder to read

neilmb:
@rshewitt What do we need to get this merged? I'd like to have something, anything "better" than we have right now and then we can iterate on the UX once it is in front of users.

rshewitt:
> @rshewitt What do we need to get this merged? I'd like to have something, anything "better" than we have right now and then we can iterate on the UX once it is in front of users.

it's ready right now. it treats ever missing required property at the root (e.g. identifier or modified) as an individual validation record error in the db. it bundles things like "contactPoint.hasEmail" into 1 record.my comment about the regex is an edge case. i think in a significant majority of cases it works just fine.

neilmb (approved):

PR #315, opened 2025-07-21
add placeholder describedByType

rshewitt:
# Pull Request

Related to [#5335](https://github.com/GSA/data.gov/issues/5335)

## About
- some ISO docs are failing validation because of bad `describedByType` values. see [job](https://datagov-harvest-dev.app.cloud.gov/harvest_job/2e3eb103-9a3b-4c7a-88e3-c242936eadd7).
- this PR uses `application/octet-stream` as the placeholder when "describedBy" is present. this value is recommended for unknown file types: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -771,6 +771,8 @@ class Record:
         """Fill in placeholder values to prevent some validation errors.
 
         We work directly on the self.transformed_data dict.
+
+        This function is only called on ISO records. It is not applied on DCATUS.
         """
         # missing contactPoint or it's empty
         if not self.transformed_data.get("contactPoint"):
@@ -791,6 +793,11 @@ class Record:
             self.transformed_data["publisher"] = {
                 "name": self.harvest_source.get_source_orm().org.name
             }
+
+        if self.transformed_data.get("describedBy"):
+            # recommended for unknown files types
+            # https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types
+            self.transformed_data["describedByType"] = "application/octet-stream"
 
     def validate(self) -> None:
         # TODO: create a different status for transformation exceptions
--- a/tests/integration/harvest/test_validate.py
+++ b/tests/integration/harvest/test_validate.py
@@ -167,3 +167,18 @@ class TestValidateDataset:
         assert valid_iso_2_record.transformed_data["publisher"] == {
             "name": organization_data["name"]
         }
+
+    def test_transformed_iso_described_by_type_placeholder(self, valid_iso_2_record):
+        valid_iso_2_record.transform()
+        valid_iso_2_record.transformed_data["describedBy"] = (
+            "https://geodesy.noaa.gov/CORS/"
+        )
+
+        # now fill in the missing items
+        valid_iso_2_record.fill_placeholders()
+        valid_iso_2_record.validate()  ## passes
+
+        assert (
+            valid_iso_2_record.transformed_data["describedByType"]
+            == "application/octet-stream"
+        )


harvester/harvest.py
```diff
@@ -792,6 +794,11 @@ def fill_placeholders(self) -> None:
                 "name": self.harvest_source.get_source_orm().org.name
             }
 
+        if self.transformed_data.get("describedBy"):
+            # recommended for unknown files types
+            # https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types
+            self.transformed_data["describedByType"] = "application/octet-stream"
+
```
jbrown-xentity, line 801: I have some concerns about this approach. This doesn't seem like a default, it seems to be overwriting.
I feel like we also need to check if describedByType is filled out. If it is, run it through the possible regex that the validator would (see https://github.com/GSA/datagov-harvester/blob/main/schemas/iso-non-federal_dataset.json#L86-L119). If the value doesn't conform, THEN provide sane defaults. This approach could overwrite the value provided, which is aggressive.

tests/integration/harvest/test_validate.py
```diff
@@ -167,3 +167,18 @@ def test_transformed_iso_publisher_placeholder(
...
+        )
+
+        # now fill in the missing items
+        valid_iso_2_record.fill_placeholders()
+        valid_iso_2_record.validate()  ## passes
+
+        assert (
+            valid_iso_2_record.transformed_data["describedByType"]
+            == "application/octet-stream"
+        )
```
jbrown-xentity, line 184: 👍 to tests

jbrown-xentity (commented):
I feel like this needs a deeper dive.

harvester/harvest.py
rshewitt: `describedByType` in the dcatus [writer](https://github.com/GSA/mdTranslator/blob/d216d932c8ca61afd9696f0a7da14d398783d5e7/lib/adiwg/mdtranslator/writers/dcat_us/sections/dcat_us_described_by_type.rb#L17) reads the protocol of the data dictionary online resource (the writer has always behaved this way). this is often times "HTTPS" or "WWW:LINK-1.0-http--link". `describedByType` is the mimetype of the resource. "applicationProfile" probably makes more sense (see example below. this is a real doc we process) although it would need to be translated into a [text/[text] mimetype format (e.g. text/plain) which we have logic for. in summary, the value we get from mdtranslator is wrong and I figure it's less change/work to have the runner do it this way. we don't even do anything with this data anyways. it's simply meant to allow validation to pass. 

```xml
<gmd:CI_OnlineResource uuid="00362270-9ce0-11e0-aa82-0800200c9a66">
          <gmd:linkage>
      <gmd:URL>http://www.ngdc.noaa.gov/metadata/published/xsd/schema/resources/extensions/MD_MetadataExtensionInformation.xml</gmd:URL>
          </gmd:linkage>
          <gmd:applicationProfile>
            <gco:CharacterString>xml</gco:CharacterString>
          </gmd:applicationProfile>
          <gmd:name>
            <gco:CharacterString>MD_MetadataExtensionInformation.xml</gco:CharacterString>
          </gmd:name>
          <gmd:description>
            <gco:CharacterString>Metadata Extension Information</gco:CharacterString>
          </gmd:description>
          <gmd:function>
            <gmd:CI_OnLineFunctionCode codeList="http://www.ngdc.noaa.gov/metadata/published/xsd/schema/resources/Codelist/gmxCodelists.xml#CI_OnLineFunctionCode" codeListValue="information">information</gmd:CI_OnLineFunctionCode>
          </gmd:function>
</gmd:CI_OnlineResource>
```

jbrown-xentity: I don't like that MDTranslator is providing a bad translation. If that field doesn't exist in the same way in ISO, then we should just remove it from the either the ISO reader or the DCAT-US writer.
Once that is in place, we can let it be blank, which will also pass validation.
What I don't want to happen is an improvement comes in the future that allows this data to flow through, only for us to find out we overwrite it without even checking the value of the field itself.

rshewitt: i'm okay with that (changing describedByType in mdtranslator). i'll close this PR.

PR #316, opened 2025-07-22
Iterate all validation errors

neilmb:
# Pull Request

In GSA/data.gov#5332 we need to record all of the validation errors present in each record, instead of just recording the first error. This uses the `jsonschema` library's `iter_errors` method to loop over all of the schema validation errors and write each one of them into the database as a `HarvestRecordError`.

## About

This refactors the validation process to stop using a `ValidationException` and instead call a `log_non_critical_error` helper method to make it easier to log multiple things instead of exceptions which are basically one-and-done. It also refactors `harvester.harvest.Record` to remove the `valid` and `validation_msg` attributes which was never used after validation. Instead, the `Record.validate()` method now returns a boolean that we can assert in our tests, and the messages are available from the database with the harvest job ID.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


harvester/exceptions.py
```diff
@@ -51,43 +53,38 @@ class SendNotificationException(HarvestCriticalException):
...
 
 class ExternalRecordToClass(HarvestNonCriticalException):
     pass
 
 
-class ValidationException(HarvestNonCriticalException):
-    def log_err(self):
-        pass
-
-
```
jbrown-xentity, line 90: Why do we want to remove this entirely? Isn't there a scenario where the validation fails (ie there is no title) and we want the normal flow to occur (exception thrown, error logged in DB, record updated as error, and log printed)?

harvester/harvest.py
```diff
@@ -792,30 +769,57 @@ def fill_placeholders(self) -> None:
...
+            return True
+        else:
+            # update the reporter only once even with multiple errors
             self.harvest_source.reporter.update("errored")
-            raise ValidationException(
-                repr(e),
-                self.harvest_source.job_id,
-                self.id,
-            )
+            return False
```
jbrown-xentity, line 822: If we're setting the return value to be truthy instead of throwing an exception, shouldn't we have to change this line here? https://github.com/GSA/datagov-harvester/blob/main/harvester/harvest.py#L657
So the output would need to be parsed and the sync job would be dependent on the output of the validate function.

jbrown-xentity (commented):
I like the way the tests flow better, so there seems to be improvement to using the "truthy" model instead of the "exception" model.
However, I think we need to update the flow in a few places to handle this fully...
- https://github.com/GSA/datagov-harvester/blob/main/harvester/harvest.py#L657
- https://github.com/GSA/datagov-harvester/blob/main/harvester/harvest.py#L880-L888

harvester/exceptions.py
```diff
@@ -51,43 +53,40 @@ class SendNotificationException(HarvestCriticalException):
...
+        "type": error_type,
+        "date_created": datetime.now(timezone.utc),
+        "harvest_job_id": job_id,
+        "harvest_record_id": record_id,
+    }
 
-        self.db_interface: HarvesterDBInterface = db_interface
-        self.logger = logging.getLogger("harvest_runner")
+    db_interface.add_harvest_record_error(error_data)
+    db_interface.update_harvest_record(record_id, {"status": "error"})
```
rshewitt, line 70: can't "status" be placed in `error_data` instead of adding the record then immediately updating it?

neilmb: Status is being set on the "parent" `HarvestRecord` while the `error_data` is making a new `HarvestRecordError` entry.

harvester/exceptions.py
neilmb: We don't raise this exception anywhere so this is dead code now. If we eventually need this exception version again, we can make it then.

harvester/harvest.py
neilmb: `sync` uses a guard for when `status == "error"` https://github.com/GSA/datagov-harvester/blob/main/harvester/harvest.py#L822 that still works. The new return value could be used to prevent the function call, but I also think that this is a more minimal intervention.

jbrown-xentity (approved):
LGTM

PR #317, opened 2025-07-23
Improve the logic to remove Orphaned Jobs

cmhedrickREI:
# Pull Request

Related to GSA/Data.gov#5254

## About

- Updates the logic to find orphaned jobs by when last time the job record was updated as the jobs should be producing logs and having the record updated over time. If the updated time is stale then it can be assumed that the task is stale
- The code now not only updated the harvest job records in the db to reflect the job stalling, but also now stops the job if it's still running.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (dismissed):
LGTM!

scripts/orphan_job_clean_up.py
```diff
@@ -47,44 +90,80 @@ def delete_jobs(job_id: Optional[Tuple[str]], dry_run: bool) -> None:
...
+    handler = CFHandler(CF_API_URL, CF_SERVICE_USER, CF_SERVICE_AUTH)
+    jobs = []
+
+    # get RUNNING tasks from CF
+    try:
+        tasks = handler.get_running_app_tasks()
+    except Exception as e:
+        click.echo(f"Error fetching tasks: {e}")
+        return
+    # import pdb
```
neilmb: **suggestion (non-blocking)**: We don't usually keep around commented-out debugging code. Take this out?

neilmb (dismissed):

scripts/orphan_job_clean_up.py
cmhedrickREI: d'oh! :facepalm:

neilmb (approved):

PR #319, opened 2025-07-23
Guess a better distribution URL

neilmb:
# Pull Request

Related to GSA/data.gov#5307, we observe lots of records that fail validation because their `distribution` items don't have valid `downloadURL` or `accessURL` when they just left off the `https://` scheme from the URL. This change adds that `https://` scheme at the beginning if it makes an invalid URL into a valid URL.

## About

Since we had the `jsonschema` library already, we can use it to check on what is a valid URL by using a tiny little schema.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -616,6 +616,15 @@ class Record:
             raise ValueError("status must be a string")
         self._status = value
 
+    @staticmethod
+    def _is_valid_url(url: str) -> bool:
+        """Return whether a string is a valid URL."""
+        return Draft202012Validator(
+            {"type": "string", "format": "uri"},
+            format_checker=Draft202012Validator.FORMAT_CHECKER,
+        ).is_valid(url)
+
+
     def harvest(self) -> None:
         """
         this is the main harvest function for a record instance. it runs the compare,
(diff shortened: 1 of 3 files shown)


harvester/harvest.py
```diff
@@ -769,6 +769,28 @@ def fill_placeholders(self) -> None:
...
             }
 
+        # If distribution items have a downloadURL or accessURL,
+        # check if it just needs an "https://" at the beginning
+        # to be valid
+        def _is_valid_url(url):
+            return Draft202012Validator(
+                {"type": "string", "format": "uri"},
+                format_checker=Draft202012Validator.FORMAT_CHECKER,
+            ).is_valid(url)
```
jbrown-xentity: non-blocking: I would put this function definition in the general class. I believe this may not be the last place/time we want to validate a URI...

jbrown-xentity (dismissed):
Tests look great! Small quibble, all looks good and is ready to 🚢 on your decision.

harvester/harvest.py
neilmb: I made it a static method of the `Record` class. If we eventually need it elsewhere, we can move it in the future.

Re-review?

jbrown-xentity (approved):
LGTM!

PR #321, opened 2025-07-28
Add restarts and task increases for staging and prod

neilmb:
# Pull Request

To run our scheduled harvest tasks in staging and production, we need the harvest applications on cloud.gov to be restarted regularly. This also adds an increase in the number of simultaneous tasks we use on each environment.

## About

We should monitor our memory usage after this change to make sure that the number of tasks we specify here doesn't use too much, but it shouldn't be a problem with our current quota.

### DIFF
--- a/.github/workflows/restart.yml
+++ b/.github/workflows/restart.yml
@@ -13,8 +13,12 @@ permissions:
 
 jobs:
   restart:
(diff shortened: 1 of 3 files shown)


jbrown-xentity (approved):
LGTM

PR #324, opened 2025-07-30
add egress check

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5352

## About

Add egress check cron. The check will fail until egress is implemented. 

The workflow is copied from gsa/catalog.data.gov. If this needs to be implemented to yet another repo, it will be time to  create a workflow template on our main repo /gsa/data.gov so that we don't repeat ourselves.

### DIFF
--- a/.github/workflows/egress-check.yml
+++ b/.github/workflows/egress-check.yml
@@ -1,0 +1,63 @@
+---
+name: Check Egress Operation
+
+on:
+  schedule:
+    - cron: '40 02,14 * * *'
+  workflow_dispatch:
+
+jobs:
+  egress-check:
+    name: ${{matrix.environ}} - ${{matrix.command}}
+    runs-on: ubuntu-latest
+    environment: ${{ matrix.environ }}
+    strategy:
+      fail-fast: false
+      max-parallel: 2
+      matrix:
+        command:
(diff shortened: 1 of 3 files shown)


.github/workflows/egress-check.yml
```diff
@@ -0,0 +1,63 @@
...
+            --name $INSTANCE_NAME -k 1500M -m 150M
+          cf_org: gsa-datagov
+          cf_space: ${{ matrix.environ }}
+          cf_username: ${{secrets.CF_SERVICE_USER}}
+          cf_password: ${{secrets.CF_SERVICE_AUTH}}
+      - name: monitor task output
+        uses: cloud-gov/cg-cli-tools@main
+        with:
+          command: >
+            tools/monitor_cf_logs.sh datagov-harvest $INSTANCE_NAME
```
neilmb: I don't see this file in this repository? Is this workflow copied from somewhere else and we need to bring in this shell script as well? Maybe `bin/` is the appropriate place for it here.

neilmb (changes requested):
This looks reasonable, but I think it's missing a file.

.github/workflows/egress-check.yml
FuhuXia: file monitor_cf_logs.sh added.

neilmb (approved):

FuhuXia:
Got error:
```
bin/monitor_cf_logs.sh: Permission denied
```

PR #328, opened 2025-07-31
add qa action 

rshewitt:
# Pull Request

Related to [#5274](https://github.com/GSA/data.gov/issues/5274)

## About
- add a manual dispatch action for running the QA script. zips the output and archives it in the action as `qa_archive`
- basic auth has been put in front of all catalog-next instances so i configured a requests session with those credentials on every request in the qa script
- apparently the action won't show up until it's on the default branch (main) which seems odd that i have to debug it (if something breaks) after it's merged in (maybe i'm not doing something right?)

## PR TASKS

- [x] Code well documented
- [ ] Files linted

### DIFF
--- a/.github/workflows/qa.yml
+++ b/.github/workflows/qa.yml
@@ -1,0 +1,59 @@
+name: Harvester QA script
+
+env:
+  PY_VERSION: "3.13"
+  POETRY_VERSION: "2.0.0"
+
+on:
+  workflow_dispatch:
+    inputs: 
+      job_type:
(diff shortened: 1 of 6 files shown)


neilmb (approved):
Excited to see this. Actions can be run from other branches, but they have to have been run successfully once first https://stackoverflow.com/questions/63362126/github-actions-how-can-i-run-a-workflow-created-on-a-non-master-branch-from-t, so testing does have to happen with multiple pull requests. Such is life.

PR #333, opened 2025-08-04
Don't include resources without urls

neilmb:
# Pull Request

Related to GSA/data.gov#5322, we were seeing errors where we tried to create CKAN packages that had "resources" with no `url` at all. This fixes the harvester to skip those resources that don't have either `accessURL` or `downloadURL`.

## About

This logs when this case happens, but doesn't register a record error. If we do get a facility to add "warnings" for catalog producers, it might be nice to include a warning here that they have a value in `distribution[]` that is invalid.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/utils/ckan_utils.py
+++ b/harvester/utils/ckan_utils.py
@@ -824,7 +824,9 @@ def create_ckan_resources(metadata: dict) -> list[dict]:
         # if we didn't find either of the url_keys don't
         # include this resource https://github.com/GSA/data.gov/issues/5322
         if "url" not in resource:
-            logger.info("Not including resource that doesn't contain accessURL or downloadURL")
+            logger.info(
+                "Not including resource that doesn't contain accessURL or downloadURL"
+            )
             continue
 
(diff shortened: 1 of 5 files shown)


harvester/utils/ckan_utils.py
```diff
@@ -821,6 +821,14 @@ def create_ckan_resources(metadata: dict) -> list[dict]:
             # if we know the mimetype add the other details
             if resource["mimetype"]:
                 change_resource_details(resource=resource)
+        # if we didn't find either of the url_keys don't
+        # include this resource https://github.com/GSA/data.gov/issues/5322
+        if "url" not in resource:
+            logger.info(
+                "Not including resource that doesn't contain accessURL or downloadURL"
```
rshewitt: do we want to include the record identifier here?

neilmb: Good suggestion, done.

rshewitt (approved):
good addition

PR #337, opened 2025-08-07
Check for more work even if failure

jbrown-xentity:
Even if the job fails, it should still check for more work so the slot can be filled.

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5373

## About

Noticed in real time, multiple failures resulted in harvester not running for long periods.

## PR TASKS

- [ ] ~Code well documented~
- [ ] ~Files linted~

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -939,7 +939,16 @@ def check_for_more_work():
     At the end of the harvest job, look for whether there are still new
     jobs to be done and schedule at most one new task.
     """
-    LoadManager()._start_new_jobs(check_from_task=True)
+    try:
+        LoadManager()._start_new_jobs(check_from_task=True)
+    except Exception as e:
+        logger.error(f"Error checking for more work: {repr(e)}")
+        # We don't want to raise here, just log the error and continue
+        # This is to ensure that the harvest job can finish gracefully
+        # even if there is an issue with checking for more work.
+        # The application should pick up jobs every 15 minutes,
+        # this is only for speed.
+        return
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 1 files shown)


harvester/harvest.py
```diff
@@ -953,4 +953,5 @@ def check_for_more_work():
         check_for_more_work()
     except Exception as e:
         logger.error(f"Harvest has experienced an error :: {repr(e)}")
+        check_for_more_work()
```
rshewitt, line 954: what if `check_for_more_work` caused the exception?

jbrown-xentity: Then it will hard error out. The next line of this is `exit(1)`, so a hard error is coming either way @rshewitt. I think this is still safe to merge.

rshewitt: i don't think it will. you'd get "During handling of the above exception, another exception occurred:" if `check_for_more_work()` threw the exception in the first place. for example...
```terminal
>>> def f():
...     raise Exception
...     
>>> try:
...     f()
... except Exception as e:
...     f()
...     
Traceback (most recent call last):
  File "<python-input-1>", line 2, in <module>
    f()
    ~^^
  File "<python-input-0>", line 2, in f
    raise Exception
Exception

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<python-input-1>", line 4, in <module>
    f()
    ~^^
  File "<python-input-0>", line 2, in f
    raise Exception
Exception
>>> try:
...     f()
... except Exception as e:
...     pass
...     
>>> 
```
if we feel that `check_for_more_work` is stable enough that an exception won't happen you could put it under the finally statement in the try/except so it runs in either situation

jbrown-xentity: I think this latest change addresses your feedback @rshewitt , can you take a look?

rshewitt (approved):
would be good to have a custom exception for checking more work but in the effort of time and to minimize change this is sufficient

PR #349, opened 2025-08-14
Add invalid lengths (sane maximums) for metadata fields

jbrown-xentity:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5325

## About

Adds real-world test cases, and makes sure they are caught in validation.
Also required some cleanup in utilization/reference of example files.

## PR TASKS

- [x] Code well documented
- [x] Files linted


FuhuXia (approved):
LGTM

PR #350, opened 2025-08-14
only delete from h20 db if deleted on ckan

rshewitt:
# Pull Request

Related to [#5385](https://github.com/GSA/data.gov/issues/5385)

## PR TASKS

- [x] Code well documented
- [x] Files linted


jbrown-xentity (approved):
LGTM, although I think there are even bigger issues to address to make sure this functions properly.

For example, we're running this for EVERY record. Including every "update" record, and every "delete" record.
We should parse these out and do the right thing for each type of record. See sketch here: https://github.com/GSA/data.gov/issues/5384.

That being said, this is definitely an improvement, and can be merged as-is.

PR #354, opened 2025-08-18
Sync Tool

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5211

## About

- Adds sync script that should sync CKAN with the DB ensuring that Harvester records are reflected with in CKAN solr, or removes them if no longer in the db.
- Adds sync job that should be able to be executed via github actions

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+        self.ckan_api_url = CKAN_API_URL
+        self.ckan_api_key = CKAN_API_TOKEN
+        self.db_interface = db_interface
+        self.batch_size = batch_size
+        self.headers = {
+            "Authorization": f"Bearer {CKAN_API_TOKEN}",
+            "Content-Type": "application/json",
+        }
+        session = create_retry_session()
+        self.ckan_tool = CKANSyncTool(session=session)
```
jbrown-xentity: Non-blocking: You could probably add the header information into the session here, instead of saving it and adding it later.
See example https://github.com/GSA/datagov-harvester/pull/355/files#diff-332a4b024a46ac5c9b23740ee07ef353316e534e851f894eab4e4b8559aa558fR489-R491

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+            result = self.db_interface.db.execute(stmt)
+            batch_ids = [row[0] for row in result.fetchall()]
+
+            if not batch_ids:
+                break
+
+            all_ids.update(batch_ids)
+            offset += len(batch_ids)
+
+            click.echo(f"Loaded {len(all_ids)} harvest_object_ids from DB...")
```
jbrown-xentity: This seems to get all harvest records, not just the latest records.
Could we use this, or something like it? https://github.com/GSA/datagov-harvester/blob/main/database/interface.py#L546

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+            # If harvest_object_id exists in DB, already synced
+            if harvest_id and harvest_id in db_harvest_ids:
+                synced.append(record)
+                # if it's in we don't need to add it.
+                to_add.discard(identifier)
+            # If identifier is in the db identifiers we can remove the identifier
+            # from the to_add set, as it means we have a record.
+            elif identifier and identifier in db_identifiers:
+                # discard is the safe method
+                to_add.discard(identifier)
```
jbrown-xentity: I'm a little confused by this logic. I can't really comprehend a scenario where the harvest_id exists on CKAN but not in our DB, but the identifier does? I would think in that use case that we need to update this record, because it seems to be out of sync.

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+                to_update.append(record)
+        return synced, to_update, to_delete, to_add
+
+    def get_db_record_by_identifier(self, identifier: str) -> Optional[HarvestRecord]:
+        """
+        Get database record by identifier using the DB interface.
+        """
+        stmt = select(HarvestRecord).where(HarvestRecord.identifier == identifier)
+        result = self.db_interface.db.execute(stmt)
+        return result.scalar_one_or_none()
```
jbrown-xentity, line 271: I don't think we need this function, just use https://github.com/GSA/datagov-harvester/blob/main/database/interface.py#L511

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+                    "Warning: DB record not found for identifier "
+                    f"{record['identifier']}"
+                )
+                return False
+
+            # Prepare update data
+            update_data = {
+                "id": record["id"],
+                "extras_harvest_object_id": db_record.harvest_object_id,
+            }
```
jbrown-xentity: This just patches with the most recent harvest object id, but not anything else? What if the title changed? The description? If we have a more recent record in our DB than what is on CKAN, then something changed... What is that change that we should make?
Don't need to recreate the wheel here. Can we leverage the [Record Class](https://github.com/GSA/datagov-harvester/blob/main/harvester/harvest.py#L497) and the some part of the [harvest function](https://github.com/GSA/datagov-harvester/blob/main/harvester/harvest.py#L617) to sync this update to CKAN?

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+        """
+        Delete a CKAN record that doesn't exist in the database.
+        """
+        try:
+            response = requests.post(
+                f"{self.ckan_api_url}/api/action/package_delete",
+                json={"id": record["id"]},
+                headers=self.headers,
+                timeout=30,
+            )
```
jbrown-xentity: Why not use the ckan_tool here that you built, that has retry built in?

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+                "id": record["id"],
+                "extras_harvest_object_id": db_record.harvest_object_id,
+            }
+
+            response = requests.post(
+                f"{self.ckan_api_url}/api/action/package_patch",
+                json=update_data,
+                headers=self.headers,
+                timeout=30,
+            )
```
jbrown-xentity: Why not use the ckan_tool here, that has the retry built in and has the CKAN API routes built in?

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+
+        with click.progressbar(records_to_update, label="Updating records") as bar:
+            for record in bar:
+                if self.update_ckan_record(record):
+                    success_count += 1
+                else:
+                    failure_count += 1
+
+                # Small delay to avoid overwhelming the API
+                time.sleep(0.1)
```
jbrown-xentity: Let's use exponential backoff in retry instead.

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+
+        return success_count, failure_count
+
+    def get_db_record_count(self) -> int:
+        """Get total count of harvest records in database."""
+        stmt = select(func.count(HarvestRecord.id))
+        result = self.db_interface.db.execute(stmt)
+        return result.scalar()
```
jbrown-xentity, line 483: If we try to get every record we've ever had, this could easily be in the millions very soon (it might already be there). That's too much data to "sync". We need only the "latest" records.

scripts/sync_db.py
```diff
@@ -0,0 +1,543 @@
...
+
+
+@click.command()
+@click.option("--batch-size", default=1000, help="Batch size for processing (<=1000)")
+@click.option(
+    "--dry-run", is_flag=True, help="Analyze only, make no changes", default=True
+)
+@click.option(
+    "--preview", is_flag=True, help="Show detailed preview of changes", default=True
+)
```
jbrown-xentity: What is the difference between dry-run and preview? Lost on this count...

jbrown-xentity (commented):
Have some comments/questions about how we're doing this below. Couple of high level comments/questions:
- Have you run this against a DB and CKAN instance? If so, how many records did it process? How long did it take?
- I think for something like this, an artifact more than just the logs (like a CSV file) would be very helpful. The overall stats can be kept in the logs, but the count of records and their success should be tracked in a list/object and dumped to a CSV on the job, so that it can be reviewed fully and in detail. You can see an example of doing this here: https://github.com/GSA/datagov-harvester/blob/main/.github/workflows/qa.yml#L49-L59

.github/workflows/sync.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: abatilo/actions-poetry@v3
+        with:
+          poetry-version: ${{ env.POETRY_VERSION }}
+
+      - name: Install Dependencies
+        run: |
+          poetry env use ${{ env.PY_VERSION }}
+          poetry install
+      
+      - name: Run QA
```
neilmb: **suggestion (non-blocking)**: This isn't the QA job, maybe:
```suggestion
      - name: Run Sync script
```

.github/workflows/sync.yml
```diff
@@ -0,0 +1,61 @@
...
+          args=''
+          if ${{ inputs.dry_run }}; then
+            args="${args} --dry_run"
+          fi
+
+          if ${{ inputs.preview }}; then
+            args="${args} --preview"
+          fi
+
+          poetry run python ./scripts/sync_db.py ${args}
```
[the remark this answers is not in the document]
neilmb: This suggestion makes good sense.

scripts/sync_db.py
```diff
@@ -0,0 +1,687 @@
...
+            record.get("metadata_modified"), db_record.date_finished
+        ):
+            return True
+        return False
+
+    def categorize_ckan_records(
+        self,
+        ckan_records: List[Dict[str, Any]],
+        db_records: Dict[str, HarvestRecord],
+    ) -> Tuple[
```
neilmb: **suggestion (non-blocking)**: This is a pretty big tuple to remember the order of things, maybe use a little `namedtuple` for the return type to be more explicit about "(synced, to_update, to_delete records and to_add identifiers)"?

scripts/sync_db.py
```diff
@@ -0,0 +1,687 @@
...
+        synced = []  # harvest_object_id exists in DB - already synced
+        to_update = []  # identifier exists in DB but harvest_object_id doesn't
+        to_delete = []  # neither identifier nor harvest_object_id exists in DB
+        processed_identifiers = set()
+
+        for record in ckan_records:
+            identifier = record.get("identifier")
+            db_record = db_records.get(identifier)
+
+            if not db_record or (db_record.action == "delete"):
```
neilmb, line 241: **suggestion (non-blocking)**: These seem like the main cases in the sync process. Can we include plain-language comments on each of the if/elif clauses about the case that they are handling?

scripts/sync_db.py
```diff
@@ -0,0 +1,687 @@
...
+
+        if not dry_run:
+            # Perform synchronization
+            stats = sync_manager.synchronize(dry_run=dry_run)
+
+            # Print results of the synchronization
+            print_sync_results(stats)
+        else:
+            # Generate CSV report even in dry run mode
+            stats = sync_manager.synchronize(dry_run=dry_run)
```
neilmb: **question (non-blocking)**: It looks like this gets called exactly the same in both branches of the if, maybe call it first and then do just the `print_sync_results` call inside of the `if`?

tests/scripts/test_sync_db.py
```diff
@@ -0,0 +1,511 @@
...
+        assert manager.ckan_api_key == "test-token"
+        assert manager.db_interface == mock_db_interface
+        assert manager.batch_size == 1000
+
+    def test_is_within_5_minutes_true(self, sync_manager):
+        """Test is_within_5_minutes returns True for times within 5 minutes."""
+        base_time = datetime(2025, 6, 28, 16, 55, 0, tzinfo=timezone.utc)
+        iso_string = "2025-06-28T16:57:00.000Z"  # 2 minutes later
+
+        assert sync_manager.is_within_5_minutes(iso_string, base_time) is True
```
neilmb: **suggestion (non-blocking)**: You can just `assert <thing>` instead of `assert <thing> is True`.

tests/scripts/test_sync_db.py
```diff
@@ -0,0 +1,511 @@
...
+
+    def test_is_within_5_minutes_false(self, sync_manager):
+        """Test is_within_5_minutes returns False for times beyond 5 minutes."""
+        base_time = datetime(2025, 6, 28, 16, 55, 0, tzinfo=timezone.utc)
+        iso_string = "2025-06-28T17:05:00.000Z"  # 10 minutes later
+
+        assert sync_manager.is_within_5_minutes(iso_string, base_time) is False
```
neilmb, line 82: comment (non-blocking)**: It's a toss-up which one is clearer, but `assert not <thing>` works the same.

neilmb (dismissed):
This is lovely, very tiny style comments, but LGTM and merge whenever it feels good to you.

scripts/sync_db.py
cmhedrickREI: This was absolutely cool to learn about btw! Thanks @neilmb!

neilmb (approved):
So good. Thanks for sticking with this one for so long.

PR #355, opened 2025-08-18
Feat/add session

jbrown-xentity:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5386

## About

I attempted to get all usage of external requests. It's possible I missed something, but I believe this covers most/all use cases:

- Traversing WAF's
- Downloading ISO, CSDGM, and DCAT-US documents
- Hitting CKAN API
- Hitting MDTranslator

## PR TASKS

- [x] Code well documented
- [x] Files linted


harvester/harvest.py
```diff
@@ -688,7 +689,7 @@ def transform(self) -> None:
 
         mdt_url = os.getenv("MDTRANSLATOR_URL")
         try:
-            resp = requests.post(mdt_url, json=data)
+            resp = requests.post(mdt_url, json=data, headers={"User-Agent": USER_AGENT})
```
rshewitt, line 674: why do we need to pass our user agent to the transformer?

jbrown-xentity: We don't NEED to. But if in the future we were trying to isolate where traffic was coming from, it's nice to know where it's  coming from (to isolate our own traffic).

tests/unit/test_utils.py
```diff
@@ -552,6 +555,36 @@ def test_process_job_complete_percentage(self, job_data, result):
     def test_is_valid_uuid4(self, job_id, result):
         assert is_valid_uuid4(job_id) == result
 
+    @patch("harvester.utils.general_utils.requests.get")
+    def test_download_file_user_agent(self, mock_get):
+        """Test that download_file includes correct User-Agent header."""
+        mock_response = Mock()
+        mock_response.status_code = 200
+        mock_response.json.return_value = {"test": "data"}
```
neilmb: **suggestion (non-blocking)**: Rather than repeating this dict literal in the assertion, maybe save it as a variable and then use the variable in the final assertion?

tests/unit/test_utils.py
```diff
@@ -552,6 +555,36 @@ def test_process_job_complete_percentage(self, job_data, result):
...
+            "http://example.com/test.json", headers={"User-Agent": USER_AGENT}
+        )
+        assert result == {"test": "data"}
+
+    @patch("harvester.utils.general_utils.requests.get")
+    def test_download_file_xml_user_agent(self, mock_get):
+        """Test that download_file includes correct User-Agent header for XML files."""
+        mock_response = Mock()
+        mock_response.status_code = 200
+        mock_response.content = b"<xml>test</xml>"
```
neilmb: **suggestion (non-blocking)**: as above, don't repeat the literal string.

neilmb (dismissed):
LGTM, only quibbles.

tests/unit/test_utils.py
jbrown-xentity: Good call! addressed here: https://github.com/GSA/datagov-harvester/pull/355/commits/9610d3f45db07eaeac4d29829843bdad2ee85d9b

tests/unit/test_utils.py
jbrown-xentity: Good call! addressed here: https://github.com/GSA/datagov-harvester/pull/355/commits/9610d3f45db07eaeac4d29829843bdad2ee85d9b

neilmb (approved):

PR #356, opened 2025-08-19
Fixes sanitization issue w/ `render_block`

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5361

## About

- uses a custom `render_block()` function that should use the flask default jinja environment similar to when `render_template` is called. 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


harvester/utils/general_utils.py
```diff
@@ -721,3 +722,24 @@ def is_valid_uuid4(uuid_string) -> bool:
...
+    if not getattr(env, "autoescape", None):
+        raise UnsafeTemplateEnvError(
+            "Jinja autoescape is disabled; enable it or use Flask's jinja_env."
+        )
+    template = env.get_template(template_name)
+
+    # Render only the named block (Jinja will still escape vars inside the block)
+    block_gen = template.blocks[block_name]
+    html = "".join(block_gen(template.new_context(context)))
+    return Response(html, mimetype="text/html; charset=utf-8")
```
jbrown-xentity: This all looks fine, but I'm not sure this is the right location. The harvester folder is mostly for the harvest runner, the tasks that get created to process a job.
I would put this somewhere in the app context, either in init, routes, or some other location.

tests/conftest.py
```diff
@@ -1071,3 +1073,48 @@ def dcatus_non_federal_schema():
...
+
+    # Create test templates
+    test_template = """
+{% block test_block %}
+<p>Hello {{ name }}!</p>
+<div>User input: {{ user_input }}</div>
+{% endblock %}
+    """
+
+    template_path = os.path.join(template_dir, "test_template.html")
```
neilmb: **suggestion (non-blocking)**: We prefer `pathlib.Path` and `/` to join paths in some other places, but maybe that isn't as easy to use here with the output of `mkdtemp`.

tests/integration/app/test_routes.py
```diff
@@ -428,3 +430,33 @@ def test_cancel_in_progress_job(
         response = client.get(f"/harvest_job/cancel/{job.id}", headers=headers)
         assert response.status_code == 302
         assert response.location == f"/harvest_job/{job.id}"
+
+
+class TestRenderBlock:
```
neilmb, line 435: **question ( non-blocking)**: Can we have a test case, however small that exercises the new custom exception for unsafe environments?

neilmb (dismissed):
LGTM. One quibble and one request for a second test case.

neilmb (approved):
I love the new test, thank you.

PR #357, opened 2025-08-19
fix clear helper

rshewitt:
# Pull Request

Related to [5384](https://github.com/GSA/data.gov/issues/5384)

## About
- update clear helper

## PR TASKS

- [x] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -22,6 +22,7 @@ from harvester import SMTP_CONFIG, HarvesterDBInterface, db_interface
 from harvester.exceptions import (
     CKANDownException,
     CKANRejectionException,
+    ClearHelperException,
     CompareException,
     DCATUSToCKANException,
     DuplicateIdentifierException,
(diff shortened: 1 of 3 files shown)


cmhedrickREI (approved):
LGTM

harvester/harvest.py
```diff
@@ -471,26 +472,50 @@ def send_notification_emails(self, job_results: dict) -> None:
...
+
+        # get all outdated/historic records and delete them
+        outdated_records = self.db_interface.get_all_outdated_records(
+            days=0, source_id=self.id
         )
-        logger.info(f"{len(db_records)} records to be cleared")
-        for db_record in db_records:
+        logger.info(f"{len(outdated_records)} historic records to be cleared")
+        for record in outdated_records:
+            self.db_interface.delete_harvest_record(record_id=record.id)
```
jbrown-xentity, line 488: This is slow. One at a time deletes could take minutes or hours.
If you can "get" the records, you can have SQLAlchemy delete them instead. This is where we let the DB do the work, and it does it extremely well.
If it fails, we should make a job error and quit (because something deeper is going on and will require manual intervention and/or code changes).

harvester/harvest.py
```diff
@@ -471,26 +472,50 @@ def send_notification_emails(self, job_results: dict) -> None:
...
+                self.db_interface.delete_harvest_record(record_id=record.id)
+            else:
+                clear_is_successful = False
+                not_deleted_count += 1
+
+        if not clear_is_successful:
+            ClearHelperException(
+                f"{not_deleted_count} records unable to be deleted from ckan",
+                self.job_id,
+            )
```
jbrown-xentity, line 518: Looks great, much better!

jbrown-xentity (commented):
Much better than it was before! I would make one change though...

PR #358, opened 2025-08-19
rework clear job

rshewitt:
update clear as if no records existed at source using normal harvest functionality

# Pull Request

Related to [#5384](https://github.com/GSA/data.gov/issues/5384)

## About
- simplifies clear by running normal harvest but with no external records

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


tests/integration/harvest_job_flows/test_harvest_job_clear.py
```diff
@@ -84,8 +84,7 @@ def test_harvest_job_clear_not_synced_with_ckan(
         assert len(interface.pget_harvest_records()) == 1
 
         harvest_source = HarvestSource(job.id, "clear")
-        harvest_source.clear_helper()
+        harvest_source.run_full_harvest()
 
-        # the record wasn't able to be deleted from ckan so it doesn't get
-        # deleted in the h20 db.
-        assert len(interface.pget_harvest_records()) == 1
+        # 1 (original successful create record) + 1 (delete before ckan action)
```
neilmb, line 89: **suggestion (non-blocking)**: Could we add these comments as tests too? Since there's no CKAN here, it would be nice to have as much validation on the effect of `clear` as possible.

neilmb (approved):
LGTM, just one ask for slightly more test assertions.

tests/integration/harvest_job_flows/test_harvest_job_clear.py
rshewitt: the assertion on line 90 confirms the same assertion as before but slightly different. the comments are substituted. in this situation, where ckan is down and deletions are not possible, we run a clear through a normal full harvest. prior to the clear, there's 1 record (from a "create" in a previous job). after the clear there's 2 records: the old successful "create" record and the comparison "delete" record which was not able to succeed. this proves the original record remained untouched. anytime we get the latest records that first "create" would be returned because it succeeded. does this answer your question or is there something else i need to add?

rshewitt: here's the [write_to_compare](https://github.com/GSA/datagov-harvester/blob/6d428c021e209dd11eeb2b05080dd46195d0407a/harvester/harvest.py#L688) i'm referring to

rshewitt: here's a [test](https://github.com/GSA/datagov-harvester/blob/6d428c021e209dd11eeb2b05080dd46195d0407a/tests/integration/harvest_job_flows/test_harvest_job_clear.py#L8) of a successful clear

PR #360, opened 2025-08-19
Bugfix/session ckan api

jbrown-xentity:
See library: https://github.com/ckan/ckanapi/blob/master/ckanapi/remoteckan.py#L35 Requires setting user-agent separate from headers.

# Pull Request

Related to https://github.com/GSA/data.gov/issues/5386

## About

Should be able to see logs to catalog and proxy, which will provide proof for closing this issue. Added a test for this use case.

## PR TASKS

- [x] Code well documented
- [x] Files linted


neilmb (approved):
LGTm, thanks for the test case.

PR #363, opened 2025-08-22
Harden our API facets parameter and validate inputs and escape outputs

neilmb:
# Pull Request

Related to GSA/data.gov#5362, some Snyk findings had us looking at input and output sanitization. This PR got away from me and I ended up fixing many more sanitization issues all together, including a big change to the `facets` argument of our API. 

It isn't safe to include SQL text directly in `facets` so we instead adopt a very simple format: comma-separated entries `column_name op value` that are ANDed together. `op` is the name of an operator from the `sqlalchemy.sql.operators` module. `eq`, `le`, `gt`, etc. are the most commonly used in our code.

We also use `jsonify` for more JSON output and add a `valid_id_required` decorator to our routes to check that the id parameters in our URL routes are possible UUIDs.

## About

This resolves a number of the Snyk tickets as checked by the Snyk CLI with `snyk code test`.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


app/routes.py
```diff
@@ -605,14 +639,15 @@ def add_harvest_source():
...
+@valid_id_required  # TODO: Use an HTML 404 page
 def view_harvest_source(source_id: str):
     htmx_vars = {
         "target_div": "#paginated__harvest-jobs",
         "endpoint_url": f"/harvest_source/{source_id}",
         "page_param": "page",
     }
     harvest_jobs_facets = (
-        f"harvest_source_id = '{source_id}' AND date_created <= '{get_datetime()}'"
+        f"harvest_source_id eq {source_id},date_created le {get_datetime()}"
```
jbrown-xentity, line 650: Why do we have this date_created check? If they pass a valid ID, do we care "when" the harvest source was created?

app/routes.py
```diff
@@ -854,6 +891,8 @@ def delete_harvest_source(source_id):
 @api.route("/harvest_source/harvest/<source_id>/<job_type>", methods=["GET"])
 @login_required
 def trigger_harvest_source(source_id, job_type):
+    if not is_valid_uuid4(source_id):
+        return JSON_NOT_FOUND()
     message = load_manager.trigger_manual_job(source_id, job_type)
```
jbrown-xentity, line 899: Non-blocking: I don't see that this will be a problem as this route already requires login, but if we're sanitizing inputs shouldn't we verify job_type? There is no check in the trigger_manual_job, and the DB model just takes a string... If I were to validate, I would think it would be the list defined in harvest_job_starter function: `harvest`, `force_harvest`, `clear`, or `validate`.

app/routes.py
```diff
@@ -1146,44 +1193,46 @@ def add_harvest_record():
...
 @main.route("/harvest_record/<record_id>/errors", methods=["GET"])
+@valid_id_required
 def get_all_harvest_record_errors(record_id: str) -> list:
     try:
         record_errors = db.get_harvest_record_errors_by_record(record_id)
         return (
-            db._to_dict(record_errors)
-            if record_errors
-            else (STATUS_STRINGS_ENUM["404"], 404)
+            jsonify(db._to_dict(record_errors)) if record_errors else JSON_NOT_FOUND()
```
jbrown-xentity, line 1204: Not sure this is right. There are perfectly valid records with no errors, that shouldn't return a 404. Looks like the db function always returns an empty array by default, so I'm not sure how to best handle this... Might require better handling in the upstream function to fully utilize properly.

harvester/utils/general_utils.py
```diff
@@ -240,23 +242,32 @@ def traverse_waf(
...
 
 
-def query_filter_builder(base, facets):
-    """Builds filter strings from base and comma separated string of filters
-    :param base str - base filter query
-    :param facets str - extra facets
+def query_filter_builder(model, facets_string):
+    """Builds a list of filter expressions from a comma-separated string of facets
 
+    Each facet is of them form "column op value" where `column` is a
```
jbrown-xentity: `of the form`

jbrown-xentity (commented):
Huge improvements, thanks @neilmb!
I know this will need another review after figuring out and fixing the e2e tests, but wanted to take a look and provide feedback. Some of the following are outstanding issues that are raised by this work, not due to the changes themselves...

app/routes.py
neilmb: These facets are for harvest jobs and we don't want to show "created" jobs from the future because that's how we do our scheduling.

app/routes.py
neilmb: Done, good suggestion.

harvester/utils/general_utils.py
neilmb: ope.

neilmb:
@jbrown-xentity I think I have the tests all fixed if you would re-review.

app/routes.py
neilmb: This replicates how it was working, so I'm not going to change it in this PR.

jbrown-xentity (approved):
LGTM!

PR #364, opened 2025-08-22
Use session-scoped interface for harvester

neilmb:
# Pull Request

Related to GSA/data.gov#5402, we have problems when the harvester tries to access the database using the Flask app's database connection. In `harvester/__init__.py` we set up an instance of `HarvesterDBInterface` with the correctly configured database session, but we were neglecting to use it in a couple of places, leading to errors that showed up in our logs.

This PR changes those two spots to use the `db_interface` from `harvester/__init__.py`.

## About

This is a relative import from a level above and it seems to be working fine, but relative imports like this can lead to circular imports in some cases, but I think that this usage is clean.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
I believe this should work... Do we have a test case replicating the failure, so that we can confirm this will fix our problem?

PR #365, opened 2025-08-22
Qa upgrades

jbrown-xentity:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5387
You can see the latest successful run from this branch [here](https://github.com/GSA/datagov-harvester/actions/runs/17166019487)

## About

Focused on 2 things:
- Add collection dataset information (as counts) from current prod catalog, for more direct comparison to catalog-beta
- Add crosswalk usability of names to titles, so that we are comparing apples to apples.

Couple of interesting notes:
- Catalog-beta has about 76% of catalog.
- Many basic QA reviews of some of the empty items show that the harvest source either no longer exists, or the job never really got going (was killed)
- Census is roughly half of all of catalog (48%)
- Census, NOAA, DOI, and NASA combine to make up 82% of catalog

## PR TASKS

- [x] Code well documented
- [ ] ~Tests written, run and passed~ (no tests for this as it won't be needed in the future, see https://github.com/GSA/datagov-harvester/actions/runs/17166019487 for latest run)
- [x] Files linted


scripts/lib/qa/harvest_sources.py
```diff
@@ -49,21 +51,32 @@ def get_harvest_sources(self):
                 self.sources = res.json()
         self.sources = {source["name"]: source for source in self.sources}
 
-    def keep_alphanumeric(self, text):
-        """
-        used to standardize harvest source names
-        """
-        return "".join(char for char in text if char.isalnum())
+    def map_names_and_titles(self, harvest_counts):
```
neilmb, line 54: **suggestion (non-blocking)**: This function is long enough to deserve a docstring. For example, what is `harvest_counts` and does it return new values or mutate its argument?

scripts/lib/qa/harvest_sources.py
```diff
@@ -49,21 +51,32 @@ def get_harvest_sources(self):
...
+        mapping = session.get(
+            (
+                f"{CATALOG_PROD_BASE_URL}/api/action/package_search"
+                f"?fq=dataset_type:harvest&fl=title,name&rows=1000"
+            )
+        ).json()["result"]["results"]
+        mapped = {}
+        for item in mapping:
```
neilmb: **suggestion (non-blocking)**: This for loop is a perfect dict-comprehension:

```
mapped = {item["title"]: item["name"] for item in mapping}
```

neilmb (dismissed):
LGTM, just some Python snark.

scripts/lib/qa/harvest_sources.py
jbrown-xentity: Absolutely, cleaned!

scripts/lib/qa/harvest_sources.py
jbrown-xentity: Good call, added and tested.

neilmb (approved):

PR #369, opened 2025-08-25
stay with python 3.12

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5352

## About

Python 3.13 failed to connect egress proxy server due to [more strict SSL validation](https://github.com/GSA/data.gov/issues/5352). It turns out all cloud.gov internal container-to-container HTTPS traffic are affected by this, and the only work-around is using python 3.12, according the the slack discussion in [cg-questions](https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1756131384894399). 

Comment was added where we pin python to `~3.12.0` in `pyproject.toml`. 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


neilmb (approved):
Glad you tracked this down, thanks. Maybe mention on the channel that we'll all need to rebuild our local docker compose and poetry setups after this?

PR #373, opened 2025-08-26
set harvest.data.gov in login.gov settings

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5344

## About

In login.gov settings we need to use the new domain

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM. I'm glad that's the only reference we need for this URL...

PR #377, opened 2025-08-27
handle when "@type" value is bad

rshewitt:
# Pull Request

Related to [#5377](https://github.com/GSA/data.gov/issues/5377)

## About
- handle situations when the "@type" value is bad

## PR TASKS

- [x] Code well documented
- [x] Files linted


neilmb (approved):
The new error message in the test looks better.

PR #388, opened 2025-08-29
handle decode exception

rshewitt:
# Pull Request
- [job](https://harvest.data.gov/harvest_job/4204b31b-be2f-4b73-a21e-892b9ef1ef5f) stopped early because of decoding error on [record](https://meta.geo.census.gov/data/existing/decennial/GEO/GPMB/TIGERline/TIGER2015/featnames/tl_2015_72131_featnames.dbf.iso.xml). 

## About
- removed raise of `ExternalRecordToClass` allows record processing to continue. i tried a couple of try/except wrappers in the parent/calling function but couldn't get anything to work. 
- add new decode error fixture (encoded it in Windows 1252) and update tests to handle it

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/example_data/iso_2_waf/decode_error.xml
+++ b/example_data/iso_2_waf/decode_error.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<root>
+    <data>Valid text</data>
+    <data>Here’s a bad character: –</data>
+</root>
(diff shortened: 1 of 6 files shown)


neilmb (changes requested):
The changes look reasonable, but I think we really want test cases to verify that these exceptions get raised when we want them to. Could you capture the files from these sources and use them in `tests/integration/harvest_job_flows/`?

neilmb (approved):

PR #391, opened 2025-09-02
Nmb/5411 record errors pagination

neilmb:
# Pull Request

Related to GSA/Data.gov#5411, this fixes the pagination for the record errors table on the harvest jobs page. When we introduced the `page_param` configuration option for our HTMX pagination, we missed setting that in this one case.

## About

I ended up testing this in Flask rather than Playwright because I found it very hard to add things programmatically in Playwright tests because we (very reasonably) don't have a database interface alongside the browser and we don't have everything piped through our API (record error creation in this case).

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


cmhedrickREI (approved):
LGTM! Looks like it just needs main merged in but looks good to go!

PR #392, opened 2025-09-02
Use unchanged instead of ignored in UI elements

neilmb:
# Pull Request

Related to GSA/Data.gov#5288, this changes the UI language for dataset records that haven't changed since the last harvest from "ignored" to "unchanged". This is the minimal intervention, I didn't change anything at the database level, I only changed the visible language in harvest emails, the job view page, and the harvest source chart. At this point, I think that this is a better mix of effort/reward than renaming fields in the database.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -778,7 +778,7 @@ def view_harvest_source(source_id: str):
                     "backgroundColor": "red",
                 },
                 {
-                    "label": "Ignored",
+                    "label": "Unchanged",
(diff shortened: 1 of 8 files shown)


jbrown-xentity (dismissed):
LGTM, all good improvements!

cmhedrickREI (approved):
lgtm

PR #394, opened 2025-09-02
Hide notification emails without login

neilmb:
# Pull Request

Related to GSA/Data.gov#5422, we decided not to show harvest source notification emails on the public pages. This makes that change and adds test cases for the presence of data on the logged-in versions of the pages.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM, nice change in tests.

PR #395, opened 2025-09-03
Record error type summary

neilmb:
# Pull Request

Related to GSA/Data.gov#5381, we wanted to add a summary of the different types of errors that are present on a job. This adds a database method to generate that summary using a SQL `group_by` and then uses that summary method to put a table on the harvest job page and test for its presence in Playwright and flask.

Screenshot 2025-09-03 at 12 14 01 PM

## About

I think that I translated my naive SQL idea for how to do this query into SQLAlchemy correctly, but I would be happy to learn differently if anyone knows better.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


database/interface.py
```diff
@@ -440,6 +440,20 @@ def get_harvest_record_errors_by_record(self, record_id: str):
...
+        query = (
+            self.db.query(
+                HarvestJob.id,
+                HarvestRecordError.type,
+                func.count(HarvestRecordError.id),
+            )
+            .filter(HarvestJob.id == job_id)
+            .join(HarvestRecordError)
+            .group_by(HarvestJob.id, HarvestRecordError.type)
+        )
```
rshewitt: a couple things:
- you don't need the harvest job id in the return set via `query(HarvestJob.id...)`
- you don't need a join since harvest record errors all have a job id
- you can just group by `HarvestRecordError.type`. 

here's what i got...
```sql

stmt = (
    self.db.query(
        func.count(),
        HarvestRecordError.type
    )
    .where(HarvestRecordError.harvest_job_id == "28ef082f-c7cd-4514-8eed-33425122361c")
    .group_by(HarvestRecordError.type)
)
```

neilmb: Thank you, that's much simpler.

tests/playwright/test_harvest_job.py
```diff
@@ -60,6 +60,16 @@ def test_harvest_job_record_errors_display(self, upage):
             "/harvest_record/0779c855-df20-49c8-9108-66359d82b77c",
         )
 
+    def test_harvest_job_record_errors_summary(self, upage):
```
rshewitt, line 63: the summary table can have more than 1 type. can that be included in a/this test?

tests/playwright/test_harvest_source.py
```diff
@@ -78,7 +78,6 @@ def test_job_summary(self, upage):
     @pytest.mark.parametrize(
         "data_term_name, glossary_term_name",
         [
-            ("notification emails", "Notification emails"),
```
rshewitt: this is for glossary stuff. why is this being removed?

neilmb: Because the test is broken. "Notification emails" does not appear on the page any more for not logged-in users.

tests/playwright/test_harvest_job.py
neilmb: Sure, I changed the fixtures to add two types of record errors.

rshewitt (approved):
LGTM

PR #396, opened 2025-09-03
Use url_for in redirects

neilmb:
# Pull Request

Related to GSA/Data.gov#5368 fixes our use of string interpolation in creating URL redirects, restoring the pattern that we use elsewhere in this file.

## About

I fixed the two occurrences I could find in this file, I hope this will fix the finding.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM

PR #398, opened 2025-09-04
update job reporter action counting

rshewitt:
# Pull Request

Related to [#5426](https://github.com/GSA/data.gov/issues/5426)

## About
- add `HarvestSource` method for updating reporter action counter per record and updates the current job with the latest counts. this update includes record errors now instead of just records added, deleted, and updated. 
- centralize all counting logic to harvest.py
- record total is immediate instead of reported at the end
- progress bar is fixed
- change is live on dev. test [source](https://datagov-harvest-dev.app.cloud.gov/harvest_job/eafdaa22-cb7e-4f56-bfc0-1a8b6f54302c).

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -202,6 +202,21 @@ class HarvestSource:
                 self.job_id,
             )
 
+    def update_job_record_count_by_action(self, action: str):
+        """updates the record action counter in the reporter
+        and updates the job with new count dict
+
+        record actions are:
+            create
+            update
+            delete
+            None (unchanged)
+            errored
+            validated
+        """
+        self.reporter.update(action)
+        self.db_interface.update_harvest_job(self.job_id, self.reporter.report())
+
     def get_source_orm(self) -> HarvestSourceORM:
         """Get the harvest source object from the database."""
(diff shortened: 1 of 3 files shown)


jbrown-xentity (dismissed):
LGTM.

harvester/harvest.py
```diff
@@ -202,6 +202,12 @@ def get_source_info_from_job_id(self, job_id: str) -> None:
                 self.job_id,
             )
 
+    def update_job_record_count_by_action(self, action: str):
+        """updates the record action counter in the reporter
+        and updates the job with new count dict"""
```
neilmb: **suggestion (non-blocking)**: Can you say something about the possible actions? I see you using `None` later and it surprised me, so maybe document that as a possible action and what it means?

neilmb (dismissed):
This looks reasonable. 

Is there a way to write a test for the harvester to make sure that reported counts are getting updated "live"? Can we somehow do part of the work and then assert what the value of the various record counters are at that point in time?

neilmb (approved):

PR #401, opened 2025-09-05
add progress bar to "in_progress" jobs

rshewitt:
# Pull Request

Related to [#5438](https://github.com/GSA/data.gov/issues/5438)

## About
- adds progress bar to "in_progress" job on metrics page and replaces progress table data on job page
- update are live on [dev](https://datagov-harvest-dev.app.cloud.gov/)
- there's a bug which causes a "not found" error when navigating to some jobs (test data). the in progress job being one of them. this prevented me from adding a test checking to see if the progress bar was visible and at 0. i'm gonna create a ticket to fix that but didn't want to include that in this PR. [ticket](https://github.com/GSA/data.gov/issues/5440)
- there's a pytest name interference with "test_metrics" which prevented it from being included so I renamed the file. 

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/app/templates/components/job-table/job_table.j2
+++ b/app/templates/components/job-table/job_table.j2
@@ -1,4 +1,4 @@
-{% macro job_table(jobs, caption, show_source=true) -%}
+{% macro job_table(jobs, caption, show_source=true, show_progress=False) -%}
 <div class="usa-table-container" tabindex="0">
     <table class="usa-table usa-table--striped usa-table--borderless">
         <caption>{{ caption }}</caption>
@@ -7,6 +7,9 @@
                 <th scope="col" title="Job ID">Job ID</th>
                 {% if show_source %}<th scope="col" title="Harvest Source">Source</th>{% endif %}
                 <th scope="col" title="Job Status">Status</th>
+                {% if show_progress %}
+                    <th scope="col" title="Job Progress">Progress</th>
+                {% endif %}
                 <th scope="col" title="Job Type">Type</th>
                 <th scope="col" title="Start Date">Started</th>
                 <th scope="col" title="Completion Date">Finished</th>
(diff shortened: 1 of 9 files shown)


app/templates/components/job-table/job_table.j2
```diff
@@ -7,6 +7,9 @@
                 <th scope="col" title="Job ID">Job ID</th>
                 {% if show_source %}<th scope="col" title="Harvest Source">Source</th>{% endif %}
                 <th scope="col" title="Job Status">Status</th>
+                {% if caption == "Jobs In Progress" %}
```
neilmb: **comment (non-blocking)**: This seems fragile to me. If we change the caption then this would break. Could we instead add a `show_progress` parameter to the Jinja macro and then include the table with `show_progress=True` on the metrics page?

neilmb (dismissed):
One suggestion for maybe another way to do the specialization of the job table component.

app/templates/components/job-table/job_table.j2
rshewitt: yeah sure good idea

neilmb (dismissed):

tests/playwright/test_harvest_job.py
```diff
@@ -23,6 +23,9 @@ def test_config_table_properties(self, upage):
 
         # Test static content that should always be present
         expect(table).to_contain_text("Harvest Source:")
+
+        expect(table).not_to_contain_text("Percent complete:")  # job is not in progress
+        expect(table).to_contain_text("Harvest Source:")
```
neilmb, line 26: **question (non-blocking)**: Did this line get duplicated from line 25?

rshewitt: yes! thanks for this

neilmb (approved):

PR #404, opened 2025-09-05
fix job status overwrite on report

rshewitt:
# Pull Request

Related to [#5425](https://github.com/GSA/data.gov/issues/5425)

## About
- ensures the job status (on error) isn't overwritten on report
- ended up changing a job fixture from "error" to "new" since it didn't make sense (records were processed so it's not a critical error). this led to an increase of 1 job in the new jobs table explaining the playwright update.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
+++ b/tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
@@ -6,7 +6,12 @@ from unittest.mock import MagicMock, patch
 
 from deepdiff import DeepDiff
 
-from harvester.harvest import check_for_more_work, harvest_job_starter
+from harvester.harvest import (
+    check_for_more_work,
+    harvest_job_starter,
+    HarvestSource,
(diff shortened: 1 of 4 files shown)


tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
```diff
@@ -187,6 +192,7 @@ def test_harvest_record_errors_reported(
         job_errors = [
             error for record in harvest_job.records for error in record.errors
         ]
+        #
```
neilmb: **comment (non-blocking)**: We don't usually keep empty things in our code.

neilmb (dismissed):

tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
rshewitt: removed

neilmb (approved):

PR #405, opened 2025-09-05
Use cloud.gov brokered ses

neilmb:
# Pull Request

Related to GSA/Data.gov#5396 we can now use a brokered email service from Cloud.gov. This changes our infrastructure-creation script to use that plan and modifies our environment variable settings to account for the change from `...ssb.data.gov` to `...appmail.cloud.gov`.

## About

This will take some care when merging. The current SMTP services from prod and staging either need to be deleted to let the new `aws-ses` one get created on deploy, or we need to make a new `aws-ses` service and rename it into place just after this is merged, then restart the applications.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/.profile
+++ b/.profile
@@ -34,7 +34,7 @@ export HARVEST_SMTP_SERVER=$(vcap_get_service smtp .credentials.smtp_server)
 export HARVEST_SMTP_STARTTLS=True
 export HARVEST_SMTP_USER=$(vcap_get_service smtp .credentials.smtp_user)
 export HARVEST_SMTP_PASSWORD=$(vcap_get_service smtp .credentials.smtp_password)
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
Since emails are only sent from the harvest jobs (aka cloud.gov task), there will be a weird period when implementing this. I think removing the SMTP service and letting the github action re-create it makes total sense. We should just do this when there aren't any jobs being run in production (just check the metrics page). If we release the code and change the service while a harvest job is running, it won't get the new configuration and the email will fail to send.
Not a huge deal in this use case, but worth trying to avoid any email failures.
This looks good to ship, in my opinion.

PR #406, opened 2025-09-05
Add XML mimetype

jbrown-xentity:
Add xml mimetype (and tests for verification) so that the content displays well in the browser.

# Pull Request

No formal ticket, just an annoyance that I wanted to correct.

## About

Simple change to support xml mimetypes, with tests.

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -1178,7 +1178,8 @@ def get_harvest_record_raw(record_id=None):
             source_raw_json = json.loads(record.source_raw)
             return jsonify(source_raw_json), 200
         except json.JSONDecodeError:
-            return record.source_raw, 200
+            # Return XML with correct mimetype
+            return Response(record.source_raw, mimetype="application/xml; charset=utf-8"), 200
     else:
         return JSON_NOT_FOUND()
 
--- a/app/routes.py
+++ b/app/routes.py
@@ -1171,12 +1171,10 @@ def get_harvest_record(record_id):
 @main.route("/harvest_record/<record_id>/raw", methods=["GET"])
 @valid_id_required
 def get_harvest_record_raw(record_id=None):
-    record = db.get_harvest_record(record_id)
-    if not record:
+    record, source = db.get_harvest_record_with_source(record_id)
+    if not record or not source:
         return JSON_NOT_FOUND()
 
-    # Get the harvest source to determine schema_type
-    source = db.get_harvest_source(record.harvest_source_id)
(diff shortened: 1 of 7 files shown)


app/routes.py
```diff
@@ -1178,7 +1178,8 @@ def get_harvest_record_raw(record_id=None):
             source_raw_json = json.loads(record.source_raw)
             return jsonify(source_raw_json), 200
         except json.JSONDecodeError:
-            return record.source_raw, 200
+            # Return XML with correct mimetype
+            return Response(record.source_raw, mimetype="application/xml; charset=utf-8"), 200
```
rshewitt: can't we be more precise by getting the source type of the harvest source instead of guessing by exception?

jbrown-xentity: Good call, I changed it up (and added a DB call to speed up performance, so it doesn't make 2 DB calls).
Can you re-review @rshewitt ?

app/routes.py
```diff
@@ -1171,16 +1171,31 @@ def get_harvest_record(record_id):
...
-    record = db.get_harvest_record(record_id)
-    if record:
+    record, source = db.get_harvest_record_with_source(record_id)
+    if not record or not source:
+        return JSON_NOT_FOUND()
+
+    schema_type = getattr(source, "schema_type", None)
+
+    # If schema_type contains 'dcatus', treat as JSON, else XML
+    if schema_type and "dcatus" in schema_type:
```
rshewitt, line 1182: schema_type is required. Are there situations it would be None?

database/interface.py
```diff
@@ -515,6 +515,20 @@ def delete_harvest_record(
     def get_harvest_record(self, record_id):
```
rshewitt, line 515: you can get the source from the record returned from this method. you shouldn't need `get_harvest_record_with_source`. testing locally i got...
```python
res = interface.get_harvest_record(db_records[0].id)
res.job.source
```

app/routes.py
jbrown-xentity: Even if I built the system, I always assume dirty data, and with a web service I'd rather give a sensical answer instead of a non-sensical one.

database/interface.py
jbrown-xentity: You are absolutely right, no need for the extra code. Removed. I now have a better understanding of how that lazy loading works, thanks!

rshewitt (approved):
looks good!

PR #409, opened 2025-09-08
Clickable Harvest Source URLs

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5435

## About
- Wraps the Harvest Source URL's in anchor tags and opens them in new tab when clicked
- Updates tests to check that the harvest source URL's have the proper attributes

## PR TASKS

- [x] Code well documented
- [x] Files linted


jbrown-xentity (approved):
LGTM

PR #413, opened 2025-09-09
Use Placeholder for `describedByType`

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5335

## About

- Uses the a placeholder (`application/octet-steam`) for non-valid `describedByTypes`

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -798,6 +798,11 @@ class Record:
                 "name": self.harvest_source.get_source_orm().org.name
             }
 
+        if not self.is_valid_describedByType(
+            self.transformed_data.get("describedByType", "")
+        ):
+            self.transformed_data["describedByType"] = "application/octet-steam"
+
         # If distribution items have a downloadURL or accessURL,
         # check if it just needs an "https://" at the beginning
         # to be valid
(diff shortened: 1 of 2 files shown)


cmhedrickREI:
Example of the difference between the validation methods to explain the use of looking back at the HarvestSource Object:
Screenshot from 2025-09-09 12-44-37

harvester/harvest.py
```diff
@@ -805,6 +813,8 @@ def _guess_better_url_in_item(item, key):
         for dist_item in self.transformed_data.get("distribution", []):
             _guess_better_url_in_item(dist_item, "downloadURL")
             _guess_better_url_in_item(dist_item, "accessURL")
+            if not self.is_valid_describedByType(dist_item.get("describedByType", "")):
```
rshewitt, line 825: `describedByType` can also occur at the root of dcatus not just in distributions. can you add that as well?

cmhedrickREI: @rshewitt I just added some code to check the root now as well and replace it. Is it safe to assume the schema pattern for the root `decribedByType` is the same as the one in the `distributions`?

rshewitt: yeah they're the same

rshewitt (approved):
helps by ensuring records aren't being rejected because of invalid "describedByType"

PR #414, opened 2025-09-09
use record identifier in exception message

rshewitt:
# Pull Request

Related to [#5432](https://github.com/GSA/data.gov/issues/5432)

## About
- uses harvest source identifier (full address path of xml file when waf or "identifier" in dcatus) in `ExternalRecordToClass` exception message instead of harvest source url

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
+++ b/tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
@@ -161,6 +161,38 @@ class TestHarvestJobFullFlow:
         ## assert call_args to package_create
         ## TODO this test wil eventually succeed. we can then assert call_args
 
+    def test_harvest_waf_iso19115_2_download_exception(
+        self,
+        interface,
+        organization_data,
+        source_data_waf_iso19115_2,
+    ):
+        interface.add_organization(organization_data)
+        interface.add_harvest_source(source_data_waf_iso19115_2)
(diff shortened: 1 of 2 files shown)


tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
```diff
@@ -161,6 +161,38 @@ def test_harvest_waf_iso19115_2(
...
+
+        job_id = harvest_job.id
+
+        harvest_source = HarvestSource(job_id, "harvest")
+        harvest_source.acquire_minimum_external_data()
+        harvest_source.external_records[0]["identifier"] = (
+            "https://localhost:1234/test.xml"
+        )
+
+        list(harvest_source.external_records_to_process())
```
neilmb, line 187: **suggestion (non-blocking)**: You aren't assigning the result of this call to anything, do you need this at all?

neilmb (approved):
LGTM, a mere quibble.

tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
rshewitt: `harvest_source.external_records_to_process()` returns a generator. i just need the loop to occur so the bad url is "downloaded" and the intended exception is thrown.

PR #417, opened 2025-09-12
Add support for single ISO document harvest sources

neilmb:
# Pull Request

Related to GSA/data.gov#5182, we want to support harvest sources that have a single document that is an ISO-format XML file rather than a DCAT-US JSON file. This adds support for that situation by entering that single file into the harvesting process as if we had observed a WAF with a single file.

## About

This was intensely, incredibly easy to add. @rshewitt is a legend.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -345,7 +345,13 @@ class HarvestSource:
                     dataset = record["content"]
 
                 if self.source_type == "document":
-                    dataset = json.dumps(sort_dataset(record))
+                    if self.schema_type.startswith("dcatus"):
+                        dataset = json.dumps(sort_dataset(record))
+                    elif self.schema_type.startswith("iso19115"):
+                        # single document ISO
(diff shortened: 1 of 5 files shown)


harvester/harvest.py
```diff
@@ -376,7 +382,13 @@ def acquire_minimum_external_data(self) -> list:
         logger.info("retrieving external records.")
         try:
             if self.source_type == "document":
-                self.external_records = download_file(self.url, ".json")["dataset"]
+                if self.schema_type.startswith("dcatus"):
+                    self.external_records = download_file(self.url, ".json")["dataset"]
+                elif self.schema_type.startswith("iso19115"):
+                    # mimic the output of traverse_waf with a single file
```
jbrown-xentity, line 388: I love comments like this, that so succinctly describe why it's doing what it's doing...

jbrown-xentity (approved):
LGTM!

PR #421, opened 2025-09-15
remove csdgm from schema type enum

rshewitt:
# Pull Request

Related to [#5088](https://github.com/GSA/data.gov/issues/5088)

## About
- removes "csdgm" from schema_type enum in harvest source model
- refactored `test_harvest_source_bad_schema_type` to check that the org has no harvest sources because it was rejected.
- removed an old test involving a "csdgm" source which we no longer support

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/tests/integration/harvest/test_exception_handling.py
+++ b/tests/integration/harvest/test_exception_handling.py
@@ -45,26 +45,6 @@ class TestHarvestJobExceptionHandling:
         harvest_error = interface.get_harvest_job_errors_by_job(harvest_job.id)[0]
         assert harvest_error.type == "ExtractExternalException"
 
-    def test_harvest_source_bad_schema_type(self, interface,
-                                            organization_data,
-                                            source_data_dcatus,
-                                            job_data_dcatus):
-        """Schema types that don't start with dcatus or iso19115 raise an exception."""
-        interface.add_organization(organization_data)
-        source_data_dcatus["schema_type"] = "csdgm"
-        interface.add_harvest_source(source_data_dcatus)
-        harvest_job = interface.add_harvest_job(job_data_dcatus)
-
-        harvest_source = HarvestSource(harvest_job.id)
-
-        with pytest.raises(ExtractExternalException) as e:
-            harvest_source.acquire_minimum_external_data()
-
-        assert harvest_job.status == "error"
-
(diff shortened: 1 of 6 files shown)


tests/integration/harvest/test_exception_handling.py
```diff
@@ -45,25 +45,16 @@ def test_bad_harvest_source_url_exception(
...
-
-        with pytest.raises(ExtractExternalException) as e:
-            harvest_source.acquire_minimum_external_data()
-
-        assert harvest_job.status == "error"
 
-        harvest_error = interface.get_harvest_job_errors_by_job(harvest_job.id)[0]
-        assert harvest_error.type == "ExtractExternalException"
+        # unable to add a "csdgm" harvest source because we don't support it
+        assert len(interface.get_harvest_source_by_org(org_id)) == 0
```
neilmb: **comment (non-blocking)**: You can also check the return value of the `add_harvest_source` call which should be `None` if the add failed.

tests/integration/harvest/test_extract.py
```diff
@@ -8,22 +8,6 @@ def test_traverse_waf_ms_iis(self, mock_requests_get_ms_iis_waf):
         files = traverse_waf(url="https://example.com")
         assert len(files) == 2
 
-    def test_extract_waf(
-        self,
-        interface,
-        organization_data,
-        source_data_waf_csdgm,
```
neilmb, line 15: **question (non-blocking)**: If we're getting rid of CSDGM here, should we also get rid of the unused source and job data fixtures?

neilmb (dismissed):
Does what it says on the tin, just a few minor queries.

tests/integration/harvest/test_exception_handling.py
rshewitt: i ended up scrapping this test in favor of "test_extract_unsupported_schema" in our db tests. they do the same thing.

tests/integration/harvest/test_extract.py
rshewitt: i removed the csdgm job fixture but kept "source_data_waf_csdgm" because a db test uses it

neilmb (approved):

PR #424, opened 2025-09-16
Add a waf-collection harvest source type

neilmb:
# Pull Request

Related to GSA/Data.gov#5445, we want to support `waf-collection` harvest sources and this is the first step. I alters our `HarvestSource` model to permit a `source_type` of `waf-collection` and adds a `collection_parent_url` field for use in that case (along with a check constraint that we don't use that field for any other source type)

## About

Migrations were tricky here with modifying the Enums in Postgres directly and the check constraint had to be added in a separate migration from the field being added due to a postgres restriction. I also got some guff from the linter that introduced whitespace changes in some unrelated files.

## PR TASKS

- [X] Code well documented
- [X] Files linted


jbrown-xentity (dismissed):
LGTM. I don't know what our harvest job/task would actually do with these types currently, but it's a start for setting up how to declare and use waf-collections.

rshewitt (changes requested):
blocking until enum lock fixed

rshewitt:
terminating db connections before running a highly restrictive migration (table lock) appeared to [work](https://github.com/GSA/datagov-harvester/actions/runs/17919643883). i added a utility [func](https://github.com/GSA/datagov-harvester/blob/e28719d4ff2caefcfdb96c7f50e305c6aac34491/migrations/utils.py#L1) to help out with this.

app/templates/view_source_data.html
```diff
@@ -36,6 +36,11 @@ <h2>Configuration:</h2>
                 <td>
                     <a href="{{value}}" target="_blank">{{value}}</a>
                 </td>
+                {% elif key == 'collection_parent_url' %}
```
rshewitt, line 39: probably should add 'collection_parent_url' to the glossary. that can be another ticket if you want to create that.

migrations/versions/3b56c7e46974_add_collection_parent_url_and_waf_.py
```diff
@@ -0,0 +1,52 @@
+"""Add collection_parent_url and waf-collection type
+
+Revision ID: 3b56c7e46974
+Revises: 1800d355e5b9
```
rshewitt: update to "a6aa1afd27b7"

migrations/versions/5416f9f5f06d_add_constraint_for_waf_collections.py
```diff
@@ -0,0 +1,29 @@
+"""Add constraint for waf-collections
+
+Revision ID: 5416f9f5f06d
+Revises: d805537d4cf2
```
rshewitt: update to "3b56c7e46974"

migrations/versions/5416f9f5f06d_add_constraint_for_waf_collections.py
```diff
@@ -0,0 +1,29 @@
...
+# revision identifiers, used by Alembic.
+revision = "5416f9f5f06d"
+down_revision = "3b56c7e46974"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    with op.batch_alter_table("harvest_source", schema=None) as batch_op:
+        batch_op.create_check_constraint(
```
rshewitt, line 25: what lock does adding a check constraint require?

rshewitt: you might be able to merge this into "3b56c7e46974"

rshewitt (approved):
looks good! terminating processes just on the first migration should help the next

PR #425, opened 2025-09-16
add notification frequency option `on_error_or_update`

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5348

## About

- Adds `on_error_or_update` option to `notification_frequency`
- Adjusts logic to send off email if there's errors or updates
- Adds migration to update the db (Note: I had to use the `revision` option because the change wasn't picked up which seems to sometimes be the case with ENUM vars. Feed back on the migration file and thoughts on alternatives is appreciated)

## PR TASKS

- [x] Code well documented
- [x] Files linted


jbrown-xentity (commented):
LGTM, but those UI errors should be investigated more fully... Looks like there's a 500 error on a lot of pages. My guess is something is breaking the harvest source details page.

rshewitt (dismissed):
blocking until enum lock fixed

rshewitt:
terminating db connections before running a highly restrictive migration (table lock) appeared to [work](https://github.com/GSA/datagov-harvester/actions/runs/17919643883). i added a utility [func](https://github.com/GSA/datagov-harvester/blob/e28719d4ff2caefcfdb96c7f50e305c6aac34491/migrations/utils.py#L1) to help out with this.

neilmb:
@cmhedrickREI Here's the changes I incorporated with Reid's helper for my enum changes: https://github.com/GSA/datagov-harvester/commit/4513ee373346599fbd13ea2b3fb45f60581b6a4d

migrations/versions/e75e287e2251_add_error_and_update_notification_type.py
```diff
@@ -0,0 +1,61 @@
...
+down_revision = "5416f9f5f06d"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    # enum change uses a restrictive table lock, so get everything out of the
+    # way so we can succeed
+    op.execute(get_terminate_processes_sql_cmd())
```
rshewitt: i don't think adding to an enum requires the same level of restriction as changing dtypes (e.g. swapping enums) so you might not need to terminate connections here. i could be wrong but we only want to do this if we have to. you can test this on dev.

cmhedrickREI: I'll give it a shot

tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
```diff
@@ -264,6 +264,48 @@ def test_harvest_record_errors_reported(
         # assert that send_notification_emails is called because of errors
         assert send_notification_emails_mock.called
 
+    @patch("harvester.harvest.ckan_sync_tool.ckan")
+    @patch("harvester.harvest.download_file")
+    @patch("harvester.harvest.HarvestSource.send_notification_emails")
+    def test_harvest_record_updates_reported(
```
rshewitt, line 270: it might be better to rework this test so that on both occasions (error or update) `send_notification_emails_mock` is called. Something like...
```python
# within the test
harvest_source = HarvestSource(job_id, "harvest")
harvest_source.reporter.update("update")
harvest_source.report()
# assert send_notification_emails.called once

harvest_source.reporter.updated = 0 # reset 
harvest_source.reporter.update("errored")
harvest_source.report()
# assert send_notification_emails.called twice
```
maybe it's not as simple as that but the idea is there.

neilmb: I don't know about how to make it do this, but I agree with Reid, I don't think that this test shows that emails are sent for updates as well as errors.

neilmb (commented):
LGTM, but I'd love to see a test that shows emails are sent when updates but not errors occurred.

tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
cmhedrickREI: I think I was able to establish a way of doing this with a patch!

neilmb (approved):
That looks great, thank you.

cmhedrickREI:
Screenshot from 2025-09-26 15-01-06

PR #426, opened 2025-09-16
set statement and idle connection timeouts

rshewitt:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5453#issuecomment-3299835248

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


harvester/__init__.py
```diff
@@ -16,8 +16,15 @@
 
 DATABASE_URI = os.getenv("DATABASE_URI")
 
+timeout = os.getenv("DB_TIMEOUT", 300000)  # 5 minutes
```
jbrown-xentity, line 19: Is it conceivable that a delete of a harvest source could take longer than 5 minutes? Not sure, but wanted to raise...

jbrown-xentity (approved):
LGTM, one "food for thought"...

harvester/__init__.py
rshewitt: the statement timeout should grant each harvest record 5 minutes to be deleted from the harvester db. as i understand it, postgres statements end on commit, rollback, or session close. we commit or rollback on each harvest record deletion.

PR #427, opened 2025-09-17
new enum approach

rshewitt:
# Pull Request

current csdgm migration breaks deployment. swapping for [approach](https://github.com/GSA/datagov-harvester/actions/runs/17770343054) which was able to be deployed on dev. 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


neilmb (approved):
LGTm, but since this was constructed by an alembic plugin can we include that in our poetry setup too?

rshewitt:
> LGTm, but since this was constructed by an alembic plugin can we include that in our poetry setup too?

i'd like to avoid adding another dependency if we can. we'll probably only ever remove/add a value from a enum or remove/add an enum entirely. we can get the approach by installing the plugin locally and generating the migration script.

neilmb:
> > LGTm, but since this was constructed by an alembic plugin can we include that in our poetry setup too?
> 
> i'd like to avoid adding another dependency if we can. we'll probably only ever remove/add a value from a enum or remove/add an enum entirely. we can get the approach by installing the plugin locally and generating the migration script.

I don't think that this is consistent with our approach to security. There is no benefit from pretending to have fewer dependencies. If we are using a dependency, then we are accepting the risk of using it, no matter how secretive we are about it. If we don't want to install the package as part of our production build and deploy process, then we have the `dev` dependencies in `pyproject.toml` that don't get exported to `requirements.txt` and so they don't get included when we build for Cloud.gov.

PR #428, opened 2025-09-17
Click-to-define only on Harvest Source Config Table Labels

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5447

## About

On the Harvest Source Detail page this makes it so we only ever select the Label that's on the table which is the first `td` of the `tr`

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


cmhedrickREI:
Screenshot showing we only have the `frequency` label set as a click-to-define, and not the "frequency" in the harvest name section.
Screenshot from 2025-09-17 12-44-21

neilmb (approved):
Looks reasonable. A playwright test might be possible, but probably not necessary.

PR #430, opened 2025-09-18
Enforce MFA w/ PIV

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5410

## About

- Enforces PIV Card MFA for login

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/app/routes.py
+++ b/app/routes.py
@@ -191,7 +191,7 @@ def login():
 
     auth_request_url = (
         f"{AUTH_URL}?response_type=code"
-        f"&acr_values=urn:acr.login.gov:auth-only http://idmanagement.gov/ns/assurance/aal/2?hspd12=true"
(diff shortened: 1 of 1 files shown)


app/routes.py
```diff
@@ -190,12 +191,13 @@ def login():
 
     auth_request_url = (
         f"{AUTH_URL}?response_type=code"
+        f"&acr_values=urn:acr.login.gov:auth-only http://idmanagement.gov/ns/assurance/aal/2?hspd12=true"
```
neilmb: **suggestion (blocking)**: The Login.gov docs say "Multiple values can be joined with a space (before being URI-escaped in the final URL)." and I don't see any obvious evidence that this URL is going to get escaped before we use it, so I think that it's safest to concatenate these with a `+` as we do with the `scope` parameter.

neilmb (changes requested):
Very good, but one thing I'd like to see addressed.

neilmb (approved):
Thanks!

PR #431, opened 2025-09-19
remove csdgm enum from schema type

rshewitt:
# Pull Request

Related to [#5088](https://github.com/GSA/data.gov/issues/5088)

## About
- removes "csdgm" from schema_type enum in harvest source model
- refactored `test_harvest_source_bad_schema_type` to check that the org has no harvest sources because it was rejected.
- removed an old test involving a "csdgm" source which we no longer support

## PR TASKS

- [x] Code well documented
- [ ] Files linted

### DIFF
--- a/app/forms.py
+++ b/app/forms.py
@@ -45,7 +45,6 @@ class HarvestSourceForm(FlaskForm):
             "dcatus1.1: non-federal",
             "iso19115_1",
             "iso19115_2",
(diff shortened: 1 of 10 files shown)


jbrown-xentity (approved):
LGTM, but I'll let you deploy as you're ready @rshewitt . Thanks for the improved documentation!

PR #434, opened 2025-09-23
Handle weird validation

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5350

## About

- Handles the duplication of the `formats` listed for the schema field in the Validation Error

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


tests/conftest.py
```diff
@@ -347,6 +347,21 @@ def source_data_dcatus_long_description(organization_data: dict) -> dict:
     }
 
 
+@pytest.fixture
+def source_data_dcatus_none_value(organization_data: dict) -> dict:
+    return {
+        "id": "2f2652de-91df-4c63-8b53-bfced20b276c",
+        "name": "Long Description Test Source",
```
neilmb: **suggestion (non-blocking)**: Is this still the "Long Description test source" or is it now the "None test source"?

tests/integration/harvest/test_validate.py
```diff
@@ -200,6 +200,41 @@ def test_invalid_description_too_long_dcatus_federal(
...
+        errors = [
+            e[0] for e in interface.get_harvest_record_errors_by_job(harvest_job.id)
+        ]
+
+        # besides the broken uri issue, the bug we're testing for is
+        # modified is Null
+        assert len(errors) == 3
+
+        expected_error_message = "$.modified, 'string' does not match any of the acceptable formats: 'string'"
+        assert expected_error_message in errors[2].message
```
neilmb: **question (non-blocking)**: I honestly don't know, is the order of the errors stable, so the third one will always be the one we want?

neilmb (dismissed):
This code looks fine, but I don't think that I completely understand the initial bug or where it was arising. Probably want @rshewitt to review since he best understands the error message code.

tests/integration/harvest/test_validate.py
cmhedrickREI: I ended up fixing the file to get rid of the other 2 errors so that we we only have the null issue.

tests/integration/harvest/test_validate.py
```diff
@@ -200,6 +200,39 @@ def test_invalid_description_too_long_dcatus_federal(
...
+        test_record: Record = next(external_records_to_process)
+        assert not test_record.validate()
+        errors = [
+            e[0] for e in interface.get_harvest_record_errors_by_job(harvest_job.id)
+        ]
+
+        # the bug we're testing for is modified is Null
+        assert len(errors) == 1
+
+        expected_error_message = "$.modified, 'string' does not match any of the acceptable formats: 'string'"
```
rshewitt: looks like the repetition has been removed which is good but the value ('string') following the field ($.modified) isn't correct. that needs to be "None". the output validation message follows the format of "[field], [invalid_value] does not match any of the acceptable formats: [format1], [format2], [format3], etc..."

cmhedrickREI: I managed to find the cause of this and was able to push the update. The validation message now is: `<ValidationError: \"$.modified, None does not match any of the acceptable formats: 'string'\">"`

neilmb (approved):
This looks great. Ship it.

PR #435, opened 2025-09-23
Harvest waf-collections

neilmb:
# Pull Request

Related to GSA/data.gov#5444, we want to support the "waf-collection" harvest source  in H2.0 like we did when we invented it for the current catalog. A "waf-collection" is a waf source with an additional URL for a "parent" dataset that every dataset inside the WAF should have as its "parent".

## About

The implementation here adds an additional "parent_identifier" field to our HarvestRecord object to persist this connection into the database. When we "transform" the XML metadata, we explicitly add this parent_identifier in as the DCAT property `isPartOf` which then gets synchronized to CKAN and dealt with there.

I copied over a standalone XML file in `example_data/valid_iso2_parent.xml` to serve as the parent dataset for the integration tests. The tests are minimal right now, just harvesting a single collection and making sure the parent information is correctly recorded and synced in `package_create`. If there are specific situations that people want to see in some re-harvesting tests, I'm happy to add them.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


database/models.py
```diff
@@ -177,6 +177,8 @@ class HarvestRecord(db.Model):
     action = db.Column(
         Enum("create", "update", "delete", name="record_action"), index=True
     )
+    # Parent information isn't in source_raw for XML records
+    parent_identifier = db.Column(db.String)
```
rshewitt, line 182: ~~i thought parent/child data was only for ckan? (excluding harvest source)~~. isn't this information available at the records harvest source?

harvester/harvest.py
```diff
@@ -301,6 +303,43 @@ def filter_waf_files_by_datetime(self) -> None:
...
+        # if the parent is new, not in this source's internal records, then we have to
+        # harvest everything
+        if self.external_records[0]["identifier"] not in self.internal_records:
+            # leave self.external_records unchanged
+            return
+
+        filtered_records = []
+        # the first record will be the parent, we don't have a modified_date yet
+        # so always harvest it
+        filtered_records.append(self.external_records[0])
```
rshewitt: not sure i understand this. the parent can fail sync (on first harvest) but its children pass. also, the parent can be filtered out if its modified date at the source is older than our copy. seems like you can just use `filter_waf_files_by_datetime` for "waf" and "waf-collection"

harvester/harvest.py
```diff
@@ -340,11 +379,17 @@ def external_records_to_process(self) -> any:
...
+                parent_identifier = None
                 if self.source_type == "waf":
                     record["content"] = download_file(record["identifier"], ".xml")
                     dataset = record["content"]
 
-                if self.source_type == "document":
+                elif self.source_type == "waf-collection":
+                    record["content"] = download_file(record["identifier"], ".xml")
+                    dataset = record["content"]
+                    parent_identifier = self.collection_parent_url
```
rshewitt, line 354: seems like you can download and set `dataset` in the same condition as "waf". just set `parent_identifier` in a new condition.

harvester/harvest.py
neilmb: We did not get the parent record by traversing a WAF so we don't have its datetime. It may not even be present inside a WAF so we shouldn't expect to be able to get a datetime for it without downloading the file.

rshewitt: you could use the default datetime [placeholder](https://github.com/GSA/datagov-harvester/blob/a6ef7a00fb827242f4fbe8baa9baecce14ca0ea0/harvester/utils/general_utils.py#L170) used in waf traversal when you add the parent in external_records in `acquire_minimum_external_data`

database/models.py
neilmb: I don't think that we want to indirect through the source because that parent URL could have been changed in that other object. I think it's simpler to store it here on the record so we can be sure we are syncing the right thing to CKAN no matter what has changed in the meantime in another table.

rshewitt: `parent_identifier` for the record is set by the [harvest source](https://github.com/GSA/datagov-harvester/blob/a6ef7a00fb827242f4fbe8baa9baecce14ca0ea0/harvester/harvest.py#L89C14-L89C35) in the runner

harvester/harvest.py
neilmb: Done.

database/models.py
neilmb: That's where it gets set when the harvest happens, but when we save that record to the database, I think that we want it to stay the same rather than referring out from that HarvestRecord to the current value in the database's HarvestSource.

harvester/harvest.py
neilmb: I think that this is clearer than that would end up being.

migrations/versions/416319f5e5bb_add_parent_identifier_to_record.py
```diff
@@ -0,0 +1,32 @@
...
+
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("harvest_record", schema=None) as batch_op:
+        batch_op.drop_column("parent_identifier")
+
+    # ### end Alembic commands ###
```
rshewitt, line 32: may want to terminate db connections at the top of each. maybe we should just turn that into a decorator we wrap around upgrade & downgrade

neilmb: Suspect that it's not an exclusive lock to add and drop columns.

rshewitt (dismissed):
looks good. paired with @neilmb and talked through things. approving but looking out for proof of successful deployment on dev

neilmb:
Dev deploy succeeded running migrations: https://github.com/GSA/datagov-harvester/actions/runs/18046850746/job/51359630079

@rshewitt Can you give a re-approval after some merge conflict fixing?

rshewitt (approved):
looks good

PR #436, opened 2025-09-23
Update cancel changes, and linting

jbrown-xentity:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/4897

## About

Tries to clean up the process, make more clear what's happening for what purpose. It would change the job information regardless of the job state, which seemed dangerous and wrong (even though it's only us right now).

Most of the changes are linting, with a few others sprinkled in.

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/harvester/lib/load_manager.py
+++ b/harvester/lib/load_manager.py
@@ -191,20 +191,35 @@ class LoadManager:
         if len(job_task) == 0:
             return f"No task with job_id: {job_id}"
 
-        if job_task[0][1] != "RUNNING":
-            updated_job = interface.update_harvest_job(
-                job_id, {"status": "complete", "date_finished": get_datetime()}
-            )
-            return (
-                f"Task for job {updated_job.id} is not running. Job marked as complete."
-            )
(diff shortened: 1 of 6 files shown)


neilmb (commented):
I like the look of this, but I'm concerned that we don't have a test case.  There's a `test_trigger_cancel_job` in `tests/app/test_load_manager.py` and "full flow" tests in `tests/integration/harvest_job_flows/`. Can we add a `tests/integration/harvest_job_flows/test_harvest_job_cancel.py` somehow? (I can imagine this might be difficult or impossible.)

jbrown-xentity:
> test_harvest_job_cancel

Good call. I updated the normal flow test case to evaluate my changes in the typical workflow (job in progess, cancels task, marks job as error, adds error message of manual cancel).
I created a new test that evaluates the "bad state" flow, where a message is returned and the job status is evaluated. This should keep us safe from "reclicks" on cancel job, or multiple people cancelling at the same time.
Can someone re-review @GSA/data-gov-dev-team ?

neilmb (approved):
Great tests, thanks.

PR #439, opened 2025-09-24
add links to catalog

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5460

## About

- add a new environment variable `ckan_url`, next to `ckan_api_url`
- provide links to `ckan_url` on org and source page
- add font-awesome for the external link icons

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


FuhuXia:
deployed to development space, screenshot captured.

image

app/routes.py
```diff
@@ -788,6 +789,7 @@ def view_harvest_source(source_id: str):
         }
         source = db.get_harvest_source(source_id)
         data = {
+            "ckan_url": os.getenv("CKAN_URL") or "",
```
neilmb: **suggestion (non-blocking)**: Because this is environment information and won't change between invocations, rather than repeating these `getenv` commands, add them at the top of the file near the existing "# Login authentication` calls to `getenv`.

README.md
```diff
@@ -94,8 +94,9 @@ To setup:
     a. On Mac's this can be derived by running `ifconfig | grep inet` and grabbing the first `192.168.xxx.xxx` address you find.
   7. Add two new values to the local `.env` file:
     ```
+      CKAN_URL=`ip-address-from-step-4a`:5000
```
neilmb, line 97: **suggestion (non-blocking)**: I'd prefer the more general name `CATALOG_URL`. If we aren't using CKAN, then we don't want to have to change this name.

neilmb (dismissed):

app/routes.py
FuhuXia: Will do.

README.md
FuhuXia: `CKAN_URL` is named side by side with `CKAN_API_URL`, too much hassle to change to `CKAN_API_URL` now since it together with `CKAN_API_TOKEN` is in many places including the github repo secrets names.

We can change everything together when time comes.

neilmb (approved):

PR #440, opened 2025-09-25
Use word-break for wrapping long identifiers

neilmb:
# Pull Request

Related to GSA/Data.gov#5353, this adds the `word-break: break-all` CSS property to the identifier in our record error display so that long identifiers don't extend off of the screen.

## About

The ticket suggests `overflow-wrap: break-word`, but that could break the line before the identifier if that would avoid breaking the word. I don't think that's what we want, so I used `word-break: break-all` instead. 

I couldn't find a way in Playwright to test the visual display of wrapped content on a particular size of screen, so I offer this screenshot from my local as evidence that this works as desired.

Screenshot 2025-09-25 at 12 45 56 PM

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/app/templates/view_job_data.html
+++ b/app/templates/view_job_data.html
@@ -130,7 +130,7 @@
                     {% for record_error in data.record_errors %}
                     <div class="error-block">
                         <h3>{{ record_error.error.id }}</h3>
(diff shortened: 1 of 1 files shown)


jbrown-xentity (approved):
LGTM!

PR #443, opened 2025-09-29
aggregate record error validation messages for job view by record

rshewitt:
# Pull Request

Related to [#5334](https://github.com/GSA/data.gov/issues/5334)

## About
- aggregates validation messages per record in the job view

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/app/templates/view_job_data.html
+++ b/app/templates/view_job_data.html
@@ -143,7 +143,19 @@
                             N/A
                             {% endif %}
                         </p>
-                        <p><strong>Error Message:</strong> {{ record_error.error.message }}</p>
+                        <!-- mixup between deployed app and testing env. TODO: fix this. -->
+                        {% if record_error.error.type in ["ValidationError", "ValidationException"] %}
+                            <p>
+                                <strong>Error Message:</strong> 
+                                <ul>
+                                    {% for msg in record_error.error.message %}
+                                        <li>{{msg}}</li>
+                                    {% endfor %}
+                                </ul>
+                            </p>
+                        {% else %}
+                            <p><strong>Error Message:</strong> {{ record_error.error.message }}</p>
+                        {% endif %} 
                         <p><strong>Type:</strong> {{ record_error.error.type }}</p>
                         <p class="utc-date" data-utc-date="{{record_error.error.date_created | utc_isoformat }}">
                             <strong>Date Created:</strong> {{ record_error.error.date_created | else_na }}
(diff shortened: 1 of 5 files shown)


neilmb:
Do we have a screenshot of what this will look like? I'm having trouble finding a job with multiple errors for a single record on dev.

app/templates/view_job_data.html
```diff
@@ -143,7 +143,19 @@ <h3>{{ record_error.error.id }}</h3>
...
+                            <p>
+                                <strong>Error Message:</strong> 
+                                <ul>
+                                    {% for msg in record_error.error.message %}
+                                        <li>{{msg}}</li>
+                                    {% endfor %}
+                                </ul>
+                            </p>
+                        {% else %}
+                            <p><strong>Error Message:</strong> {{ record_error.error.message }}</p>
```
neilmb, line 157: **question (non-blocking)**: It looks like the `message` element here will be a list of one item (from the `split` in `make_new_record_error_contract`). Are the brackets from the representation of the list going to appear in every error message?

database/interface.py
```diff
@@ -396,6 +396,37 @@ def get_harvest_job_errors_by_job(self, job_id: str) -> list[dict]:
...
+        # aggregate the validation messages by harvest_record_id using 1st row for
+        # all other fields. sql indexing starts at 1
+        instance_idx = 1
+        agg = (
+            self.db.query(
+                func.array_agg(query.c.harvest_record_id)[instance_idx],
+                func.array_agg(query.c.harvest_job_id)[instance_idx],
+                func.array_agg(query.c.date_created)[instance_idx],
+                func.array_agg(query.c.type)[instance_idx],
+                func.array_to_string(func.array_agg(query.c.message), ","),
```
neilmb: **question (non-blocking)** If the text of our error messages contain commas, when we split on this, won't we get that message split in half? Maybe use a less likely character for serializing this list, like `\0` or `\t`?

neilmb (dismissed):
I'd love to see how this looks on dev, but the code looks reasonable. A couple questions about commas and lists.

app/templates/view_job_data.html
rshewitt: good catch. this has been fixed.

database/interface.py
rshewitt: how about `::`? postgres doesn't support using null characters (`\0`) and I would say tabs are more common than my proposal

rshewitt:
example on [dev](https://datagov-harvest-dev.app.cloud.gov/harvest_job/95e471e1-ff77-4e53-ace7-824308c30b3d) @neilmb

neilmb (approved):

PR #444, opened 2025-09-29
remove length response body from error message

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5290

## About

CKANAPIError contains the whole response text, which is too much
https://github.com/ckan/ckanapi/blob/ckanapi-4.8/ckanapi/common.py#L134
let us throw away response from [url, status, response]

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


neilmb (approved):
LGTM

PR #446, opened 2025-09-30
Add the Dataset table

neilmb:
# Pull Request

Related to GSA/Data.gov#5487, this defines the data schema that we will use for our Dataset objects. One future possibility is to replace this table with a `VIEW` in GSA/Data.gov#5491, but I think it's worth specifying the structure of this table now.

## About

This is just the most basic data model, it isn't populated with any data ever right now.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


rshewitt:
everything makes sense. do we want to merge this into main or dev?

neilmb:
> everything makes sense. do we want to merge this into main or dev?

I guess we really need it on `develop` so that it will land in the `development` space. Let me edit the target of the PR. We'll have to merge a bunch of this back to `main` when we're ready to productionalize it.

rshewitt (approved):
looks good

PR #447, opened 2025-10-01
add dataset materialized view

rshewitt:
# Pull Request

Related to [#5491](https://github.com/GSA/data.gov/issues/5491)

## About
- creates the "dataset" materialized view using all fields from `harvest_record` 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (commented):
LGTM as a starting point.
I would consider creating some of the tsvector fields, but that can remain for another ticket.
Will the models.py version create the materialized view by default? Or will that only be run on the migration? We should probably standardize on a single location so updates don't have to occur in multiple locations...

rshewitt:
> e models.py version create the materialized view by default?

the migration should handle the actual creation but what's in models.py is the source of truth. i switched the destination branch back to "main". i pushed this commit on dev to make sure it worked

database/models.py
```diff
@@ -217,3 +218,16 @@ class Locations(db.Model):
     display_name = db.Column(db.String)
     the_geom = db.Column(Geometry(geometry_type="MULTIPOLYGON"))
     type_order = db.Column(db.Integer)
+
+
+# views
+dataset_mv = PGMaterializedView(
```
neilmb, line 224: **question (blocking)**: I don't see any tests for how this view is supposed to be used from Python. 

I see the library that allows declaring this object here in the models and then auto-generating the migrations from it, but what does this `dataset_mv` instance do here? Can we query from it? If so, how about some database tests that verify that.

neilmb (changes requested):

database/models.py
rshewitt: yeah sure i'll add those tests

neilmb:
@rshewitt Did we decide not to use this PR?

rshewitt:
> @rshewitt Did we decide not to use this PR?

yeah i'm cool with ditching this

neilmb:
We have a plan for moving forward without this.

PR #448, opened 2025-10-01
Org update

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5493
- https://github.com/GSA/data.gov/issues/5461
- https://github.com/GSA/data.gov/issues/5462

## About
Adding slug for #5493, and complete some other ticket in one scoop.
- add slug
- add description
- expose organization_type

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

TODO:
DB migration: for this PR, slug is a nullable field. after deploy and all slug column filled with a valid string, we will do another db migration to make it non-nullable field.


app/forms.py
```diff
@@ -85,6 +91,53 @@ class OrganizationForm(FlaskForm):
...
+
+    def validate_slug(self, field):
+        from database.models import Organization, db
+
+        query = db.session.query(Organization.id).filter(Organization.slug == field.data)
+
+        if self.organization_id:
+            query = query.filter(Organization.id != self.organization_id)
+
+        if query.first() is not None:
```
neilmb: **suggestion (non-blocking)**: I think that this is idiomatically `if query.exists():` in SQLAlchemy https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.exists

app/forms.py
```diff
@@ -85,6 +91,53 @@ class OrganizationForm(FlaskForm):
...
+            raise ValidationError("Slug must be unique.")
+    organization_type = SelectField(
```
neilmb: **question (non-blocking)**: Is this meant to go with the other fields up at the top of the class?

app/forms.py
```diff
@@ -85,6 +91,53 @@ class OrganizationForm(FlaskForm):
...
+        "Organization Type",
+        choices=[
+            ("", "Select an organization type"),
```
neilmb: **question (non-blocking)**: This seems odd to have to repeat these choices here when they are already defined in `database/models.py`. Can we use a python `enum.Enum` type there and then import that type here and use it to construct this choices list?

migrations/versions/7a5d0b40bd85_add_description_and_slug_to_organization.py
neilmb, line 1: **suggestion (non-blocking)**: I'd prefer both of these column additions to be done in one migration because we added both fields to the model at once. It should be as simple as copying the upgrade and downgrade commands into the other file.

neilmb (dismissed):
Some questions and suggestions, but this looks great and the test case for duplicate slugs is very nice.

neilmb (dismissed):

neilmb (dismissed):

FuhuXia:
TODO:
Need to do DB migration before deployment.

FuhuXia:
Did a minor change to get ready to deployment. Can someone approve it?

neilmb (approved):

PR #449, opened 2025-10-03
save transformed data

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5194

## About

- new db column `source_transform` with type `JSONB` added 
- transformed data is added after some edits by other function(`add_parent` and `fill_placeholders`)
## TODO

update the migration down_revision right before merging
 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM. Please note that this is only in the transformed case, not in the raw case. When we use this field (either for catalog, or for secondary view of source data information) we need to note that this field is sparse.

PR #451, opened 2025-10-03
Nmb/5487 dataset model

neilmb:
# Pull Request

**This PR targets the `develop` branch.**

This model got dropped when we merged it to develop and then over-wrote develop with main. I took out the migrations for the `Dataset` table, so the model is defined here, but the database won't work with this code without manual intervention on the database.

This should be fine on the `develop` branch since that environment has the table (view) we need. We shouldn't merge this to `main` until we work out an Alembic migration on `develop` that creates a view to go along with this model.

## About

We should also update the `Check DB sync` action in `datagov-catalog` to check against the `develop` branch of this repo.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


FuhuXia:
The `develop` branch was treated as a disposable branch, where anyone could add test code without caution, since it could always be hard reset to `main` if something went wrong. This change will require the team to treat `develop` differently.

neilmb:
> The `develop` branch was treated as a disposable branch, where anyone could add test code without caution, since it could always be hard reset to `main` if something went wrong. This change will require the team to treat `develop` differently.

I completely agree. I think that with the work rushing ahead on `datagov-catalog` in the `development` environment and the changes needed in this code base to support it, I think it makes sense to start treating this repo's `develop` branch as a long-lived feature branch.

FuhuXia (approved):

FuhuXia:
We can use submitting PRs to keep a list of sticky commits you want to stay on the non-protected `develop` branch. In case of hard reset, we can come back to the PRs to rebuild.

PR #454, opened 2025-10-06
add org and source id to the view

FuhuXia:
## About

Added `organization_id` and `source_id` to dataset to make it easy for querying. 

This dataset table is supposed to be a view, a `ForeignKey` does not make sense, therefore I did not define it in the new fields, also removed it from `harvest_record_id` to be consistent.

### DIFF
--- a/database/models.py
+++ b/database/models.py
@@ -5,7 +5,7 @@ import uuid
 from flask_sqlalchemy import SQLAlchemy
 from geoalchemy2 import Geometry
 from sqlalchemy import CheckConstraint, Column, Enum, String, func
-from sqlalchemy.dialects.postgresql import JSONB, TSVECTOR
+from sqlalchemy.dialects.postgresql import JSONB, TSVECTOR, UUID
 from sqlalchemy.orm import DeclarativeBase, backref
 
 from shared.constants import ORGANIZATION_TYPE_VALUES
@@ -193,16 +193,32 @@ class Dataset(db.Model):
     # Base has a string `id` column that is uuid by default
 
     # slug is the string that we use in a URL for this dataset
-    slug = db.Column(db.String, nullable=False)
+    slug = db.Column(
+        db.String,
+        nullable=False,
+        index=True,
+        unique=True
+    )
 
     # This is all of the details of the dataset in DCAT schema in a JSON column
     dcat = db.Column(JSONB, nullable=False)
 
-    # JOIN other tables at query time if we need the source and organization
-    # harvest_source is dataset.record.source
-    # organization is dataset.record.source.org
+    organization_id = db.Column(
+        UUID(as_uuid=True),
+        nullable=False,
+        index=True,
+    )
+
+    harvest_source_id = db.Column(
+        UUID(as_uuid=True),
+        nullable=False,
+        index=True,
+    )
+
     harvest_record_id = db.Column(
-        db.String(36), db.ForeignKey("harvest_record.id"), nullable=False
+        UUID(as_uuid=True),
+        nullable=False,
+        index=True,
     )
 
     popularity = db.Column(db.Numeric)
--- a/database/models.py
+++ b/database/models.py
@@ -50,6 +50,11 @@ class Organization(db.Model):
         cascade="all, delete-orphan",
         lazy=True,
     )
+    datasets = db.relationship(
+        "Dataset",
+        backref=backref("organization", lazy="joined"),
+        cascade="all, delete-orphan",
+    )
 
 
 class HarvestSource(db.Model):
@@ -112,6 +117,11 @@ class HarvestSource(db.Model):
         nullable=False,
     )
     collection_parent_url = db.Column(db.String)
+    datasets = db.relationship(
+        "Dataset",
+        backref="harvest_source",
+        cascade="all, delete-orphan",
+    )
 
 
 class HarvestJob(db.Model):
@@ -184,6 +194,12 @@ class HarvestRecord(db.Model):
     parent_identifier = db.Column(db.String)
     status = db.Column(Enum("error", "success", name="record_status"), index=True)
     errors = db.relationship("HarvestRecordError", backref="record", lazy=True)
+    datasets = db.relationship(
+        "Dataset",
+        backref="record",
+        cascade="all, delete-orphan",
+        lazy=True,
+    )
 
 
 class Dataset(db.Model):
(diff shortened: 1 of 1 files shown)


database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
+        unique=True
+    )
 
     # This is all of the details of the dataset in DCAT schema in a JSON column
     dcat = db.Column(JSONB, nullable=False)
 
-    # JOIN other tables at query time if we need the source and organization
-    # harvest_source is dataset.record.source
-    # organization is dataset.record.source.org
+    organization_id = db.Column(
```
neilmb, line 205: **suggestion (non-blocking)**: I think it would be nice to also add the ORM relationship on the `Organization` model (like https://github.com/GSA/datagov-harvester/blob/main/database/models.py#L47-L52) to look up the organization's datasets as `org.datasets` and the reverse property of `dataset.organization` which is even more useful.

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
+        UUID(as_uuid=True),
+        nullable=False,
+        index=True,
+    )
+
+    harvest_source_id = db.Column(
```
neilmb, line 211: **suggestion (non-blocking)**: As above, I think it's nice to declare the forward and reverse relationships in the ORM.

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
+    )
 
     # This is all of the details of the dataset in DCAT schema in a JSON column
     dcat = db.Column(JSONB, nullable=False)
 
-    # JOIN other tables at query time if we need the source and organization
-    # harvest_source is dataset.record.source
-    # organization is dataset.record.source.org
+    organization_id = db.Column(
+        UUID(as_uuid=True),
```
neilmb: **question (non-blocking)**: This isn't the datatype for organization IDs, they are just strings. Doesn't this column need to match the type of the organization ID?

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
+        nullable=False,
+        index=True,
+    )
+
+    harvest_source_id = db.Column(
+        UUID(as_uuid=True),
```
neilmb: question (non-blocking): As above, shouldn't this column have the same type as the Harvest Source ID?

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
+        nullable=False,
+        index=True,
+    )
+
     harvest_record_id = db.Column(
-        db.String(36), db.ForeignKey("harvest_record.id"), nullable=False
+        UUID(as_uuid=True),
```
neilmb: question (non-blocking): As above, shouldn't this column have the same type as the Harvest Record ID?

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
     # Base has a string `id` column that is uuid by default
 
     # slug is the string that we use in a URL for this dataset
-    slug = db.Column(db.String, nullable=False)
+    slug = db.Column(
+        db.String,
+        nullable=False,
+        index=True,
```
neilmb, line 198: **comment (non-blocking)**: @rshewitt since we aren't using migrations (yet?) to produce the underlying view, we will need to add `CREATE INDEX` statements on these columns after we create the materialized view.

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
     # This is all of the details of the dataset in DCAT schema in a JSON column
     dcat = db.Column(JSONB, nullable=False)
 
-    # JOIN other tables at query time if we need the source and organization
-    # harvest_source is dataset.record.source
-    # organization is dataset.record.source.org
+    organization_id = db.Column(
+        UUID(as_uuid=True),
+        nullable=False,
+        index=True,
```
neilmb, line 208: **comment (non-blocking)**: @rshewitt As above, we will need to add `CREATE INDEX` statements on these columns after we create the materialized view.

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
+    )
+
+    harvest_source_id = db.Column(
+        UUID(as_uuid=True),
+        nullable=False,
+        index=True,
```
neilmb, line 214: comment (non-blocking): @rshewitt As above, we will need to add CREATE INDEX statements on these columns after we create the materialized view.

database/models.py
```diff
@@ -193,16 +193,32 @@ class Dataset(db.Model):
...
+    )
+
     harvest_record_id = db.Column(
-        db.String(36), db.ForeignKey("harvest_record.id"), nullable=False
+        UUID(as_uuid=True),
+        nullable=False,
+        index=True,
```
neilmb, line 220: comment (non-blocking): @rshewitt As above, we will need to add CREATE INDEX statements on these columns after we create the materialized view.

rshewitt:
@neilmb yes, we need to create those indexes. my [PR](https://github.com/GSA/datagov-catalog/pull/11/files) for updating the mv sql command in the readme. if it can be done in the model definition then great but i've been doing it in raw sql as seen in the readme.

FuhuXia:
So this TABLE will never be created/populated in real. In the development space an equivalent VIEW will be manually created. Whatever constraint or index  we define here do not matter.

What about foreign keys? If we define foreign key relationship, what behaviors do we expect when the upstream record got deleted? cascade or not, the FLASK/DB will try to do something to the TABLE but it is an actual VIEW. We might get into weird issue. That is why I did not define any relationship.

FuhuXia:
Back to the fundamental question. why view, not a real table?

database/models.py
FuhuXia: the index will not be materialized since an in reality equivalent view will be created manually. This is just a reference what constraint we need for the view design.

FuhuXia:
ok, I am going to make it fully loaded table with all constraint added. If the constraint does not exist in the DB, the app might continue to work. Data integrity will be an issue but the data is in a view that gets refreshed regularly. We might be OK.

database/models.py
FuhuXia: added

database/models.py
FuhuXia: the index will not be materialized since in reality an equivalent view will be created manually. This is just a reference what constraint we need for the view design.

database/models.py
FuhuXia: added

FuhuXia:
it turns out we cant fool ORM with FKs that does not exist on materialized views. ORM will try to update the views when referenced record is deleted. updating materialized views is not allowed.

All FK and backref are removed from the Dataset model.

database/models.py
```diff
@@ -193,22 +192,45 @@ class Dataset(db.Model):
...
 
     popularity = db.Column(db.Numeric)
-    last_harvested_date = db.Column(db.DateTime)
+    last_harvested_date = db.Column(
+        db.DateTime,
+        index=True
+    )
     search_vector = db.Column(TSVECTOR)
 
+    __table_args__ = (
```
neilmb, line 230: **question (non-blocking)**: Seriously non-blocking, I just want to understand better, does this help at the ORM layer? Or is it just a breadcrumb that we should be using the same kind of index when we create our materialized view?

neilmb (approved):

database/models.py
FuhuXia: All indexes defined here does not matter, including this one. I added here because I added all other indexes trying to give the view a reference what index we need, so why not make it complete.

PR #468, opened 2025-10-21
make dcat JSONB mutable

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5511

## About
added a mutable wrapper so that SQLAlchemy is aware of the addition of a `spatial` field during tests.  

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/database/models.py
+++ b/database/models.py
@@ -201,7 +201,8 @@ class Dataset(db.Model):
     )
 
     # This is all of the details of the dataset in DCAT schema in a JSON column
-    # make it mutable so that in-place mutations (e.g., dcat["spatial"] = "...", for tests) are tracked
+    # make it mutable so that in-place mutations (e.g.,
+    # dcat["spatial"] = "...", for tests) are tracked
     dcat = db.Column(MutableDict.as_mutable(JSONB), nullable=False)
 
     organization_id = db.Column(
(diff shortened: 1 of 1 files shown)


database/models.py
```diff
@@ -200,7 +201,9 @@ class Dataset(db.Model):
     )
 
     # This is all of the details of the dataset in DCAT schema in a JSON column
-    dcat = db.Column(JSONB, nullable=False)
+    # make it mutable so that in-place mutations (e.g.,
+    # dcat["spatial"] = "...", for tests) are tracked
+    dcat = db.Column(MutableDict.as_mutable(JSONB), nullable=False)
```
jbrown-xentity, line 206: I believe this to be a cosmetic SQL-Alchemy change and not a PostgreSQL change, so I'm tentatively approving. If this makes changes in PostgreSQL, then I have concerns and it should be manually deployed...

jbrown-xentity (approved):
Approving, see comment.

database/models.py
FuhuXia: Valid concern. This is indeed cosmetic, no DDL change to the PostgreSQL DB.

PR #469, opened 2025-10-23
add DatasetViewCount

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/datagov-catalog/pull/31

## About

create a table to hold view count for each dataset slug.

### DIFF
--- a/database/models.py
+++ b/database/models.py
@@ -242,6 +242,13 @@ class DatasetViewCount(db.Model):
     view_count = db.Column(db.Integer, nullable=False, default=0)
 
 
+class ResourceViewCount(db.Model):
+    __tablename__ = "resource_view_count"
+    # url from Google Analytics resource path is truncated to 100 characters
+    resource_url = db.Column(db.String(100), nullable=False, unique=True, index=True)
+    view_count = db.Column(db.Integer, nullable=False, default=0)
+
+
 class HarvestJobError(Error):
     __tablename__ = "harvest_job_error"
 
--- a/database/models.py
+++ b/database/models.py
@@ -235,6 +235,13 @@ class Dataset(db.Model):
     )
 
 
+class DatasetViewCount(db.Model):
+    __tablename__ = "dataset_view_count"
+
+    dataset_slug = db.Column(db.String(100), nullable=False, unique=True, index=True)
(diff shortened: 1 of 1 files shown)


database/models.py
```diff
@@ -235,6 +235,13 @@ class Dataset(db.Model):
     )
 
 
+class DatasetViewCount(db.Model):
+    __tablename__ = "dataset_view_count"
+
+    dataset_slug = db.Column(db.String(100), nullable=False, unique=True, index=True)
```
neilmb, line 241: **comment (non-blocking)**: This is great as-is, just noting for posterity that another consequence of our materialized view decision is we can't put any consistency or foreign-key constraints on this column, so we will need to be careful when we extract this data to match it up with Datasets.

database/models.py
```diff
@@ -235,6 +235,13 @@ class Dataset(db.Model):
...
+    view_count = db.Column(db.Integer, nullable=False, default=0)
```
neilmb, line 242: **question (non-blocking)**: Do we anticipate over-writing these every time we import view data? Or do we want to add an indexed timestamp column for when the data was imported and only append new rows and then do a `GROUP BY slug` and take the latest data when we use it in our Dataset view?

neilmb (approved):
Totally reasonable change, I'm just blathering on.

database/models.py
FuhuXia: No foreign-key needed. The data is from GA export. It is ok to have some slug that is not in dataset view, and vice versa.

rshewitt: create a FK against ckan_name (which is our slug) in HarvestRecord?

database/models.py
FuhuXia: This table is supposed to get last two weeks (or 30 days) view count from GA. it is not intended to hold any all-time count, or any historical data. So it will be a overwriting.

database/models.py
FuhuXia: > create a FK against ckan_name (which is our slug) in HarvestRecord?

No need. Plus the HarvestRecord.ckan_name itself do not have a unique key constraint therefore not qualified to be a FK.

FuhuXia:
added another table `resource_view_count`

neilmb (approved):

PR #479, opened 2025-11-05
stop sync with ckan

rshewitt:
# Pull Request

Related to [#5532](https://github.com/GSA/data.gov/issues/5532)

## About
The purpose of this PR is to demonstrate the least amount of change (excluding tests) required to stop syncing with ckan while including a uniqueness function for slug creation which ckan handled for us. ckan data and utilities, for the most part, remain (removing those would be another commit/pr).

My proposal is that we continue to use the materialized view and focus our attention on the UI. The dataset model is an optimization and so far there's no evidence indicating the materialized view is insufficient (e.g. issues from a lack of foreign keys). For any miscellaneous data we want to add (e.g. converted spatial) we store it in an "extras"-type field. for example, 
```json
{
  "new_spatial": "data",
  "misc_field1": "other data", 
  "misc_field2": "data 2"
}
```

I haven't fixed the tests after these changes but intend to.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


harvester/harvest.py
```diff
@@ -961,22 +955,48 @@ def validate(self) -> None:
...
-            CKANDownException,
-            CKANRejectionException,
-        ):
+
+            if self.action == "create":
+                metadata = (
+                    json.loads(self.source_raw)
+                    if self.transformed_data is None
+                    else self.transformed_data
+                )
```
jbrown-xentity, line 981: I understand the only time we should be generating a slug is in the "create" case. In the "update" case, will the record already naturally have the slug (or ckan_name), or would that need to be pulled into that record?

rshewitt: on "update" the record would have the slug/ckan_name from the "create"

neilmb (changes requested):
This is great to see. Can you please also clean up the `CKAN_` environment variables in the manifest and the .profile and the Github action workflows?

neilmb (approved):
Great, thanks.

PR #484, opened 2025-11-14
Add aliases field to Organization

neilmb:
# Pull Request

Related to GSA/Data.gov#5562, we want to be able to add "aliases" to our Organizations so we can offer a better search experience in the public-facing catalog app.

## About

This adds the field to the model/database schema, we need to do some work on our production/staging database to add the desired alias values for our existing organizations.

## PR TASKS

- [x] Code well documented
- [x] Files linted


FuhuXia:
Does the input `first, first, second, sEcoNd` result in 2, 3, or 4 aliases?

neilmb:
> Does the input `first, first, second, sEcoNd` result in 2, 3, or 4 aliases?

Four. We are setting these values and we can repeat ourselves as many times as we want. The catalog organization search will work just the same.

FuhuXia (approved):

neilmb:
Watching the deploy and making sure migrations succeed.

neilmb:
Migrations are not succeeding on staging, more debugging.

PR #485, opened 2025-11-18
Sync dataset

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5555

## About

- Sync to dataset table after records are harvested
- use command `flask dataset check --apply` to backfill existing records to dataset table
- remove trace of CKAN code
- add translated spatial to dataset table

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


database/interface.py
```diff
@@ -563,6 +564,110 @@ def delete_harvest_record(
...
+            except Exception:
+                self.db.rollback()
+                raise
+            if dataset is None:
+                dataset = self.get_dataset_by_slug(slug)
+            if dataset is not None:
+                self.db.refresh(dataset)
+            return dataset
+
+    def delete_dataset_by_slug(self, slug: str) -> bool:
```
neilmb, line 649: **suggestion (non-blocking)**: Docstring please. In particular, what gets returned if there is no dataset with that slug?

database/models.py
```diff
@@ -187,6 +187,14 @@ class HarvestRecord(db.Model):
     status = db.Column(Enum("error", "success", name="record_status"), index=True)
     errors = db.relationship("HarvestRecordError", backref="record", lazy=True)
 
+    @property
+    def dataset_slug(self) -> Optional[str]:
+        dataset = getattr(self, "dataset", None)
+        if dataset is None:
+            return None
+        return dataset.slug
+        return dataset.slug
```
neilmb: Typo?

```suggestion
```

harvester/utils/general_utils.py
```diff
@@ -427,6 +430,230 @@ def validate_geojson(geojson_str: str) -> bool:
...
+
+    if len(string) < min_length:
+        string += "_" * (min_length - len(string))
+    if len(string) > max_length:
+        string = string[:max_length]
+    return string
+
+
+def substitute_ascii_equivalents(text_unicode: str) -> str:
+    """Replace Latin-1 characters with ASCII equivalents when possible."""
```
neilmb: **question (non-blocking)**: If we grabbed this code from somewhere else, can we include a reference to where it originated?

harvester/utils/general_utils.py
```diff
@@ -427,6 +430,230 @@ def validate_geojson(geojson_str: str) -> bool:
...
+        '{{"type": "Polygon", "coordinates": [[[{minx}, {miny}], '
+        '[{minx}, {maxy}], [{maxx}, {maxy}], [{maxx}, {miny}], [{minx}, {miny}]]]}}'
+    )
+    geojson_point_tpl = '{{"type": "Point", "coordinates": [{x}, {y}]}}'
+
+    spatial_value = spatial_value.replace("+", "")
+    spatial_value = spatial_value.replace(".,", ",").replace(".]", "]")
+    if spatial_value and spatial_value[-1] == ".":
+        spatial_value = spatial_value[:-1]
+    spatial_value = re.sub(
```
neilmb, line 570: **suggestion (non-blocking)**: The substitution here isn't obvious to me what it's being used for. Could we have a comment on what this is intended to do?

harvester/utils/general_utils.py
```diff
@@ -427,6 +430,230 @@ def validate_geojson(geojson_str: str) -> bool:
...
+        except Exception:  # pragma: no cover - defensive lookup
+            pass
+
+    if isinstance(spatial_value, str):
+        return munge_spatial(spatial_value)
+
+    return ""
+
+
+def translate_spatial_to_geojson(spatial_input):
```
neilmb, line 645: **comment (non-bloocking)**: Can we add to this docstring about when this method might return None?

tests/unit/test_sync_datasets.py
```diff
@@ -0,0 +1,82 @@
...
+
+    result = _records_missing_datasets(session)
+
+    assert result is query
+    session.query.assert_called_once()
+    session.query.return_value.join.assert_called_once()
+    session.query.return_value.join.return_value.outerjoin.assert_called_once()
+
+
+def test_records_missing_datasets_excludes_failed_ids(session):
```
neilmb, line 55: **suggestion (non-blocking)**: This feels subtle, can we have a docstring here to explain that we are excluding two ids so we expect two calls to `filter`?

tests/unit/test_sync_datasets.py
```diff
@@ -0,0 +1,82 @@
...
+    session.query.assert_called_once()
+    session.query.return_value.join.assert_called_once()
+
+
+def test_datasets_with_unexpected_records_excludes_failed_ids(session):
+    query = session.query.return_value.join.return_value
+
+    _datasets_with_unexpected_records(session, exclude_ids=[3])
+
+    assert query.filter.call_count == 2
```
neilmb, line 94: **comment (non-blocking)**: Again, this feels subtle. Can we have a comment to explain that one excluded ID combines with one other filter call to get this magic number?

neilmb (dismissed):
This is great, so glad to see the CKAN leaving our system. Nothing big at all, comments about docstrings, etc.

cmhedrickREI (approved):
LGTM

PR #508, opened 2025-12-16
Use the same user-agent as CKAN harvester

neilmb:
# Pull Request

Related to GSA/Data.gov#5602

## About

Changes the user-agent header to match (hopefully exactly) the string that we currently use in our CKAN harvester which is found at https://github.com/GSA/ckanext-datajson/blob/105aaf01f70d053fabeefb61de9dff49f3e707bf/ckanext/datajson/harvester_datajson.py#L29


jbrown-xentity (approved):
LGTM

PR #509, opened 2025-12-16
support 'pre' waf html

rshewitt:
# Pull Request

Related to [#5579](https://github.com/GSA/data.gov/issues/5579)

## About
- census waf sources have changed how documents are presented in the html. so far sources use `td` elements. this add support for text within a `pre` element.

## PR TASKS

- [x] Code well documented
- [x] Files linted


jbrown-xentity (approved):
LGTM

PR #517, opened 2026-01-14
update packages to rid snyk issues

cmhedrickREI:
# Pull Request

- updates:
    - aiohttp (poetry add aiohttp@3.13.3)
- add SNYK-PYTHON-WERKZEUG-14908843 to ignore because the affected package is for local dev only.    

## PR TASKS

- [x] Code well documented
- [x] Files linted


rshewitt (changes requested):
so for aiohttp and urlib just `poetry add [library]@[version]`  (e.g. `poetry add aiohttp@3.13.3`). you can actually probably just say `poetry add [library]` and that'll bump it to the latest. that'll update the poetry.lock file. for werkzeug you can add that to the .synk file. so the changes should be poetry.lock, pyproject.toml, and .synk

rshewitt:
still need [urllib](https://github.com/GSA/datagov-harvester/actions/runs/21041805107/job/60505355045)

PR #520, opened 2026-01-20
use latest poetry

FuhuXia:
use latest poetry in CI

### DIFF
--- a/.github/workflows/commit.yml
+++ b/.github/workflows/commit.yml
@@ -6,7 +6,7 @@ permissions:
 
 env:
   PY_VERSION: "3.12"
-  POETRY_VERSION: "2.0.0"
+  POETRY_VERSION: "latest"
 
 jobs:
   lint:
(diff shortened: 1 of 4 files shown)


neilmb (dismissed):
**question (non-blocking)**: Do we also need to change anything in our `pyproject.toml` or `poetry.lock` so that the poetry version used locally matches that on the Github Action?

rshewitt:
from what i've seen it's not common to configure the poetry version in `pyproject.toml` or the lock file. we could add a make command? 
```make
poetry-install
  pipx install poetry # should install the latest version
```

FuhuXia:
it was a mistake that we pin POETRY_VERSION to 2.0.0 on CI. it should not pin to anything so that we always use the latest, regardless it is local or CI, until one day the latest POETRY_VERSION gives us issues.

neilmb (approved):
And tests pass now!

neilmb:
@FuhuXia after you merge this, can you comment "@dependabot rebase" on the two dependabot PRs to incorporate this change so their tests can pass?

PR #521, opened 2026-01-20
fix duplicate point issue for point geometries

rshewitt:
# Pull Request

Related to [5647](https://github.com/GSA/data.gov/issues/5647)

## About
when we get a spatial value like `"-92.109, 15.132, -92.109, 15.132"` it ends up being transformed into a polygon geojson when it should be a point geojson this causes opensearch sync failures

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


neilmb (approved):
Looks like a reasonable decision for dealing with strange input data.

neilmb:
FYI @rshewitt you probably need to merge main with Fuhu's "latest poetry" fix to get the tests to pass.

PR #540, opened 2026-02-02
remove and log dcatus records without identifier

rshewitt:
# Pull Request

Related to [#5480](https://github.com/GSA/data.gov/issues/5480)

## About
[commerce-non-spatial-data-json-harvest-source](https://harvest.data.gov/harvest_source/7660cf65-6dbc-49a7-9ebb-6c9736fcb006) has been failing for the past couple weeks because of an unhandled exception. the cause of the exception is because one of the records in the source doesn't have an "identifier" field when it's expected. i've emailed the provider.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -27,6 +27,7 @@ from harvester.exceptions import (
     ExternalRecordToClass,
     ExtractExternalException,
     ExtractInternalException,
+    NoIdentifierException,
     SendNotificationException,
     TransformationException,
     log_non_critical_error,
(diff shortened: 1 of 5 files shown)


harvester/harvest.py
```diff
@@ -297,6 +298,44 @@ def filter_waf_files_by_datetime(self) -> None:
                 records.append(data)  # create
         self.external_records = records
 
+    def filter_datasets_with_no_identifier(self) -> None:
+        """
+        identifies datasets without an 'identifier' field and logs the error in the db
```
neilmb: **suggestion (non-blocking)**: Can you be more clear about the side-effects of this method? i.e. that it changes the self.external_records list?

neilmb (dismissed):
LGTM, just a quibble about a docstring

harvester/harvest.py
rshewitt: done

neilmb (approved):

PR #542, opened 2026-02-10
Generate Openapi spec "automatically"

neilmb:
# Pull Request

Related to GSA/Data.gov#4732

## About

As requested in the ticket, to produce an OpenAPI spec for the read-only part of our API, we use the [APIFlask](https://apiflask.com/) library. It extends Flask's `Blueprint` with a new `APIBlueprint` class that instruments the routes in that blueprint and documents them in to the `openapi.json` file. To make it work right, we had to make some other substantial changes:

1. Commands have been taken out of the `app/routes.py` file and put into an `app/commands/` module following the pattern we use in `datagov-catalog`.
2. All enumerated values from `database/models.py` got moved into `shared/constants.py` so they can be re-used in the documentation.
3. Lots of `url_for` calls changed as the methods moved from the `main` blueprint into the `api` blueprint.
4. Our `login_required` decorator was restructured to work the APIFlask `auth` system so if we eventually want to produce OpenAPI spec for the full authenticated API, that configuration is ready.

## PR TASKS

- [ ] Code well documented
- [x] Files linted

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,8 +1,8 @@
 import logging
 import os
 
+from apiflask import APIFlask
 from dotenv import load_dotenv
-from flask import Flask
 from flask_bootstrap import Bootstrap
 from flask_htmx import HTMX
 from flask_migrate import Migrate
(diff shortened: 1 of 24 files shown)


cmhedrickREI (approved):
The openapi.json looks good to me! I used the explorer to look at the rendering of the spec and it looks excellent! Thanks for also taking the time to reorganize all the routes!

PR #543, opened 2026-02-10
add db record error when spatial value cant be translated

rshewitt:
# Pull Request

Related to [5576](https://github.com/GSA/data.gov/issues/5576)

## About
- add db record error when spatial value cant be translated

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


cmhedrickREI (approved):
Looks good!

PR #546, opened 2026-02-11
support successful record on non critical error log

rshewitt:
# Pull Request

Related to [5576](https://github.com/GSA/data.gov/issues/5576)

## About
harvest record isn't erroneous just because we can't translate the spatial value.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/exceptions.py
+++ b/harvester/exceptions.py
@@ -57,7 +57,9 @@ class ClearJobException(HarvestCriticalException):
     pass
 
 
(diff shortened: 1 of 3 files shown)


neilmb (changes requested):
The code LGTM. 

Because we are adding this so that we can get information for this spatial translation error into the harvest's reporting email, can we make a test case that explicitly tests that "round-trip" behavior, where we harvest a source with this kind of translation error and then verify that that (mock) report email has that information in it?

neilmb (approved):

PR #547, opened 2026-02-13
slightly improve validation message format for regexes

rshewitt:
# Pull Request

Related to [5349](https://github.com/GSA/data.gov/issues/5349)

## About
reduces backslashes from 4 to 2 (e.g. `\\\\` to `\\`). was fiddling around with it but couldn't figure out how to get it to just have one.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


neilmb (approved):
2 is better than 4 I guess, too bad we can't get down to 1, but thanks for this.

PR #552, opened 2026-02-23
Organization edit bug

neilmb:
# Pull Request

Editing orgs on H2.0 gives an internal server error.

## About

Our filter was missing a type check, this should fix it.

## PR TASKS

- [X] Code well documented
- [X] Files linted


app/forms.py
```diff
@@ -35,6 +35,8 @@ def comma_separated_filter(data):
     # of a list
     if isinstance(data, list):
         return ", ".join(data)
+    if not isinstance(data, str):
+        return data
```
FuhuXia, line 38: If we want to handle None, can we be explicit about it?

FuhuXia (dismissed):

app/forms.py
neilmb: I added a better comment about the idea. Is that better?

FuhuXia (approved):

PR #554, opened 2026-02-27
Bootstrap5

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5743

Changes:
- bump up bootstrap
- add custom css to preserve old look and feel

Deployed on dev
https://datagov-harvest-dev.app.cloud.gov/
image


neilmb (dismissed):
I totally support this PR as a compliance exercise and I'm really glad to see the jquery dependency go away. 

However, I also think it's bonkers that we are using both Bootstrap and USWDS and I think that the right thing to do in the future is to replatform the harvester app to just use USWDS for page layout and drop the Bootstrap dependency entirely. I'll make a ticket to that effect.

neilmb (approved):
Hopefully tests pass

PR #555, opened 2026-02-27
add validator page and api route

rshewitt:
# Pull Request

Related to [4638](https://github.com/GSA/data.gov/issues/4638)

## About
adds dcatus catalog validator page and api route

## PR TASKS

- [x] Code well documented
- [x] Files linted


neilmb:
Playwright tests need to be updated to account for the new header link.

app/templates/view_validators.html
```diff
@@ -0,0 +1,117 @@
...
+
+{% block content %}
+
+<h2>JSON Schema Validator</h2>
+
+<form method="POST">
+    {{ form.hidden_tag() }}
+    
+    <!-- schema and fetch_method can't be empty so they won't have errors to display -->
+    <div class="form-group">
```
neilmb, line 11: **suggestion**: We should use the USWDS form classes here. There's an example in https://designsystem.digital.gov/templates/form-templates/address-form/

app/templates/view_validators.html
```diff
@@ -0,0 +1,117 @@
...
+        </div>
+        {% endif %}
+    {% endif %}
+</div>
+
+{% endblock %}
+
+{% block scripts %}
+
+<script>
```
neilmb: **suggestion (blocking)**: Inline scripts need the CSP nonce.
```suggestion
<script nonce="{{ csp_nonce() }}">
```

app/templates/view_validators.html
```diff
@@ -0,0 +1,117 @@
...
+    </div>
+</form>
+
+<div class="section">
+    {% if submitted %}
+        <h2>Validation Results</h2>
+        {% if not data.record_errors %}
+            <p>No validation errors found</p>
+        {% else %}
+        <div class="usa-table-container" tabindex="0">
```
neilmb: **question (non-blocking)**: There isn't a table here, why do we need a table container div around this?

app/templates/view_validators.html
```diff
@@ -0,0 +1,117 @@
...
+            <div class="error-list">
+                {% for record_error in data.record_errors %}
+                <div class="error-block">
```
neilmb: **question (non-blocking)**: Could the USWDS alert component workfor these divs? https://designsystem.digital.gov/components/alert/

It has a header and body that could do the same as the two `<p>` blocks here?

app/util.py
```diff
@@ -44,3 +64,93 @@ def make_new_org_contract(form):
...
+            ):
+                return False
+        return True
+    except Exception:
+        return False
+
+
+def fetch_json_from_url(url: str) -> dict:
+
+    MAX_CONTENT_LENGTH = 10 * 1024 * 1024  # 10MB limit
```
neilmb: **suggestion (non-blocking)**: This is a local variable so its name shouldn't be all caps which is usually reserved for module-level constants.

neilmb (dismissed):
A few nits, but otherwise this is good to see.

app/templates/view_validators.html
rshewitt: the form for adding & editing harvest sources and organizations all use this bootstrap class `form-group` which is where i got it. none of the remaining forms we have in harvester use uswds (are `config-actions` or `[mode]-config-actions` uswds classes?). I can create a ticket to fix all forms to use uswds rather than adding that work to this PR.

app/templates/view_validators.html
rshewitt: copy/paste situation. good find!

app/templates/view_validators.html
rshewitt: we use `error-block` when displaying errors on a job page (which applies to validation errors) so I use it here for continuity. would you like me to create a ticket to replace both with uswds?

app/util.py
rshewitt: I was planning on moving it to the top as a module level variable but decided against it. i updated it.

neilmb (dismissed):

app/util.py
```diff
@@ -44,3 +66,93 @@ def make_new_org_contract(form):
...
+                or ip_obj.is_link_local
+                or ip_obj.is_multicast
+            ):
+                return False
+        return True
+    except Exception:
+        return False
+
+
+def fetch_json_from_url(url: str) -> dict:
```
neilmb, line 102: **question (non-blocking)**: Is this going to be subject to our egress proxy and so it can only download from .gov or .mil domains? If so, can we provide some explanation of that on the HTML page, and possibly in the API response?

neilmb (commented):
Just one question and also could we get this on the `develop` branch so we can see it at https://datagov-harvest-dev.app.cloud.gov/?

app/util.py
rshewitt: tooltip and error response in browser on dev

Screenshot 2026-03-17 at 9 55 51 AM

neilmb (approved):

PR #559, opened 2026-03-05
Improve lint test actions

neilmb:
# Pull Request

This is the same as GSA/datagov-catalog#193, it fixes the lint action to run all of our checks and then makes the changes so that the new test passes.

## PR TASKS

- [X] Code well documented
- [X] Files linted


FuhuXia (approved):
we need this

PR #560, opened 2026-03-05
Fix local test

FuhuXia:
Fail to run make `test-ci` and `make test-e2e-ci` on local. This should fix it.

- stop those tests/badges changes from showing up in git status
- fix `AssertionError: OpenSearch unavailable` error due to due to local docker disk-based allocation limits.
- dynamic Playwright session auth
- no CF handler on local


neilmb (approved):
So glad to get rid of the playwright auth issue.

PR #561, opened 2026-03-05
Get org by slug

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5682

## About
All access organization page by slug.
11ty /metric/ sends harvest source page to url path `/organization/<slug>`;
Changed organization-page URL generation to prefer slugs over UUIDs.


app/routes.py
```diff
@@ -145,6 +145,16 @@ def decorated_function(*args, **kwargs):
     return decorated_function
 
 
+def get_org_by_identifier(org_identifier: str):
```
neilmb: **suggestion (non-blocking)**: This seems like a helper method and could use a name that starts with an underscore: `_get_org_by_identifier`?

app/routes.py
```diff
@@ -311,8 +321,10 @@ def add_organization(**kwargs):
         }
     }
 )
-@valid_id_required
```
neilmb, line 322: **question (non-blocking)**: What error do people get now if they pass in a non-UUID that isn't a dataset slug? Can we document that in the docstring and make a test for that behavior?

neilmb (dismissed):
This looks good, I just want to make sure we have documentation and tests for this kind of polymorphic method that can work with two different types of input.

FuhuXia:
UPDATE:
Changed organization-page URL generation to prefer slugs over UUIDs.

neilmb (approved):
LGTM

cmhedrickREI (approved):
This looks good!

PR #569, opened 2026-03-17
5732 make slug permanent

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#5732

## About

Takes steps to ensure datasets use a consistent slug that we can also update if needed.

Of Dataset's populating on Harvest Source
Example: https://datagov-harvest-dev.app.cloud.gov/harvest_source/41e085a3-995f-4f5f-87c0-10f60986c34c

- Creates a Dataset Detail Page
    - Contains dataset detail information
    - Form to update slug (when authed)
    - Form validation
        - checks for duplicate slug name
        - ensures slug is lowercase
- Creates tests to ensure functionality
- Updates Harvest Source Detail Page to contain table of datasets.
- Updates to the generate_fixtures file to help generate Datasets with the test organization
- Updates `testdata` to now also include a `reset` flag to reset the db if needed

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


cmhedrickREI:
Example of Datasets populating on a Harvest Source (set to page 5 as I was testing pagination)
Screenshot from 2026-03-18 12-08-03

Example of the Dataset Detail Page, with `slug` edit form for atuhed users
Screenshot from 2026-03-18 12-08-20

Example of Dataset Deetail Form withouth the `slug` form for an unauthed user
Screenshot from 2026-03-18 12-10-20

app/routes.py
```diff
@@ -713,6 +758,66 @@ def harvest_source_list():
     return render_template("view_source_list.html", data=data)
 
 
+@main.route("/dataset/<string:dataset_slug>", methods=["POST"])
+@api.get("/dataset/<string:dataset_slug>")
```
neilmb: **suggestion (blocking)**: If this method doesn't ever return JSON, then we should put this entirely in the `main` blueprint so we don't document it along with the rest of our API. If you're committed to making this an API method (I wouldn't, we already have an API on the catalog for datasets), then you need to add an `if request.is_json:` section to the `GET` and document that return type in `api_schema.py`.

database/interface.py
```diff
@@ -673,6 +673,48 @@ def get_dataset_by_slug(self, slug: str):
...
+        Results are ordered by `last_harvested_date` descending so the most
+        recently harvested datasets appear first. Rows with a `NULL`
+        `last_harvested_date` are sorted to the end.
+        """
+        query = (
+            self.db.query(Dataset)
+            .filter(Dataset.harvest_source_id == source_id)
+            .order_by(Dataset.last_harvested_date.desc().nullslast())
+        )
+        if count:
```
neilmb: **suggestion (non-blocking)**: I think we could instead use our `@count` and `@paginate` decorators as on the `get_harvest_record_errors_by_job` method.

tests/playwright/test_openapi.py
```diff
@@ -12,5 +12,5 @@ class TestOpenAPI:
     def test_swagger(self, page):
         expect(page.locator("h2.title")).to_have_text("Datagov Harvester 0.1.0 OAS 3.0")
         expect(page.locator("h3")).to_have_text("Api")
-        expect(page.locator(".opblock-get")).to_have_count(14)
+        expect(page.locator(".opblock-get")).to_have_count(15)
```
neilmb: As above, if we don't have a JSON return path for the dataset object, we don't need this here.

neilmb (changes requested):
One blocking comment about whether the dataset detail view gives an API response or not. 

There was a question on whether `slug` is stable against harvested changes in `title` (since we derive slug from there). Have we verified in a test-case that once a dataset is created, then harvesting it again with a different value for the title won't change the slug?

neilmb (approved):
This is great. Thanks for the additional tests. LGTM

PR #580, opened 2026-04-07
Validator Enhancements

cmhedrickREI:
# Pull Request

Related to GSA/data.gov#4638

## About

- Adds chevron to the select fields
- Only show the first 10 error messages
- Add download button when there are more than 10 errors. Constructs CSV in the browser to avoid taxing the server (e.g: NOAA file errors)
- Playwright tests for download button

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


cmhedrickREI:
Changes are on [DEV](https://datagov-harvest-dev.app.cloud.gov/validate/)
https://datagov-harvest-dev.app.cloud.gov/validate/

Suggested test source by URL: https://data-fairfaxcountygis.opendata.arcgis.com/data.json

rshewitt:
we may want to limit the size of the request body server-side (see below). also, we could set a max character length on the text area input but that can be bypassed. for normal honest users it could be useful. that number'll be obnoxiously high though since we support 10mb for urls. for example, `https://data.transportation.gov/data.json` is a 3.5 MB file containing 3494150 bytes (should be all ASCII though so that should be the character count?). looking back we probably should have used a file upload instead of a text area.
```python 
app.config["MAX_CONTENT_LENGTH"] = 10 * 1024 * 1024 # 10MB limit
```

app/templates/view_validators.html
```diff
@@ -54,19 +54,30 @@ <h2>JSON Schema Validator</h2>
...
 </form>
 
 <div class="section">
     {% if submitted %}
         <h2>Validation Results</h2>
+        {% set total = data.record_errors | length %}
         {% if not data.record_errors %}
             <p>No validation errors found</p>
         {% else %}
+        {% if total > 10 %}
```
rshewitt: we probably want the user to be able to download all errors regardless of how many there are. we still want to limit what's displayed on the page to the first 10 though.

app/templates/view_validators.html
```diff
@@ -54,19 +54,30 @@ <h2>JSON Schema Validator</h2>
...
+        <p><strong>Showing First 10 of {{ total }}</strong></p>
+        <button class="btn btn-success" id="btn-download"
+            title="Download all {{ total }} error(s) as CSV"
+            aria-label="Download all errors as CSV">
+            <i class="fa fa-download me-1" aria-hidden="true"></i>
+            Download all errors (.csv)
+        </button>
+        <script id="all-errors-data" type="application/json">{{ data.record_errors | tojson }}</script>
```
neilmb: **suggestion (blocking)**: I don't think that this inline script will run if it doesn't have the CSP nonce added:
```suggestion
        <script nonce="{{ csp_nonce() }}" id="all-errors-data" type="application/json">{{ data.record_errors | tojson }}</script>
```

neilmb (dismissed):
LGTM. I love the way Playwright handles testing the file download. 

Do you want to merge this as-is and follow up with a new PR changing from paste to file upload? Or do it in this one?

app/templates/view_validators.html
cmhedrickREI: > **suggestion (blocking)**: I don't think that this inline script will run if it doesn't have the CSP nonce added:

You are probably right, let me make this addition at least.

> LGTM. I love the way Playwright handles testing the file download.
> 
> Do you want to merge this as-is and follow up with a new PR changing from paste to file upload? Or do it in this one?

Since I have to touch the template for the nonce I'll make one last adjustment with the button appearing and I say lets merge it. I enjoy seeing green go to purple :)

neilmb (approved):

PR #581, opened 2026-04-09
Opensearch connection

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5607#issuecomment-4208365843

## About

Record update goes to OpenSearch in local, but not in cloud.gov environments. It turns out we need some extra steps in order to for harvester app to talk to OpenSearch.

- make it a required service for the app
- get credentials into cg env
- use correct service name
- bypass proxy

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/.profile
+++ b/.profile
@@ -8,6 +8,9 @@ function vcap_get_service () {
   name="$1"
   path="$2"
   service_name=${APP_NAME}-${name}
+  if [ "$name" = "opensearch" ]; then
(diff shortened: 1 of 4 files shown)


cmhedrickREI (approved):
Thanks for getting everything working! I think once we get the tests tweaked to compensate for the mdtranslator updates this will be good to go 🚀

PR #586, opened 2026-04-13
Return JSON for /harvest_job/ with Accept header

neilmb:
# Pull Request

Related to GSA/Data.gov#5783

## About

This fixes the reported bug that the `/harvest_job/<job_id>` endpoint returns HTML no matter what headers are used unless we provide the `type=json` argument. This changes that endpoint to use the header-based system we use other places, wherre providing either `Content-type: application/json` or `Accept: application/json` leads to the method producing a JSON response.

## PR TASKS

- [X] Code well documented
- [X] Files linted


app/routes.py
```diff
@@ -995,7 +995,10 @@ def _load_json_title(json_string):
         )
     else:
         job = db.get_harvest_job(job_id)
-        if request.args.get("type") and request.args.get("type") == "json":
+        if request.is_json or (
+            request.accept_mimetypes.accept_json
```
rshewitt, line 999: seems fine. looking back we should have separated view routes and api routes entirely so `/api/*` always return JSON and `/organization/*`, for example, always returns html (excluding posts i guess though?)

rshewitt (approved):

app/routes.py
neilmb: We have a ticket for that I think? https://github.com/GSA/data.gov/issues/5610

PR #587, opened 2026-04-13
add harvest source count per org

rshewitt:
# Pull Request

Related to [5753](https://github.com/GSA/data.gov/issues/5753)

## About
adds a harvest source count per org for `/organizations/`. it's not a new column so no need for a migration it's computed at query time.

### correlate except explanation
correlation is needed when subqueries reference thing in outer queries. here's what the child counting query looks like as a sql statement
```sql
SELECT
  organization.id,
  (
    SELECT count(harvest_source.id)
    FROM harvest_source
    WHERE harvest_source.organization_id = organization.id
  ) AS source_count
FROM organization;
```

`organization` is referenced within the inner query calculating `source_count` but it's not in the `FROM` statement. how does it know where to derive from? from the outer query. that's correlation. `.correlate_except` says to correlate the subquery with everything from the outer query EXCEPT `HarvestSource`. without correlation sqlalchemy could produce this sql command which would break.

```sql
SELECT count(harvest_source.id)
WHERE harvest_source.organization_id = organization.id
```

### organization view change
source count is now included which is an unintended addition but nice i think
Screenshot 2026-04-14 at 11 08 41 AM

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/database/models.py
+++ b/database/models.py
@@ -5,10 +5,10 @@ from typing import Optional
 
 from flask_sqlalchemy import SQLAlchemy
 from geoalchemy2 import Geometry
-from sqlalchemy import CheckConstraint, Column, Enum, String, func
+from sqlalchemy import CheckConstraint, Column, Enum, String, func, select
 from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.ext.mutable import MutableDict
-from sqlalchemy.orm import DeclarativeBase, backref
+from sqlalchemy.orm import DeclarativeBase, backref, column_property
 
 from shared.constants import (
     FREQUENCY_VALUES,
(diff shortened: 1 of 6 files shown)


database/models.py
```diff
@@ -115,6 +115,16 @@ class HarvestSource(db.Model):
     collection_parent_url = db.Column(db.String)
 
 
+# to avoid moving models around adding this here since it references
+# 2 models (Organization & HarvestSource)
+Organization.source_count = column_property(
```
neilmb, line 120: **comment (non-blocking)**: Just noting that we keep `models.py` in datagov-catalog in sync with this, so there will need to be the corresponding addition made there eventually.

tests/integration/app/test_routes.py
```diff
@@ -365,6 +365,26 @@ def test_json_builder_query(
...
             assert res.data.decode() == response
 
+    def test_organizations(self, client, interface_with_fixture_json):
+        """
+        checks the content of the json response when navigating to "/organizations/"
+        """
+        res = client.get("/organizations/")
+        assert res.status_code == 200
+
+        assert json.loads(res.data) == [
```
neilmb: **suggestion (non-blocking)**: The `res` object has a `json` property that does the `json.loads` for you:
```suggestion
        assert res.json == [
```

neilmb (dismissed):
Thanks, the column property seems like just the right tool for this.

neilmb (dismissed):

database/models.py
rshewitt: [PR](https://github.com/GSA/datagov-catalog/pull/229)

neilmb (approved):

PR #588, opened 2026-04-14
File Upload For DCAT Validator

cmhedrickREI:
# Pull Request

Related to gsa/data.gov#4638

- Adds a File upload method
- Limits the File upload size to 10MB ([based on this suggestion](https://github.com/GSA/datagov-harvester/pull/580#issuecomment-4207737056))

Live on Dev: https://datagov-harvest-dev.app.cloud.gov/validate/

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


app/templates/view_validators.html
```diff
@@ -104,15 +114,28 @@ <h2>Validation Results</h2>
...
+        upLoadField.style.display = "none"
+        jsonFile.value = ""
+    } else {
+        urlField.style.display = "none";
+        urlInput.value = "";
+        jsonField.style.display = "none";
+        jsonTextArea.value = "";
+        upLoadField.style.display = "block"
     }
 }
```
rshewitt: at this point we could probably consolidate this into something more concise since the logic is repeated but no biggie.

cmhedrickREI: I made a small update where I refactored the JS so there's less duplication of code, and keeping it more DRY.

neilmb (approved):
Looks good, I appreciate the Playwright tests for the client-side form-changing code.

PR #590, opened 2026-04-16
pin python version to 3.12.12

rshewitt:
# Pull Request

Interestingly related to [5773](https://github.com/GSA/data.gov/issues/5773)

## About
see [comment](https://github.com/GSA/data.gov/issues/5773#issuecomment-4261465406)

unless, for some reason, this introduces a breaking change with catalog i'm not going to do the same thing over there for now

it would be awesome if our python version could be a single source of truth to avoid having to edit so many files but there's a handful of things that need special handling. for example,
github action
```yml
- name: Read version file
    run: echo "PY_VERSION=$(cat .python-version)" >> $GITHUB_ENV
```
but for the cf runtime.txt, the dockerfile base image, and pyproject.tml (maybe this can inherit from `.python-version`?) it's hard to keep those synced without some script to edit it correctly

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


neilmb (approved):
Again, I'm surprised this changed so much in 3.12 versions, but this seems reasonable. Do we have any future idea of when we'll be updating these pinned versions?

cmhedrickREI (approved):
Very happy to us pinning the python version number :)

rshewitt:
> comment

we're at the mercy of what cf [supports](https://github.com/cloudfoundry/python-buildpack/releases). i suppose we could subscribe to the python buildpack releases and ping us on slack? otherwise we'll find out on build failure

FuhuXia:
pinning to patch version is ok, but it might not be the right fix for this mimetype issue.

rshewitt:
> pinning to patch version is ok, but it might not be the right fix for this mimetype issue.

ah yeah looking more into it looks like mimetypes depend on OS-level MIME data. for example,
mimetype calculation in the harvester docker app
```terminal
Python 3.12.12 (main, Feb 24 2026, 19:45:37) [GCC 14.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import mimetype
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'mimetype'
>>> import mimetypes
>>> u = "https://data.ngdc.noaa.gov/platforms/ocean/nos/coast/F00001-F02000/F00001/Smooth_Sheets/F00001.sid.gz"
>>> mimetypes.guess_type(u)
(None, 'gzip')
>>> 
```

mimetype calculation in my local terminal REPL
```terminal
Python 3.12.12 (main, Apr 16 2026, 11:14:48) [Clang 21.0.0 (clang-2100.0.123.102)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
Cmd click to launch VS Code Native REPL
>>> 
>>> u = "https://data.ngdc.noaa.gov/platforms/ocean/nos/coast/F00001-F02000/F00001/Smooth_Sheets/F00001.sid.gz"
>>> import mimetypes 
>>> mimetypes.guess_type(u)
('image/x-mrsid-image', 'gzip')
```

both use the same python version

PR #596, opened 2026-04-23
More logs

FuhuXia:
# Pull Request

Need more logs to debug openid login issue, and did a few more log related changes
- added log entries when user log in and out
- logs who did what change
- other logs to help debug

Sample log message:
```
Login initiated from ip=#.#.#.#
...
Login succeeded for user=test.user@gsa.gov ...
...
Logout completed for user=test.user@gsa.gov ...
...
Audit edit organization ... organization_slug=updated-slug path=/organization/edit/... user=test.user@gsa.gov
...
```


neilmb (approved):
These are on the verbose side now, particularly logging page rendering. Traffic to harvest probably won't be so high that we push too much info into New Relic, but we have had problems before with sending too much logging info. I'm fine with this for now, just noting that we might have to revisit it in the future if it gets too big for some limit somewhere.

FuhuXia:
> These are on the verbose side now, particularly logging page rendering. Traffic to harvest probably won't be so high that we push too much info into New Relic, but we have had problems before with sending too much logging info. I'm fine with this for now, just noting that we might have to revisit it in the future if it gets too big for some limit somewhere.

The harvest app generated two orders of magnitude fewer logs than the catalog app, it was almost sad to see the app has so few logs, therefore this ticket 🤣

PR #597, opened 2026-04-24
add harvest record URLs to OpenSearch documents

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5890

## About

/harvest_record/<record_id>/raw and /harvest_record/<record_id>/transformed are added to OS index in catalog app but not in harvest app. 

- add /harvest_record/<record_id>/raw
- add /harvest_record/<record_id>/transformed
- set catalog app url


neilmb (approved):
Looks right to me. Thanks for taking care of the customer issue.

PR #598, opened 2026-04-24
align harvester OpenSearch fields with catalog

FuhuXia:
# Pull Request

We did some OS mapping/index change on the catalog side. It is time to update harvest app.

- KEYWORD_NORMALIZER
- isPartOf
- publisher
- distribution_titles


jbrown-xentity (approved):
I'm approving this, but would prefer a test showing the normalization of keywords into lower case...

FuhuXia:
> I'm approving this, but would prefer a test showing the normalization of keywords into lower case...

test added in a new PR
https://github.com/GSA/datagov-harvester/pull/601

PR #602, opened 2026-04-28
use external route for openapi doc

FuhuXia:
# Pull Request

Related to
- https://gsa-tts.slack.com/archives/C2N85536E/p1777334340409169

## About
image

Use external route as the server address.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -28,6 +28,7 @@ htmx = None
 
 
 def _external_route_to_server_url(route: str | None) -> str | None:
+    """Return a normalized external server URL, or None for empty input."""
     if not route:
         return None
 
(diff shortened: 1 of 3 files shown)


app/__init__.py
```diff
@@ -26,6 +27,20 @@
 htmx = None
 
 
+def _external_route_to_server_url(route: str | None) -> str | None:
+    if not route:
+        return None
```
neilmb, line 33: **suggestion (non-blocking)**: I would want a docstring to explain that the method might return None or some actual answer.

neilmb (dismissed):
Great test case, thanks.

neilmb (approved):

PR #608, opened 2026-05-01
fix uncaught exception in munge spatial

rshewitt:
# Pull Request

Related to [API issue ticket](https://github.com/GSA/data.gov/issues/5883#issuecomment-4360403329)

## About
wraps code which threw uncaught exception

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


FuhuXia:
What scenario is this broad try/except meant to cover? Based on the current munge_spatial logic, the tested `"0,0,0,0"` falls into the duplicate-point branch and should return `{"type": "Point", "coordinates": [0, 0]}` rather than an empty string.

FuhuXia (approved):

PR #611, opened 2026-05-04
upgrade patched OpenSSL packages

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5938

## About

upgrade openssl to patch a CRITICAL issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,8 @@
 FROM python:3.12.12-slim
 
-# Update PAM packages to fix CVE-2025-6020 SNYK-DEBIAN12-PAM-10378969
-RUN apt-get install -y --only-upgrade libpam-modules libpam0g libpam-runtime
-
-# increase security by removing unnecessary packages
-RUN apt-get upgrade -y && \
+# Refresh APT indexes and pull the latest patched packages from the base distro.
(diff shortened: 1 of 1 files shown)


akf (commented):
I notice we lose this line, which probably wasn't intentional: 
```
RUN apt-get upgrade -y
```

Probably we want to keep that so any packages get upgraded. Might we not have been getting the updated packages because we weren't doing `apt-get update` before?

FuhuXia:
> I notice we lose this line, which probably wasn't intentional:
> 
> ```
> RUN apt-get upgrade -y
> ```
> 
> Probably we want to keep that so any packages get upgraded. Might we not have been getting the updated packages because we weren't doing `apt-get update` before?

You are right. let us keep it simple and do a broad upgrade, not worrying about individual pacakges.

akf (approved):
Bringing up the docker container locally, I can see the OpenSSL version is as desired: 

```
# openssl --version
OpenSSL 3.5.5 27 Jan 2026 (Library: OpenSSL 3.5.5 27 Jan 2026)
# apt info openssl
Package: openssl
Version: 3.5.5-1~deb13u2
Status: install ok installed
```

PR #612, opened 2026-05-05
no snyk test on every main push. remove snyk monitoring

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5912

## About
Reduce snyk tests. 
We have weekly snyk test and integrated test on each PR, no need for the snyk test and snyk monitoring on each main push.


akf (commented):
I'm still scratching my head over a few things. I think by "Integrated snyk test" you mean the tests that are triggered via webhook to `api.snyk.io`, right? 

So I agree that we should either turn this `snyk test` off, or the webhook, and I'm in favor of keeping the webhook running just because it is actually giving a "failure" status when it fails. (I didn't see any instances of this workflow finding a vulnerability though, so I might be wrong there.) 

I am unclear on whether we need to be running `snyk monitor` periodically. The [docs](https://docs.snyk.io/developer-tools/snyk-cli/commands/monitor) say it creates a project to be continuously monitored. Default is that the monitored project would be scanned daily, and we might have changed that in our settings. 

BUT! This doc also says "The `monitor` command is not supported for Snyk Code." Snyk Code is where we're hitting a usage limit, so I don't think the monitor commands are hurting us and we should keep them (but also, the scheduled job is probably sufficient). 

Since the scheduled job is part of this same workflow, I would keep the monitor code in here and just remove lines 8-10 (run on push to main), as you have. Unless you've determined that running `snyk monitor` is not needed at all ... 

Regarding the webhook, I see that it's currently set to run on "Pull requests" and "Pushes." Do we need the pushes (which I think is all branches), or are pull requests sufficient?

FuhuXia:
Our repos including this one are all public repos, with correct configuration, `Integrated PR check`, `snyk test` and `snyk monitor` should not count towards private test usages. But if configured wrong, or if snyk changes its counting mechanism on their end, or for some other weird reason, some of our tests are counted as private test usage and busted the usage limit for the last billing period.

This PR is trying to reduce snyk usage to bare minimal to get us out of the usage quote issue to unlock the failed PR check on all repos. We keep Integrated PR check, we run a weekly PR test. That will get us covered. Ideally we want to bring snyk-monitor-on-each-push-on-main back, but not until we are are sure it is configured correctly and wont count towards private test usage.

jbrown-xentity (approved):
LGTM.

@FuhuXia do you remember WHY we went to this manual snyk scan as a github action instead of the snyk monitoring available via Snyk itself? I'm trying to remember what we were trying to solve that the snyk tools on the website don't already give us...

FuhuXia:
> LGTM.
> 
> @FuhuXia do you remember WHY we went to this manual snyk scan as a github action instead of the snyk monitoring available via Snyk itself? I'm trying to remember what we were trying to solve that the snyk tools on the website don't already give us...

Yeah, I remember one time the Integrated snyk check on PR have an issue, it scans outdated snapshot and if it happens to be a vulnerable one, all PRs are blocked. We have to make it unrequired and rely on manual snyk test in GH action to get us covered.

PR #619, opened 2026-05-14
add another datetime regex for wafs

rshewitt:
# Pull Request

Related to [5982](https://github.com/GSA/data.gov/issues/5982)

## About
- add another datetime regex to waf datetime parsing. I also updated the split str which wasn't compatible with the files here https://edg.epa.gov/WAFer_harvest/ISO/. [source](https://harvest.data.gov/harvest_source/63703e7b-b96b-4918-bc73-c743435ed497) . i think splitting based on `.xml` makes sense.
- i also changed the permission of `to_be_filtered_by_datetime.xml` to read-only across the board but that won't be detected by git. i've noticed the file can be modified by some test and then break the datetime filtering test. hopefully whatever is responsible for modifying it doesn't break something?

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM

PR #620, opened 2026-05-15
Update dev instructions and readmes

akf:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/5876

## About

Updates the developer setup instructions. Some has been tested by a new team member (me), some has not, as noted. 

I decided to move most of these instructions out of the top-level README.md and into a separate document. I've attempted to make the new README.md sensible for members of the public as well as Data.gov team members. 

This PR removes the references to CKAN that were in the README and adds a link to the harvester wiki page. 

Adds the standard CONTRIBUTING.md and SECURITY.md

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


docs/developer.md
```diff
@@ -0,0 +1,218 @@
...
+5. Visit the site at `http://localhost:8080` and invoke the route which contains the code you've set the breakpoint on.
+
+
+## Testing
+
+### Install poetry
+
+We use `poetry` to manage this project, and to run the tests. Install poetry [here](https://python-poetry.org/docs/#installation). (Poetry is also installed and run automatically within the app container, which is why you didn't need it to get the app up and running.)
+
+Currently, this project is pinned to python version 3.12.12 (check `pyproject.toml` to verify). To install and use this specific version:
```
rshewitt: we're at 3.12.13 now since last week

docs/developer.md
```diff
@@ -0,0 +1,218 @@
...
+Once poetry is installed, `poetry install` installs dependencies into a local virtual environment.
+
+To update poetry itself locally (matching CI, which will always use the latest version), run `poetry self update` (or `make poetry-update`).
+
+
+### Running tests
+
+A number of "test" and "test-*" targets are defined in the `Makefile`.
+
+For tests to pass, you may have to pull the latest MDTranslator. Use `docker compose pull` to get the latest versions of the docker images.
```
rshewitt, line 91: thanks for adding this. failing transform tests because of an outdated image can be a big headache.

docs/developer.md
```diff
@@ -0,0 +1,218 @@
...
+#### User provided
+
+A user provided service by the name of `datagov-harvest-secrets` is also expected to be in place and populated with the following secrets:
+
+- CF_SERVICE_AUTH
+- CF_SERVICE_USER
+- FLASK_APP_SECRET_KEY
+- OPENID_PRIVATE_KEY
+
+### Manually Deploying the Flask Application to development
```
rshewitt, line 171: we try to discourage people from deploying via the command line and instead push to the `develop` branch which deploys to the development space for better team visibility of what changes are happening but this is still good documentation to have because there are circumstances where deploying from the command line is necessary (failing action step which prevents deployment) or quicker

docs/developer.md
akf: Actually it was already in there, I just moved it!

docs/developer.md
```diff
@@ -0,0 +1,218 @@
...
+    route_external: harvest-dev.data.gov
+    route_internal: datagov-harvest-dev.apps.internal
+    proxy_instances: 1
+    basic_auth_enabled: on
+    ```
+
+2. Deploy the application using Cloud Foundry's `cf push` command with the variable file:
+
+   ```bash
+   poetry export -f requirements.txt --output requirements.txt --without-hashes
```
rshewitt, line 189: exporting `requirements.txt` is an important step which i think deserves its own brief section prior to pushing the app to cf. do you mind adding that?

docs/developer.md
```diff
@@ -0,0 +1,218 @@
...
+- Validation
+  - DCAT-US
+    - `jsonschema` validation using draft 2020-12.
+- Load
+  - DCAT-US
+    - Conversion of dcat-us catalog into ckan dataset schema
+    - Create, delete, update, and patch of ckan package/dataset
+- Report
+  - Update harvest job db record with job results
+  - Email results using SMTP client
```
rshewitt: you can delete this

docs/developer.md
```diff
@@ -0,0 +1,218 @@
...
+
+
+#### user management
+
+`cf run-task datagov-harvest-admin --name "add new user" --command "flask user add xxx@gsa.gov --name xxx"`
```
rshewitt, line 208: do you mind adding a note that the data.gov team are the only intended users of the harvester admin app?

docs/developer.md
```diff
@@ -0,0 +1,218 @@
...
+
+Or, if doing for local development:
+
+`docker compose exec app flask user add your.i.name@gsa.gov --name yourName`
+
+#### add organizations
+
+`cf run-task datagov-harvest-admin --name "add new org" --command "flask org add 'Name of Org' --log https://some-url.png --id 1234"`
```
rshewitt, line 218: we primarily add organizations and harvest sources (not included here) via the UI but it's good for developers to be aware there's more than 1 way to achieve that.

README.md
```diff
@@ -1,235 +1,37 @@
 # datagov-harvester
 
-| Test Suite | Count | Coverage |
-| --- | --- | --- |
-| Unit | [![Unit Test Count](./tests/badges/unit/tests.svg)](./tests/unit)| [![Unit Test Coverage](./tests/badges/unit/coverage.svg)](./tests/unit)|
```
rshewitt, line 5: looks like this is currently broken so we need to fix it but we don't want to remove our test coverage.

docs/developer.md
akf: I'm taking this bit out -- probably not needed.

docs/developer.md
akf: I put it right after the "Running tests" section. It might need some more words.

rshewitt (approved):
looks good! thanks for this

PR #623, opened 2026-05-18
add missing/outdated datasets script & action

rshewitt:
# Pull Request

Related to [5946](https://github.com/GSA/data.gov/issues/5946)

## About
adds script & corresponding github action to check missing datasets

ran the script on dev. here's the log
`2026-05-20T10:55:44.98-0400 [APP/TASK/check missing datasets/0] OUT number of missing/outdated datasets 0`
i updated the python version to 3.12.13 because apparently 3.12.12 isn't supported anymore? i couldn't deploy on dev 
`2026-05-20T10:44:08.34-0400 [STG/0] OUT **ERROR** Could not install python: no match found for 3.12.12 in [3.10.20 3.11.15 3.12.13 3.13.13 3.14.4]`

i spoke with @FuhuXia and we're cool with just using a sample of missing/outdated records (we're the ones who are going to be reviewing it anyways)

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


.github/workflows/missing_outdated_datasets.yml
```diff
@@ -0,0 +1,70 @@
...
+            Please investigate harvest records / dataset sync.
+            `;
+
+            await github.rest.issues.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              title,
+              body,
+              labels: ["bug"]
+            });
```
jbrown-xentity: What if instead of reporting just the counts, it reported the actual records? It's often not trivial to be able to SSH into production DB to run these SQL statements manually...
Could the SQL not just get the counts, but actual records and spit them into a CSV? That way it would be easier to identify the source of the problem (particular harvest source, particular DCAT structure, etc).

jbrown-xentity (dismissed):
Good test, and this will be helpful. If it's firing a lot I think it would be better if we had the actual records to be able to debug/review directly, but this is a good first step in building confidence in the new system.

.github/workflows/missing_outdated_datasets.yml
rshewitt: the script needs to be run as a `cf run-task` command which makes cf log stdout parsing more complicated to get a csv archived in the action. i think the script should return 1 if the output is > 0 and 0 otherwise. i imagine the script could return a handful of records (although i'm not positive) to account for active harvesting which i would consider false positives. we can fine tune that but what i'm getting at is the script should almost never fail (create a ticket because we're missing datasets).

.github/workflows/missing_outdated_datasets.yml
```diff
@@ -0,0 +1,70 @@
+name: Check missing datasets
+
+on:
+  schedule:
+    - cron: "0 0 * * 0" # sunday at midnight EDT
```
FuhuXia: It does not really matter what time we run it, but this is not midnight EDT.

.github/workflows/missing_outdated_datasets.yml
FuhuXia: I agree that we need to output the 10 sample records, if found, for the O&M person to validate it. We dont want the poor guy go thru all the trouble then find out it is a false positive that the dataset is already created and indexed.

runtime.txt
```diff
@@ -1 +1 @@
-python-3.12.12
```
jbrown-xentity, line 1: Non-blocking, maybe future PR:
Can we use x to denote patch versions we don't care about? That way we'll stay on the latest? Seems like this is supported (https://docs.cloudfoundry.org/buildpacks/python/#runtime)...

jbrown-xentity (approved):
LGTM, thanks for that addition of sample records!

runtime.txt
rshewitt: i think it's good our local setup reflects the runtime environment as accurately as possible but patch versions shouldn't break anything so i'm cool with that

PR #624, opened 2026-05-19
Reset db

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5794

## About
Build a flask command to wipe the DB and rebuild it from migrations:

```bash
flask database reset --yes
```


jbrown-xentity (approved):
Dangerous command, but shouldn't be able to run this in production.

PR #625, opened 2026-05-19
API use X-API-Key, not Authorization

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5864

## About

When the harvest app is protected by Basic Auth, the Authorization request header is shared by both the API key and Basic Auth credentials. This creates conflicts and makes credential configuration and CloudFront origin policy settings more complicated (though still manageable). Renaming the API key header from Authorization to X-API-Key is the cleanest way to fix it.

### DIFF
--- a/app/auth.py
+++ b/app/auth.py
@@ -12,12 +12,13 @@ logger = logging.getLogger("harvest_admin.auth")
 
 
 class LoginRequiredAuth(APIKeyHeaderAuth):
+    API_KEY_HEADER_NAME = "X-API-Key"
 
     def get_security_scheme(self) -> dict[str, t.Any]:
         """Hardcode our particular security scheme."""
         security_scheme = {
             "type": "apiKey",
-            "name": "Authorization",
+            "name": self.API_KEY_HEADER_NAME,
             "in": "header",
         }
         return security_scheme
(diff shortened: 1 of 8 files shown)


app/auth.py
```diff
@@ -12,12 +12,13 @@
...
 
 class LoginRequiredAuth(APIKeyHeaderAuth):
+    API_KEY_HEADER_NAME = "X-API-Key"
 
     def get_security_scheme(self) -> dict[str, t.Any]:
         """Hardcode our particular security scheme."""
         security_scheme = {
             "type": "apiKey",
-            "name": "Authorization",
+            "name": self.API_KEY_HEADER_NAME,
```
rshewitt, line 21: i don't understand the need for this function (let alone the change here). this is supposed to generate a security definition in openapi/swagger documentation but we hide documentation for routes behind auth. maybe i'm not understanding something.

rshewitt: i'll resolve this. not a big deal.

scripts/ckan_migration/migrate_harvest_sources.py
```diff
@@ -337,13 +337,13 @@ def migrate_source(
...
     if harvester_api_token:
-        CONFIG.harvester_auth_headers = {"Authorization": harvester_api_token}
+        CONFIG.harvester_auth_headers = {"X-API-Key": harvester_api_token}
     else:
         logger.error("Please set a non-empty Harvester API token.")
         return 1
 
     if new_catalog_api_token:
-        CONFIG.new_catalog_auth_headers = {"Authorization": new_catalog_api_token}
+        CONFIG.new_catalog_auth_headers = {"X-API-Key": new_catalog_api_token}
```
rshewitt, line 346: we should really scrap this entire script but it seems like this is a search and replace sort of situation.

rshewitt (approved):
good to avoid auth header name collisions. also saw your email on adding `X-API-Key` to cache policy and forward to app via origin request policy

PR #647, opened 2026-06-01
patch snyk vulnerability

FuhuXia:
fix a few snyk finding

image

- Switch the app container from the Debian-based python:3.12.13-slim image to python:3.12.13-alpine3.22. The Debian image was introducing perl/perl-base@5.40.1-6, which Snyk reports as CVE-2026-42496 / SNYK-DEBIAN13-PERL-17037434. Alpine does not install Perl in this image, so the vulnerable Debian package is no longer present.
- pyjwt
- idna
- mako

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ flask-wtf = "^1.2.1"
 flask-migrate = "^4.0.7"
 bootstrap-flask = "^2.5.0"
 cloudfoundry-client = "^1.40.0"
-pyjwt = "^2.12.1"
+pyjwt = ">=2.13.0,<3.0.0"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ flask-wtf = "^1.2.1"
 flask-migrate = "^4.0.7"
 bootstrap-flask = "^2.5.0"
 cloudfoundry-client = "^1.40.0"
-pyjwt = "^2.12.1"
+pyjwt = ">=2.13.0,<3.0.0"
```
jbrown-xentity, line 31: It would probably be good to sort/order these in some recognizable way at some point...

jbrown-xentity (approved):
LGTM

PR #650, opened 2026-06-04
enable rate limit detection

FuhuXia:
We have added rate limit on cloudfront. this change will eanble nginx to capture those request and block them


rshewitt (approved):
header and value match up with what's in the google doc

PR #655, opened 2026-06-09
6035 - Adding missing urls for csp for gtag to work correctly

cody-seibert-gsa:
work for https://github.com/GSA/data.gov/issues/6035:

Adding the necessary https://developers.google.com/tag-platform/security/guides/csp#google_analytics_4 CSP records to not get CSP errors:


FuhuXia (approved):
LGTM

PR #656, opened 2026-06-09
6007 - Throw Error on Missing Env Vars

cody-seibert-gsa:
https://github.com/GSA/data.gov/issues/6007

Adding checks to throw an error if certain required env variables are not defined on start up.


FuhuXia (dismissed):
LGTM

nonblocking, we can add an extra protection by rejecting known null-like placeholders values
`{"null", "none", "nil", "undefined"}`

cody-seibert-gsa:
@FuhuXia good idea, I think I'll add that in now just for extra protection.

FuhuXia (approved):

cody-seibert-gsa:
deployed to develop and it seemed to work fine

PR #659, opened 2026-06-10
consider dataset geospatial with theme geospatial

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5949
- https://github.com/GSA/datagov-catalog/pull/302

## About

Synchronize with datagov-catalog on how to consider dataset geospatial


harvester/opensearch.py
```diff
@@ -287,8 +287,17 @@ def _walk(value: Any) -> None:
...
-            dataset.translated_spatial is not None
+        themes = dataset.dcat.get("theme") or []
+        if isinstance(themes, str):
+            themes = [themes]
+        has_spatial_theme = any(
+            isinstance(theme, str) and theme.strip().lower() == "geospatial"
+            for theme in themes
+        )
+        has_spatial = (
+            bool(spatial_value and str(spatial_value).strip())
```
rshewitt, line 298: why not just `str(spatial_value).strip()` for this first bool check?

tests/unit/test_opensearch.py
```diff
@@ -119,6 +119,41 @@ def test_dataset_to_document(sample_dataset):
...
+@pytest.mark.parametrize(
+    "theme",
+    [
+        ["Geospatial"],
+        ["GEOSPATIAL"],
+        ["Health", " geospatial "],
+        "Geospatial",
+    ],
+)
+def test_dataset_to_document_has_spatial_theme(sample_dataset, theme):
```
rshewitt, line 149: you're only determining whether it's spatial based on the theme in these tests. do you mind adding another which considers translated or spatial?

harvester/opensearch.py
rshewitt: oh i see if spatial_value is `None` then it won't work. nvm.

jbrown-xentity (dismissed):
It looks good to me, but having that other check would be good as Reid mentioned. Or maybe that test already exists?

FuhuXia:
Testing the other two scenario do not exist. I just added two more tests.

rshewitt (approved):
great, thanks for adding those additional tests

PR #660, opened 2026-06-11
5829 - adding hsts headers

cody-seibert-gsa:
work for https://github.com/GSA/data.gov/issues/5829

adding the hsts response headers


jbrown-xentity (approved):
LGTM

PR #662, opened 2026-06-12
move OpenSearch indexing command to harvester

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5733

## About

move OpenSearch sync actions to harvester with some improvements.

- get missing fixes on OS indexing from catalog
- keep OpenSearch `compare` command,  add `reset-mapping` command. no more `sync` command.
- add scheduled OpenSearch workflow 
- create issues when synchronization fails
- suppress verbose OpenSearch request logs
- normalize mappings during reset verification


jbrown-xentity (approved):
Compare function mostly taken from catalog: https://github.com/GSA/datagov-catalog/blob/aa2cfbd4bd04fdd73d0546e9970280f56685c0b1/app/commands.py#L306-L551. LGTM.

PR #672, opened 2026-06-15
isolate OpenSearch index writer

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5733

## About

Refactor to move the OpenSearch dataset index-writing implementation into a self-contained package owned by datagov-harvester.

  - Added shared/opensearch_index/ as the canonical index writer package.
  - Moved index settings, mappings, document transformation, bulk indexing, refresh, and deletion logic into
    writer.py.
  - Kept harvester/opensearch.py as a compatibility import so existing callers continue to work.
  - Added shared to the Poetry package configuration.
  - Added documentation explaining how catalog tests can vendor the package.
  - Added a dependency-boundary test preventing Flask, SQLAlchemy, or application-specific imports.
  - Updated OpenSearch unit tests to test the canonical package directly.
  
  This makes it easier for catalog app to use the index-writing code in its tests.


jbrown-xentity:
Overcome by events. Closing.

PR #676, opened 2026-06-16
5849 - updating github action to create an issue if the index failed

cody-seibert-gsa:
work for https://github.com/GSA/data.gov/issues/5849

Updating the existing monitor cf logs script to check for a passed in string of "failed to index in this batch", and if it is present, we create (or update) a github issue with that error so we know something went wrong.


FuhuXia (approved):
LGTM

PR #678, opened 2026-06-16
add db indexes

rshewitt:
# Pull Request

## About
- adds composite index to harvest record to improve `get_latest_harvest_records_by_source`
- adds index to harvest record error table by job id

i'm using `concurrently` for zero write downtime. if lock prevent the indexes from being created then i'll need to use the util `terminate_processes_sql`

i can update the catalog with these new indexes but it would be useless (not to mention adding space for the index) other than keeping them in sync. we may want to consider allowing them to deviate.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/database/models.py
+++ b/database/models.py
@@ -5,7 +5,7 @@ from typing import Optional
 
 from flask_sqlalchemy import SQLAlchemy
 from geoalchemy2 import Geometry
-from sqlalchemy import CheckConstraint, Column, Enum, String, func, select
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
LGTM

FuhuXia:
What do you mean by "adding space for the index" and "allowing them to deviate"?
> i can update the catalog with these new indexes but it would be useless (not to mention adding space for the index) other than keeping them in sync. we may want to consider allowing them to deviate.

PR #683, opened 2026-06-16
Add db indexes

rshewitt:
# Pull Request

## About
terminates existing processes to get index build to work. works on [develop](https://github.com/GSA/datagov-harvester/actions/runs/27651403743/job/81776021881). ignore lint. that's been fixed here.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


FuhuXia:
Can you explain why this migration needs to terminate existing DB sessions, or link to the issue/context that requires it?

the index operations are using `CONCURRENTLY`, so they should not require clearing active connections in the normal case. If we do need to terminate blockers, I think that should be an explicit operational step outside the Alembic migration rather than hidden inside one migration among many.

rshewitt:
> Can you explain why this migration needs to terminate existing DB sessions, or link to the issue/context that requires it?
> 
> the index operations are using `CONCURRENTLY`, so they should not require clearing active connections in the normal case. If we do need to terminate blockers, I think that should be an explicit operational step outside the Alembic migration rather than hidden inside one migration among many.

it's a brute-force solution we've implemented in the past to get things to work. i'll look into a less severe solution to avoid having to do this moving forward.

FuhuXia:
I see you moved sql termination command out of alembic upgrade and downgrade functions, and added it into `migrations.env.py`. Does it mean that every time the app is the restarted, all DB operation will terminated and transaction rolled back, with or without any actual migration? I can see a lot of problems with this, because the app is restarted every 15 mins.

FuhuXia (dismissed):
LGTM. This looks much better now, we only terminate sql connections when real migration change is detected.

For future enhancement:
- move this out of hidden migrations/env.py file and make a `flask db check-terminate` command before `flask db upgrade` command.
- selectively terminate bad/blocking connections, not all connections.

FuhuXia (approved):

PR #690, opened 2026-06-17
Add local development login

cody-seibert-gsa:
Adding in support for local login that by passes the need for login.gov so it's easier to test and run locally


jbrown-xentity (commented):
The reason we never implemented this is so that we wouldn't have to worry about any security holes.
The tests all look good, but what I don't see is confirmation of what `FLASK_ENV` is set to in dev, staging, and production environments in cloud.gov. I would like to see that posted here before moving forward.

FuhuXia:
`FLASK_ENV` is deprecated, so we should stop using it for any critical behavior.

I would prefer one clear setting that controls whether local login is disabled. That setting should be the only switch for this behavior. As a secondary safeguard, it should disabled when VCAP_APPLICATION is present, regardless what the setting says.

docker-compose.yml
```diff
@@ -72,6 +73,7 @@ services:
       CF_SERVICE_AUTH: ${CF_SERVICE_AUTH}
       FLASK_APP: run.py
       FLASK_ENV: development
+      ENABLE_LOCAL_DEV_LOGIN: "true"
```
FuhuXia: This value should only be read from .env, should not be set in the docker-compose.yml.

.env.sample
```diff
@@ -16,6 +16,9 @@ CATALOG_BASE_URL=http://localhost:8080
 FLASK_APP_SECRET_KEY=Px4o0MPEsCzHJRRSUDyt1VRi-lYF2W5VEzR13lylAgI # strictly for local dev
 HARVEST_API_TOKEN=local-dev-harvest-api-token # strictly for local dev
 
+# Local dev login (admin/admin). Only works when FLASK_ENV=development.
+ENABLE_LOCAL_DEV_LOGIN=true
```
FuhuXia: default to false

FuhuXia (changes requested):

FuhuXia (approved):
LGTM after the edits.

PR #691, opened 2026-06-17
Remove Bootstrap from admin UI

cody-seibert-gsa:
work for https://github.com/GSA/data.gov/issues/5744

Removing bootstrap library from harvester


akf:
`usa-table--borderless` would probably be a good addition to the classes on the tables that used to be borderless. I'm decidedly not a designer, but that's a thing that looked a little jarring (and you'd mentioned it) so I looked at the table examples in designsystem.digital.gov.

cody-seibert-gsa:
@akf thanks for pointing that out, I've refactored to use the borderless tables and it's redeployed to dev if you want to check it out again

akf:
@cody-seibert-gsa the borderless tables look good! 

I'm looking at more pages and I see that /metrics has some nearly-unreadable text that has the class "text-base-light". 

That's all that stood out to me. The "Add an organization" form looks a lot better with USWDS IMHO.

cody-seibert-gsa:
@akf I fixed the text contrast issues as well, thanks for pointing those out

app/static/gulpfile.js
```diff
@@ -59,9 +59,17 @@ uswds.paths.dist.theme = './_scss';
...
 exports.compile = uswds.compile;
-exports.watch = uswds.watch;
 exports.init = uswds.init;
 exports.copyAll = uswds.copyAll;
 exports.copyAssets = uswds.copyAssets;
 exports.updateUswds = uswds.updateUswds;
-// exports.default = uswds.watch;
+
+function watchDev() {
+    watch('./js/**/*.js', jsTask);
```
rshewitt, line 68: didn't know uswds watch didn't handle js so this is a good addition

app/templates/edit_data.html
```diff
@@ -9,24 +9,25 @@ <h2>{{action}} {{data_type}}</h2>
...
         {% endif %}
         {% if field.description %}
-            <small class="form-text text-muted">{{ field.description }}</small>
+            <span class="usa-hint">{{ field.description }}</span>
         {% endif %}
     </div>
     {% endfor %}
 
-    <button type="submit" class="btn btn-primary">{{button}}</button>
+    <button type="submit" class="usa-button">{{button}}</button>
```
rshewitt: can we wrap another `<div class="usa-form-group">` around this? the button ends up hugging the element above it

cody-seibert-gsa: addressed

rshewitt (approved):
the site definitely looks different and that's not something we want to get into a pattern of doing but the fact is that we're migrating from 2 different frameworks. had we not used bootstrap from the beginning we wouldn't be here. this also unifies the display with catalog.

PR #695, opened 2026-06-18
normalize DCAT values

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/6071

## About
To fix the found errors and potential errors from other fields:
- walks DCAT metadata containers and serializes any nested object/list values
- dcat.spatial uses a explicit top-level handling becasue it can be the structured value

### DIFF
--- a/harvester/opensearch.py
+++ b/harvester/opensearch.py
@@ -243,6 +243,12 @@ class OpenSearchInterface:
                     normalized_dcat[field] = value.isoformat()
                 elif value is not None and not isinstance(value, str):
                     normalized_dcat[field] = str(value)
+        spatial = normalized_dcat.get("spatial")
+        if spatial is not None and not isinstance(spatial, str):
+            if isinstance(spatial, (dict, list)):
+                normalized_dcat["spatial"] = json.dumps(spatial, sort_keys=True)
+            else:
+                normalized_dcat["spatial"] = str(spatial)
         return normalized_dcat
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


jbrown-xentity:
This might be a short term fix but not a good long term fix @FuhuXia .
DCAT-US1.1 allows for JSON, but only in GeoJSON format. See https://resources.data.gov/resources/dcat-us/#spatial.
The error you referred to shows a different format; looks like DCAT-US3.0: https://resources.data.gov/standards/catalog/dcat-us-3/temporal-spatial-metrics/#class-location-
I think the better short term fix is to hide the spatial field and/or munge it correctly (see https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/general_utils.py#L885-L1014). If it's not valid reject the value so it doesn't mess up openSearch, see https://github.com/GSA/datagov-harvester/blob/main/harvester/harvest.py#L1088-L1100 and https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/general_utils.py#L1014.

FuhuXia:
@jbrown-xentity 
This dataset's spatial value has gone through munging/translation and ended up with `None` as its `translated_spatial` value, so nothing is sent to OpenSearch for the `spatial_shape` geo field.

The failure is on the separate `dcat.spatial` metadata field. OpenSearch maps `dcat.spatial` as text, but this source value is a JSON object, so OpenSearch rejects it.

jbrown-xentity (dismissed):
Stringifying the object is the best option right now.

jbrown-xentity:
This isn't really valid 1.1, but it is technically valid, so approving for now.

FuhuXia:
preview in staging, error is gone.

image

jbrown-xentity (approved):
LGTM

PR #696, opened 2026-06-18
remove ckan sync content

rshewitt:
# Pull Request

## About
removes needless count of latest set of records synced with ckan

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
This looks good to me. I wish we had a more clear feedback loop on individual items synced and searchable in catalog; however bulk updating is very convenient and useful.

PR #705, opened 2026-06-22
keep publisher as  object

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/6078
- https://github.com/GSA/datagov-harvester/pull/695

## About

last PR serialized any nested object/list values except `spatial`. it turns out `publisher` is another value that need to be preserved.


jbrown-xentity (approved):
As long as the name is there, it should show up ok on the website. LGTM

PR #708, opened 2026-06-22
5999 - adding dcat 3 support for validation and source

cody-seibert-gsa:
work for https://github.com/GSA/data.gov/issues/5999

## Summary

Adds DCAT-US 3.0 as a supported harvest source schema type, including per-dataset JSON Schema validation against the 3.0 definitions and a database migration to allow `dcatus3.0` sources.

Introduces a `create_task_handler()` abstraction so harvest jobs can run locally via `LocalTaskHandler` when Cloud Foundry is not configured, while preserving existing CF task behavior in deployed environments.

deployed to https://harvest-dev.data.gov/


rshewitt:
[5999](https://github.com/GSA/data.gov/issues/5999) doesn't involve changing anything with the load or task manager.  can you revert all those changes?

harvester/utils/general_utils.py
```diff
@@ -1381,9 +1381,17 @@ def is_valid_uuid4(uuid_string) -> bool:
         return False
 
 
-def build_dcatus3_validator(definitions_dir):
+def build_dcatus3_validator(
+    definitions_dir,
+    root_ref="https://resources.data.gov/dcat-us/3.0.0/definitions/catalog",
```
rshewitt, line 1414: @jbrown-xentity do we know what we need from the catalog? or should this ticket just involve processing datasets like dcatus v1.1?

harvester/harvest.py
```diff
@@ -115,6 +112,14 @@ def __post_init__(self) -> None:
             self.schema_file = (
                 self.schemas_root / "dcatus1.1" / "non-federal_dataset.json"
             )
+        elif self.schema_type == "dcatus3.0":
+            # dcatus3.0 has a single dataset schema (no federal/non-federal split).
+            # NOTE: a 3.0 dataset identifier may be a string or an object; the
+            # harvester uses it as a dict key / db identifier, so only string
+            # identifiers are supported. object identifiers are out of scope.
+            self.schema_file = (
```
rshewitt, line 125: i think we want to support identifiers as an object too in this PR. @jbrown-xentity thoughts?

harvester/utils/general_utils.py
rshewitt: after reviewing the ticket it seems like this PR is just supposed to handle datasets not catalogs

tests/unit/test_dcatus3_validator.py
```diff
@@ -0,0 +1,42 @@
+from pathlib import Path
+
+from harvester.utils.general_utils import build_dcatus3_validator
+
+ROOT_DIR = Path(__file__).parents[2]
+DCATUS3_DEFINITIONS = ROOT_DIR / "schemas" / "dcatus3.0" / "definitions"
+
+DATASET_REF = "https://resources.data.gov/dcat-us/3.0.0/definitions/dataset"
+
+VALID_DATASET = {
```
rshewitt: do you mind putting this in `conftest`? i know there's examples of fixtures existing in places they shouldn't (i'm probably guilty of that) but i'd like to keep it under better watch

tests/unit/test_dcatus3_validator.py
```diff
@@ -0,0 +1,42 @@
...
+class TestBuildDcatus3Validator:
+    def test_dataset_root_ref_validates_single_dataset(self):
+        """With the dataset root ref, a single dataset dict validates standalone."""
+        validator = build_dcatus3_validator(DCATUS3_DEFINITIONS, root_ref=DATASET_REF)
+        assert validator.is_valid(VALID_DATASET)
+
+    def test_dataset_root_ref_flags_missing_required_field(self):
+        """A dataset missing the mandatory contactPoint produces errors."""
+        validator = build_dcatus3_validator(DCATUS3_DEFINITIONS, root_ref=DATASET_REF)
+        invalid = {k: v for k, v in VALID_DATASET.items() if k != "contactPoint"}
```
rshewitt: why not just?
```python
del VALID_DATASET["contactPoint"]
```

rshewitt: oh right if you did that then it would mutate it for other tests but not if you used it as a fixture

example_data/dcatus/dcatus3_0.json
```diff
@@ -0,0 +1,42 @@
...
+      "publisher": {
+        "@type": "Organization",
+        "name": "Test Agency"
+      },
+      "contactPoint": {
+        "@type": "Kind",
+        "fn": "Test Contact One",
+        "hasEmail": "mailto:one@example.gov"
+      }
+    },
```
jbrown-xentity, line 25: This is only a minimally required example. I would suggest trying other complete records, see something like https://github.com/GSA/dcat-us/blob/main/jsonschema/examples/Dataset/good/complete_example.json

harvester/lib/task_handler.py
```diff
@@ -0,0 +1,32 @@
...
+
+    if on_cloud_foundry or cf_api_url:
+        try:
+            return CFHandler(cf_api_url, cf_service_user, cf_service_auth)
+        except Exception as e:
+            if on_cloud_foundry:
+                raise
+            logger.warning("CFHandler init failed (%s); using LocalTaskHandler.", e)
+
+    logger.warning("Cloud Foundry task API is not configured; using LocalTaskHandler.")
```
jbrown-xentity: Can we setup a New Relic Alert for this to make sure we would be notified if this occurred?

harvester/harvest.py
```diff
@@ -115,6 +112,14 @@ def __post_init__(self) -> None:
             self.schema_file = (
                 self.schemas_root / "dcatus1.1" / "non-federal_dataset.json"
             )
+        elif self.schema_type == "dcatus3.0":
+            # dcatus3.0 has a single dataset schema (no federal/non-federal split).
+            # NOTE: a 3.0 dataset identifier may be a string or an object; the
+            # harvester uses it as a dict key / db identifier, so only string
+            # identifiers are supported. object identifiers are out of scope.
```
jbrown-xentity: Object identifiers are not out of scope. If object, use `@id` field. See https://resources.data.gov/standards/catalog/dcat-us-3/identifiers-and-relationships/#identifier--anyOf_i1

jbrown-xentity (commented):
In general I understand needing a local task manager, however I think it needs separate consideration before adding to this. If we want to consider it a requirement (which I'm open to considering having that setup and working as a requirement for this work), then let's separate out that PR and make that blocking this. Please consider working with @FuhuXia or @rshewitt on setting up New Relic alert to make sure we never hit the misconfigured use case...

As to the development, just make sure that various full cases are utilized and not just the minimal use case. We absolutely have to handle `identifier` being an object, not just a string.

harvester/lib/task_handler.py
cody-seibert-gsa: I've pulled the local task handler changes into a separate branch.

harvester/harvest.py
cody-seibert-gsa: I added logic to grab the identifier directly if it's a string; otherwise, it uses the identifier.id normalize_dataset_identifier https://github.com/GSA/datagov-harvester/pull/708/changes#diff-332a4b024a46ac5c9b23740ee07ef353316e534e851f894eab4e4b8559aa558fR464

harvester/harvest.py
cody-seibert-gsa: https://github.com/GSA/datagov-harvester/pull/708/changes#diff-332a4b024a46ac5c9b23740ee07ef353316e534e851f894eab4e4b8559aa558fR464 

updated the logic to handle both scenarios

example_data/dcatus/dcatus3_0.json
cody-seibert-gsa: I pulled in that complete_example for another test

tests/unit/test_dcatus3_validator.py
```diff
@@ -0,0 +1,53 @@
...
+
+from harvester.utils.general_utils import (
+    build_dcatus3_validator,
+    normalize_dataset_identifier,
+    open_json,
+)
+
+ROOT_DIR = Path(__file__).parents[2]
+DCATUS3_DEFINITIONS = ROOT_DIR / "schemas" / "dcatus3.0" / "definitions"
+DCATUS3_COMPLETE_EXAMPLE = (
```
rshewitt, line 13: this is a fixture so I would expect it in `conftest.py`

tests/unit/test_dcatus3_validator.py
```diff
@@ -0,0 +1,53 @@
...
+    / "complete_example.json"
+)
+
+DATASET_REF = "https://resources.data.gov/dcat-us/3.0.0/definitions/dataset"
+
+
+class TestBuildDcatus3Validator:
+    def test_dataset_root_ref_validates_single_dataset(self, valid_dcatus3_dataset):
+        """With the dataset root ref, a single dataset dict validates standalone."""
+        validator = build_dcatus3_validator(DCATUS3_DEFINITIONS, root_ref=DATASET_REF)
```
rshewitt: you could consider creating the validator as a global at the top of the module to avoid repeating it 3 times. you'd obviously overwrite it in the catalog test. your call. feel free to dismiss.

tests/unit/test_dcatus3_validator.py
rshewitt: i do see how `conftest` is getting to be quite big so putting a fixture with the tests which use it is valuable. I can see making the exception and doing it this way. it's your call.

rshewitt (dismissed):
looks good. we're making good progress on dcatus 3

tests/unit/test_dcatus3_validator.py
cody-seibert-gsa: yeah I think co-locating test data with the test makes it easier to maintain potentially

tests/unit/test_dcatus3_validator.py
cody-seibert-gsa: good suggestions, I will clean it up

harvester/opensearch.py
rshewitt, line 1: there's a separate [ticket](https://github.com/GSA/data.gov/issues/6001) for work on opensearch. i understand the overlap and the idea of "the car is raised I might as well change the oil while i'm at" but we need to stick to the defined scope of work for a ticket. if you want to stash it and pick that ticket up next that's cool.

rshewitt: but if things were to break without making these changes then i understand

cody-seibert-gsa: sure thing, I can do that

rshewitt (approved):
thanks for making those changes. looks good!

PR #709, opened 2026-06-23
5992 change default waf datetime to now

jaredb96:
# Pull Request

Work for [Issue #5992](https://github.com/GSA/data.gov/issues/5992)

When traversing waf directories, we want to replace the default time with "now".

### DIFF
--- a/harvester/utils/general_utils.py
+++ b/harvester/utils/general_utils.py
@@ -37,7 +37,7 @@ FREQUENCY_ENUM = {"daily": 1, "weekly": 7, "biweekly": 14, "monthly": 30}
 # User-Agent header for all HTTP requests
 USER_AGENT = "HarvesterBot/0.0 (https://data.gov; datagovhelp@gsa.gov) Data.gov/2.0"
 
-DT_PLACEHOLDER = datetime(1900, 1, 1, 0, 0)
+DT_PLACEHOLDER = datetime.now()
 
 PACKAGE_NAME_MAX_LENGTH = 90
 PACKAGE_NAME_MIN_LENGTH = 2
--- a/tests/integration/harvest/test_extract.py
+++ b/tests/integration/harvest/test_extract.py
@@ -1,4 +1,4 @@
-from harvester.harvest import HarvestSource
+from harvester.harvest import DT_PLACEHOLDER, HarvestSource
 from harvester.utils.general_utils import traverse_waf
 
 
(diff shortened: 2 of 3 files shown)


harvester/utils/general_utils.py
```diff
@@ -37,7 +37,6 @@
 # User-Agent header for all HTTP requests
 USER_AGENT = "HarvesterBot/0.0 (https://data.gov; datagovhelp@gsa.gov) Data.gov/2.0"
 
-DT_PLACEHOLDER = datetime(1900, 1, 1, 0, 0)
```
rshewitt, line 40: i was anticipating you'd just change this value instead of removing it entirely. like...
```python
DT_PLACEHOLDER = datetime.now()
```

tests/integration/harvest/test_extract.py
```diff
@@ -63,3 +65,32 @@ def test_extract_source_with_dataset_missing_identifier(
...
+        after_time = datetime.now()
+
+        parent_record = harvest_source.external_records[0]
+        assert (
+            parent_record["identifier"]
+            == source_data_waf_collection["collection_parent_url"]
+        )
+
+        # check the timestamp is now
+        parent_dt = parent_record["modified_date"]
```
rshewitt: we could probably check the exact value since harvester would import `DT_PLACEHOLDER` which  would execute `datetime.now()` so from this file you might be able to do something like?
```python
from harvester.harvest import DT_PLACEHOLDER

parent_record["modified_date"] == DT_PLACEHOLDER
```

tests/unit/test_utils.py
```diff
@@ -236,6 +236,33 @@ def test_get_waf_datetimes(self):
...
+
+        page_html = """<html><body><pre>
+          <a href="file1.xml">file1.xml</a>   12K  
+          <a href="file2.xml">file2.xml</a>   12K  
+          </pre></body></html>"""
+
+        before_time = datetime.now()
+
+        soup = BeautifulSoup(page_html)
+        datetimes = get_waf_datetimes(soup, 2)
```
rshewitt, line 249: similar comment as above you should be able to import `DT_PLACEHOLDER` from general utils and assert the exact value instead of whether it's in between

harvester/utils/general_utils.py
jaredb96: I was also anticipating doing that, however I was a bit worried that that solution would give multiple files an a stale now value that's set once when this line is evaluated. But if there's nothing wrong with that I'm happy to revert this back to using the constant.

tests/integration/harvest/test_extract.py
jaredb96: Yup, okay so I'll return the constant and simplify this a bit, ty

tests/unit/test_utils.py
jaredb96: Will do!

harvester/utils/general_utils.py
rshewitt: using the constant should be fine

jaredb96:
Addressed your comments @rshewitt

rshewitt (dismissed):
awesome thanks for making those updates

rshewitt (approved):
looks good!

PR #710, opened 2026-06-24
Return clear error when harvest source URL already exists

cody-seibert-gsa:
work for [#6102 ](https://github.com/GSA/data.gov/issues/6102)

When trying to add a duplicate harvest source url, the UI would just throw an internal server error.  Now, it will flash an error alert at the top so it's obvious what went wrong for users.
Screenshot 2026-06-24 at 9 42 19 AM


jbrown-xentity (approved):
LGTM. Saves us from ourselves.

PR #715, opened 2026-06-24
Improve missing dataset check

FuhuXia:
## About

- include staging on the sheduled run
- correct issue owner
- the issue is created only on scheduled run
- rename `get_missing_or_outdated_datasets` since we only get 10 samples, not all
- add `10 sample record ids for troubleshooting` otherwise no idea what the 10 UUID in the log are for


rshewitt (approved):
good additions

PR #716, opened 2026-06-24
Add LocalTaskHandler to run harvest jobs locally without Cloud Foundry

cody-seibert-gsa:
https://github.com/GSA/data.gov/issues/6140

To allow a developer to easily test harvest sources locally by clicking the harvest button, this PR adds a LocalTaskHandler which runs the exact same script which the CFHandler runs without needing CF credentials setup.  This will allow for a more robust e2e setup locally or in github actions so we can actually verify the functionality all works e2e.


harvester/lib/local_task_handler.py
```diff
@@ -0,0 +1,109 @@
...
+    Task state lives in a class-level registry so that it survives the
+    short-lived handler instances that LoadManager creates. Note that harvest
+    jobs run as separate subprocesses, so a handler created inside a harvest
+    subprocess starts with an empty registry; that is fine because the
+    duplicate-job guard in LoadManager.start_job is enforced at the database
+    level, not via this registry.
+    """
+
+    _tasks: dict[str, dict] = {}
+    _lock = threading.Lock()
```
rshewitt, line 37: non-blocking just a comment. this seems like the safe thing to do to avoid any potential race conditions

harvester/lib/local_task_handler.py
```diff
@@ -0,0 +1,109 @@
...
+                "updated_at": datetime.now(timezone.utc)
+                .isoformat()
+                .replace("+00:00", "Z"),
+            }
+            self._tasks[task_id] = task
+            return task
+
+    def stop_task(self, task_id):
+        with self._lock:
+            for task in self._tasks.values():
```
rshewitt: can we just try to get by key name instead of looping through them?

rshewitt:
confirming the local task handler is used when running jobs locally without credentials

Screenshot 2026-07-01 at 1 01 19 PM

rshewitt (dismissed):
good addition. approving. my comment about the loop is non-blocking (just a suggestion)

harvester/lib/local_task_handler.py
cody-seibert-gsa: fixed

rshewitt (approved):
good addition. thanks for doing this!

PR #718, opened 2026-06-26
6001 - Transition OpenSearch Indexes to use DCAT 3.0 structure

cody-seibert-gsa:
initial work on https://github.com/GSA/data.gov/issues/6001 as I accidently did it on another story thinking it was a bug.

Note this also may have some overlap with https://github.com/GSA/data.gov/issues/6002 as since we updated the index schema on opensearch we have to update catalog to be able to properly search on it.

should be merged together with https://github.com/GSA/datagov-catalog/pull/338

## Summary

This PR normalizes DCAT-US metadata before indexing it in OpenSearch so DCAT-US 1.1 and DCAT-US 3.0 records can coexist in the same index.

DCAT-US 3.0 changes the shape of several fields that are indexed for search. For example, `theme` changes from strings to Concept objects, while `identifier` and dataset series membership still need to preserve the legacy search contracts used by catalog today.

Because OpenSearch cannot map one field as multiple data types, indexing both schema versions without normalization would either fail or produce inconsistent search behavior. This branch converts indexed DCAT fields into consistent shapes at index time while preserving the original `dcat` payload for reference.

### Added DCAT Index Transformation

Added `harvester/opensearch_transform.py`, which introduces `DcatIndexTransformer`, a pure transformation module with no I/O. It maps DCAT metadata into canonical OpenSearch field shapes.

The transformer normalizes indexed DCAT fields:

- `identifier` stays a scalar string; DCAT-US 3.0 Identifier objects are reduced to `@id`
- `theme` values are normalized to lists of Concept objects with fields such as `prefLabel`, `@id`, and `altLabel`
- `publisher`, `keyword`, `title`, `description`, and `distribution_titles` preserve their existing indexed contracts

### Updated OpenSearch Indexing

Updated `harvester/opensearch.py` so document building and mappings use the normalized field shapes:

- `theme` is now an object mapping with structured subfields
- `identifier` remains the legacy scalar text mapping
- Text and keyword subfields are included where needed for search, faceting, and exact matching
- `dataset_to_document()` now uses `DcatIndexTransformer` for derived search fields
- The full `dcat` blob is still stored for reference
- DCAT-US 3.0 `inSeries` can populate the legacy `dcat.isPartOf` alias when `isPartOf` is absent
- `has_spatial` also checks for a `geospatial` theme using `concept_labels()`

## Field Mapping Summary

| OpenSearch field | DCAT-US 1.1 source | DCAT-US 3.0 source | Indexed shape |
| --- | --- | --- | --- |
| `identifier` | string | Identifier object | string from the original value or object `@id` |
| `theme` | list of strings | list of Concept objects | list of Concept objects with `prefLabel`, `@id`, `altLabel`, etc. |
| `publisher` | organization name string | organization name string | string |
| `keyword` | list of strings | list of strings | list of strings |

## Documentation

Added `docs/dcat3-opensearch-flow.html`, a visual flow diagram showing the path from harvest, through DCAT transformation, to OpenSearch indexing.

## Out Of Scope

Fields stored only inside the free-form nested `dcat` blob, such as `contactPoint` and `accessRights`, are not transformed by `DcatIndexTransformer`. They continue to be flattened to searchable text in the nested `dcat` document.


harvester/opensearch.py
```diff
@@ -237,6 +249,53 @@ def _serialize_dcat_value(value: Any) -> str:
             return json.dumps(value, sort_keys=True)
         return str(value)
 
+    @staticmethod
+    def _normalize_dataset_identifier(identifier: Any) -> str | None:
```
rshewitt: do you mind providing some documentation here? Really just to indicate that "identifier" within a dcatus 3.0 dataset object comes from 2 different places

rshewitt: also mentioning that this handles both dcatus 1.0 & 3.0

harvester/opensearch.py
```diff
@@ -237,6 +249,53 @@ def _serialize_dcat_value(value: Any) -> str:
...
+    def _dcat_text_from_object(cls, value: dict, field: str) -> str:
+        if field == "identifier":
+            identifier = cls._normalize_dataset_identifier(value)
+            if identifier:
+                return identifier
+
+        preferred_keys = (
+            ("prefLabel", "title", "name", "label", "notation", "@id")
+            if field in {"accessRights", "status", "theme"}
+            else ("@id", "title", "name", "prefLabel", "label", "notation")
```
rshewitt: are these ordered in any particular way? if so @jbrown-xentity do you mind looking at this?

rshewitt: also is this exhaustive or are there other fields in object not listed here? i'm not a dcatus 3.0 expert

jbrown-xentity: I think transforming these objects into singular strings is the wrong long-term approach, because it strips out information and removes searchability.

There are some other possibilities we should consider:

- Could we define 3.0 as a new type of object, and provide mappings for searchability for key fields such as `keyword`?
- Could we map 1.1 to 3.0 and explode out the text fields into objects with fields, so as to hold all the complete metadata?
- Could we just stringify the JSON into the singular field so it is all searchable? What ramifications could that have to views (such as landing page or theme)?

I would review the code @cody-seibert-gsa @rshewitt and see which of these 3 options makes the most sense, or if there's an alternative I'm missing.

rshewitt: i agree we don't want to reduce the information from the source. As a first-pass could we not just put it in like it's a dcatus 1.1 document? Parse out all the individual fields as normal and store the dcatus 3.0 dict in the `dcat` open search doc? the `dcat` document wouldn't have objects reduced to strings. it would look almost identical to the source (except the dates normalized and the spatial JSON serialized)

cody-seibert-gsa: after our meeting we discussed just making dcat 3 schema the one we use in opensource to try and allow a more flexible upgrade path, but we will not make everything searchable, as the current set of searchable terms was meticulously deteremined.  The pr now transforms dcat 1 data into the dcat 3 format before storing into opensearch.  in the future, supporting new fields should be easy to extend.

harvester/opensearch.py
cody-seibert-gsa: I'd recommend re-reviewing as we pivoted the approach after our meeting

harvester/opensearch.py
```diff
@@ -237,6 +326,53 @@ def _serialize_dcat_value(value: Any) -> str:
...
+    def _normalize_dataset_identifier(identifier: Any) -> str | None:
+        if identifier is None:
+            return None
+        if isinstance(identifier, str):
+            return identifier if identifier.strip() else None
+        if isinstance(identifier, dict):
+            value = identifier.get("@id")
+            if isinstance(value, str) and value.strip():
+                return value
+        return None
```
jbrown-xentity, line 284: @cody-seibert-gsa I'm confused on this. The above opensearch definition looks like we are standardizing on DCAT3 and the object model for identifier. This function seems to do the opposite, where we are taking an object and simplifying to the id.

Which is getting used, and why? Or how am I misunderstanding how the pieces fit together?

harvester/opensearch.py
```diff
@@ -100,15 +119,85 @@ class OpenSearchInterface:
...
+                                "normalizer": KEYWORD_NORMALIZER,
+                            }
+                        },
+                    },
+                    "description": {
+                        "type": "text",
+                        "analyzer": TEXT_ANALYZER,
+                        "search_analyzer": TEXT_ANALYZER,
+                    },
+                },
```
jbrown-xentity: This one I know I'm going back on what I said moving to 3.0, however this one is weird. A datasetSeries is essentially it's own dataset, with it's own references and information. I would keep this out of search for now. I think at the harvester level we may parse this record as it's own dataset and link them back together in a more unique way...

jbrown-xentity (commented):
@cody-seibert-gsa after seeing this, I'm not sure how much of the functionality we want to move forward for DCAT-US3.
Identifier for example seems like maybe we just keep the old model; I don't see much value in adding the extra stuff. Same with inSeries.
But the theme does seem valuable, some of that extra info may become valuable for hierarchy and other cross-searching between sources.
Let's talk about this tomorrow as a team if possible.

harvester/opensearch_transform.py
```diff
@@ -0,0 +1,341 @@
...
+Fields that live solely inside the free-form ``dcat`` blob (e.g.
+``contactPoint``, ``accessRights``) are deliberately out of scope.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any, Callable, Iterable
+
+__all__ = [
```
akuny, line 35: Based on current usage, we could trim the public API for this module to:

```
__all__ = [
    "DcatIndexTransformer",
    "concept_labels",
    "distribution_titles",
]
```

From what I can tell, those are the pieces of functionality from this module that are used in actual application code.

However, now that I'm thinking about it, the design of this module does imply that somewhere down the road application logic will be configuring and wiring up a FieldSpec with callables, whereas now we're just doing:

```
DCAT_INDEX_TRANSFORMER = DcatIndexTransformer()
```

If we're anticipating a concrete use case for the configurable nature of this module then disregard, but if not I think we could make the machinery simpler here: say, a dict literal calling the coercion helpers directly.

Not a blocker by any means, but wanted to surface these thoughts for the record.

harvester/opensearch.py
akuny: If I'm reading this right, the document sent to OpenSearch contains the identifier in two places:

1. the top-level `identifier` field: string to object
2. `identifier` inside the stored `dcat` blob, where everything is flattened from object to string

So both are used, just on different fields: strings go up for the object-mapped search field, objects come down for the text-only blob. If that's right, renaming this to something like `_identifier_text_for_dcat_blob` might make the split clearer.

cody-seibert-gsa: _normalize_dataset_identifier only flattened dcat.identifier inside the dcat blob; catalog search uses the top-level identifier field instead. My confusion was around the blob mapping: because dcat was dynamic, preserving the source shape could let OpenSearch dynamically map dcat.identifier as either a DCAT 1 string or a DCAT 3 object, depending on which record indexed first. That was the reason I originally flattened it. I’ve removed that helper and set dcat.dynamic = false, so the blob can keep the Postgres/DCAT shape in _source  without OpenSearch trying to infer mappings for mixed identifier structures.

harvester/opensearch.py
cody-seibert-gsa: sure I can remove this field, but are you saying that we should normalize dcat3 inSeries[0] into the previous inPartOf field?

harvester/opensearch_transform.py
cody-seibert-gsa: good suggestion, I reduced what we export from this module

PR #723, opened 2026-06-29
DB optimizing for slow pages

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/5957

## About
- replace index `ix_hr_source_status_identifier_created_desc` with a more targeted index `ix_harvest_record_source_identifier_created_success`. 
- match model constraints with existing database schema so that "flask db check" does not gives error
- refactor the harvest job detail error query so the expensive grouping and pagination happen before joining harvest_record
- avoid creating a temp file with whole csv on disk. streaming the csv export instead.
- add missing indexes. 

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


FuhuXia:
This is deployed on staging. If it is overwritten by other merged PRs, I can redeploy it for the reviewer who want to compare the performance difference between prod and staging. Because some page will simply crash the prod site, I wont post the URLs here.

app/routes.py
```diff
@@ -1412,90 +1412,41 @@ def download_harvest_errors_by_job(job_id, error_type):
...
+                            else ""
+                        ),
+                        str(error["type"]) if error["type"] else "",
+                        str(error["message"]) if error["message"] else "",
+                        str(error["date_created"]) if error["date_created"] else "",
+                    ]
+                    yield csv_writer.writerow(row)
+
+        return Response(
+            stream_with_context(generate_csv()),
```
rshewitt: doesn't look like you're accessing anything from the request context (e.g. session, request.arg, g) so is `stream_with_context` necessary?

FuhuXia: We dont need to access any session request... we need the flask-SQLAlchemy scoped session alive during streaming the response. I have added comments to the line.

rshewitt (dismissed):
been looking forward to seeing these improvements!

rshewitt (approved):

PR #724, opened 2026-06-29
tell bot to ignore babel major version

FuhuXia:
babelify@10 requires @babel/core 7. We stay with 7 unless there is a security update demanding 8.


rshewitt (approved):
LGTM

PR #729, opened 2026-06-29
Add error when JSON file > 10mb.

### LINKED ISSUE #6081: harvester validator doesn't enforce file size for fetch via url
validator accepts fetching a dcatus document via url but should raise an exception when the file size of document exceeds 10mb. this [source](https://data.ca.gov/data.json) which is ~20mb can be fetched no problem. that's not the intended behavior.

## How to reproduce
use the source above in the validator and fetch it via url

## Expected behavior
error to be displayed on the page indicating the file is too big

## Actual behavior
file is fetched and processed normally

## Sketch

jaredb96:
Work for Issue [#6081](https://github.com/GSA/data.gov/issues/6081)


app/util.py
rshewitt, line 1: can we include a test verifying the intended behavior?

jaredb96: Sure thing

jaredb96: Added unit tests for this

rshewitt (dismissed):
fixes the issue thanks for doing this. a couple improvements (non-blocking but if you wanted to do it)
- check the `content-length` of the file (if the server sends it) and throw the `ValueError` if > 10mb
- stream in chunks and stop downloading if download size exceeds 10mb. currently we download the whole thing and THEN check if it's bigger than 10mb 
- improve the value error by stating what our limit is (10mb)

jaredb96:
> fixes the issue thanks for doing this. a couple improvements (non-blocking but if you wanted to do it)
> 
> * check the `content-length` of the file (if the server sends it) and throw the `ValueError` if > 10mb
> * stream in chunks and stop downloading if download size exceeds 10mb. currently we download the whole thing and THEN check if it's bigger than 10mb
> * improve the value error by stating what our limit is (10mb)

Sure, I'll implement these suggestions, ty!

rshewitt (approved):
looks good. thanks for those additions!

PR #732, opened 2026-06-30
Set status of harvest record to successful after dataset change.

### LINKED ISSUE #5935: set harvest record "success" status on successful dataset change
## User Story

In order to provide better meaning to what a "successful" harvest record is, datagov wants to set that status after the dataset change.

## Acceptance Criteria

- [ ] GIVEN we write harvest records as successful before applying the change to the dataset table  \
  WHEN the dataset change/write happens \
  THEN the harvest record status is "success" \

## Sketch

jaredb96:
Work for [Issue #5935](https://github.com/GSA/data.gov/issues/5935)


jaredb96:
Failing some tests, returning this to draft mode until tests pass

harvester/harvest.py
```diff
@@ -1217,12 +1217,14 @@ def sync(self):
         return False
 
     def update_self_in_db(self) -> bool:
-        finished_date = get_datetime()
+        # set date_finished if it hasn't been set yet
```
jaredb96, line 1216: Since we're now calling `update_self_in_db()` twice, to preserve the previous behavior, we should only be setting `date_finished` once (without this `test_harvest_single_record_created()` fails).

rshewitt:
can you deploy this on dev? all commits to the `develop` branch are deployed to dev app. i'm asking because migrations can be weird sometimes and not work as intended. here's a migration [example](https://github.com/GSA/datagov-harvester/blob/main/migrations/versions/a6aa1afd27b7_remove_csdgm_from_schema_type.py) for updating enums which has worked so far. there's always more than 1 way to approach something i just want to verify it works (migrations can often times be merged and fail to deploy)

jaredb96:
Sure, going forward I'm assuming we should usually be deploying to dev?

rshewitt:
> Sure, going forward I'm assuming we should usually be deploying to dev?

i'll always do that for UI stuff. generally it's good practice to do it but you don't have to all the time. sometimes the tests are good enough but it's up to you. i'm only saying this because migrations can be a headache

jaredb96:
Got it, merged 5935 into `develop`

rshewitt (approved):
looks good. enum has been updated on develop

```terminal
 public | record_status          | record_status          | 4     | error                 +| u6lue1yix22e5kab |                   | 
        |                        |                        |       | success               +|                  |                   | 
        |                        |                        |       | dataset_pending        |                  |                   | 
```

PR #734, opened 2026-06-30
Add static asset cache busting

cody-seibert-gsa:
for work on https://github.com/GSA/data.gov/issues/6133

# Add static asset cache busting

## Summary

Adds catalog-style cache-busted static asset filenames for the harvester app. Static URLs now include a short asset hash in the filename, and the Flask static handler maps those versioned requests back to the on-disk files.

This also adds repo-local deploy wiring, similar to datagov-catalog, so GitHub Actions computes the static asset hash after building assets and passes it to Cloud Foundry as `ASSET_VERSION`.

Screenshot 2026-06-30 at 4 48 50 PM

deployed to https://harvest-dev.data.gov/organization_list/

### DIFF
--- a/.github/actions/deploy/action.yml
+++ b/.github/actions/deploy/action.yml
@@ -1,0 +1,90 @@
+name: "Deploy datagov-harvester"
+description: "Build assets, export dependencies, and deploy to Cloud.gov"
+
+inputs:
+  cf_space:
+    description: "Cloud Foundry space (development, staging, prod)"
+    required: true
+  cf_org:
+    description: "Cloud Foundry org"
+    required: false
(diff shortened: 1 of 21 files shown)


FuhuXia:
this is basically the same way as we did for catalog in
- https://github.com/GSA/datagov-catalog/pull/293 and
- https://github.com/GSA/datagov-catalog/pull/300

The resulted action file are largely in common, we can actually create a deployment template with ASSET_VERSION and caching busting features. But we only two apps doing this, so keeping them separate or using shared template both are ok.

FuhuXia (approved):

PR #754, opened 2026-07-07
6015 cookie hardening

akuny:
Related to https://github.com/GSA/data.gov/issues/6015.

Add `HttpOnly` and `SameSite` attributes to cookie deletion functions in `app/__init__.py`.


jbrown-xentity (approved):
LGTM

PR #756, opened 2026-07-10
Use application/octet-stream placeholder for root describedByType

c-tonneslan:
`Record.fill_placeholders` substitutes a standard media type when `describedByType` is invalid. The root-level branch used a misspelled value, `application/octet-steam` (missing the `r`), which is not a real IANA media type:

```python
# harvester/harvest.py (root level)
self.transformed_data["describedByType"] = "application/octet-steam"   # typo

# ~18 lines down, distribution level — correct
dist_item["describedByType"] = "application/octet-stream"
```

Because `octet-steam` still satisfies the schema's shape-only media-type regex, validation passes and the bogus value is persisted and synced to the catalog — a silent data-quality issue rather than a crash. For example an ISO record whose root `describedByType` is `WWW:LINK-1.0-http--link` (invalid) gets rewritten to `application/octet-steam` instead of `application/octet-stream`.

This aligns the root-level placeholder with the distribution-level one. The existing test `test_transformed_iso_root_describedByType_placeholder` asserted the misspelled value (while its sibling `..._distribution_..._placeholder` already asserted the correct one); its assertion is updated to match.


cody-seibert-gsa (approved):
thank you for this contribution!

FuhuXia:
This PR cant be merged because the commits are not verified.
image

FuhuXia:
replaced by new PR
- https://github.com/GSA/datagov-harvester/pull/902

PR #761, opened 2026-07-13
6001 - Flatten DCAT-US 3.0 fields into existing OpenSearch shapes

cody-seibert-gsa:
## Summary

Adds index-time normalization so DCAT-US 1.1 and 3.0 can share the existing OpenSearch mappings without catalog changes.

- Flattens DCAT3 `theme` Concepts to `prefLabel` strings and Identifier objects to `@id`
- Leaves `inSeries` unmapped (no `isPartOf` alias)
- Sets nested `dcat.dynamic: false` so mixed schema shapes cannot lock incompatible dynamic types

Unlike #718, this keeps legacy string/text contracts instead of promoting `theme` to Concept objects in the index.

## Field mapping

| OpenSearch field | Indexed shape |
| --- | --- |
| `theme` | list of strings (`prefLabel` from Concepts) |
| `identifier` | string (from `@id` when object) |
| `publisher` / `keyword` / `title` / `description` / `distribution_titles` | unchanged contracts |

New mapped fields can be added via `INDEX_FIELDS` in `harvester/opensearch_transform.py`.

### DIFF
--- a/harvester/opensearch.py
+++ b/harvester/opensearch.py
@@ -10,6 +10,7 @@ from botocore.credentials import Credentials
 from opensearchpy import AWSV4SignerAuth, OpenSearch, RequestsHttpConnection, helpers
 from opensearchpy.exceptions import ConnectionTimeout
 
+from database.models import Dataset
 from harvester.opensearch_transform import DcatIndexTransformer
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


harvester/opensearch.py
```diff
@@ -65,6 +68,10 @@ class OpenSearchInterface:
             "last_harvested_date": {"type": "date"},
             "dcat": {
                 "type": "nested",
+                # DCAT-US 1.1 and 3.0 nest differently shaped fields under dcat.
+                # Disable dynamic mapping so the first indexed schema version
+                # cannot lock incompatible types for later records.
+                "dynamic": False,
```
jbrown-xentity: 👍

tests/unit/test_opensearch_transform.py
```diff
@@ -0,0 +1,358 @@
...
+
+
+def test_transform_real_dcat3_complete_example():
+    dcat = json.loads(DCAT3_COMPLETE_EXAMPLE.read_text())
+
+    result = DcatIndexTransformer().transform(dcat)
+
+    assert result["title"] == "National Climate Observations 2024"
+    assert result["publisher"] == "National Climate Data Center"
+    assert result["identifier"] == ""
```
jbrown-xentity, line 335: This makes me sad, but it is correct. That's a bad example on my part; I will clean up as part of https://github.com/GSA/data.gov/issues/6165

harvester/opensearch.py
```diff
@@ -339,21 +346,26 @@ def _walk(value: Any) -> None:
...
         )
         has_spatial = (
             bool(spatial_value and str(spatial_value).strip())
             or dataset.translated_spatial is not None
             or has_spatial_theme
         )
         normalized_dcat = self._normalize_dcat_dates(dataset.dcat)
+        # Keep nested dcat theme/identifier compatible with text/keyword shapes.
+        normalized_dcat["theme"] = indexed["theme"]
+        normalized_dcat["identifier"] = indexed["identifier"]
```
jbrown-xentity: I don't think we want to use `indexed` here. This then changes the dcat object in opensearch for these fields, which we are trying to avoid. Let's leave those alone, but use the indexed/normalized version in the top level for theme and identifier.

jbrown-xentity (commented):
@cody-seibert-gsa one small change, and I think this is good to ship!

harvester/opensearch.py
cody-seibert-gsa: good catch, this should be addressed now

jbrown-xentity (approved):
LGTM

PR #762, opened 2026-07-14
Hide translate spatial payload

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/datagov-harvester/security/code-scanning/38

## About

hide the payload from the log, fix the code QL finding.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/utils/general_utils.py
+++ b/harvester/utils/general_utils.py
@@ -1035,9 +1035,9 @@ def translate_spatial_to_geojson(spatial_input):
         return json.loads(translated_value)
     except json.JSONDecodeError as exc:  # pragma: no cover - defensive
         logger.warning(
-            "Unable to decode translated spatial value %s: %s",
-            translated_value,
-            exc,
+            "Unable to decode translated spatial value as JSON (length=%s, error=%s)",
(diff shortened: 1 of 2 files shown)


harvester/utils/general_utils.py
```diff
@@ -1035,9 +1035,9 @@ def translate_spatial_to_geojson(spatial_input):
         return json.loads(translated_value)
     except json.JSONDecodeError as exc:  # pragma: no cover - defensive
         logger.warning(
-            "Unable to decode translated spatial value %s: %s",
-            translated_value,
-            exc,
+            "Unable to decode translated spatial value as JSON (length=%s, error=%s)",
+            len(translated_value),
+            exc.__class__.__name__,
```
jbrown-xentity, line 1040: Do we really need to hide items from logging @FuhuXia ? This is public data... I wouldn't think hiding the value object from our logging is necessary...

FuhuXia: you are right. i will set to ignore to the issue.

FuhuXia:
no fix needed.

PR #763, opened 2026-07-14
6126: DCAT-US v3 Warnings Infrastructure

akuny:
# Update Harvester Infrastructure to Handle DCAT-US v3.0 Warnings

Addresses https://github.com/GSA/data.gov/issues/6126.

Ahead of adding DCAT-US v3.0 warnings to the harvester, the underlying infrastructure needs to distinguish between "errors" and "warnings" on harvest record issues. This PR includes infrastructure changes exclusively, so no warnings are written or displayed yet.

## Change Summary

- Schema: added a severity enum column (error | warning, default error) to `harvest_record_error`, with a matching `HarvestRecordError` model field and an up/down migration. Verified the up/down/up roundtrip against the local DB.
- `log_non_critical_error` accepts an optional severity param (defaults to error), threaded through `HarvestNonCriticalException`. This is distinct from the existing `is_error` flag, which drives record status.
- `get_harvest_record_errors_by_job` (both the direct and `for_view` query paths), `get_harvest_record_errors_by_record`, and `pget_harvest_record_errors` gained a severity param. Default is severity="error" to preserve existing behavior; pass severity=None to return all issues.
- `get_harvest_record_issues(job_id)` returns all issues (errors + warnings) for a job.

## Notes for Future Work

There are two smells in the code this PR touches, deferred to keep scope tight:

### is_error vs. severity naming collision (harvester/exceptions.py)

`log_non_critical_error` now takes both. They're independent: `is_error` drives the record's status, severity classifies the issue row. The valid-but-confusing combination `is_error=False`, `severity="error"` (the spatial-translation case) shows why. Renaming `is_error` to something like `record_failed` would make the two axes obviously distinct, but it touches call sites beyond this PR.

### Positional-tuple contract on the for_view path

The view path returns raw column tuples that `make_new_record_error_contract` maps back to fields by position (error[3] is type, last two are `identifier/source_raw`). Adding a column to that SELECT silently misaligns the mapping. This is why this PR filters the view path by severity but does not surface severity in its output. The warnings-display work must account for this before adding severity (or any column) to the view output.

## Reviewer Note

This PR updates `database/models.py`, which needs to be kept in sync with https://github.com/GSA/datagov-catalog/blob/main/app/models.py

## PR Tasks

- [x] Code well documented
- [x] Files linted


jbrown-xentity:
```log_non_critical_error now takes both. They're independent: is_error drives the record's status, severity classifies the issue row. The valid-but-confusing combination is_error=False, severity="error" (the spatial-translation case) shows why. Renaming is_error to something like record_failed would make the two axes obviously distinct, but it touches call sites beyond this PR.```

This is funny to me; I think the spatial-translation "error" today should become a warning. It's not something that stops the dataset from loading, and I would consider that a warning, not an error.

akuny:
> `log_non_critical_error now takes both. They're independent: is_error drives the record's status, severity classifies the issue row. The valid-but-confusing combination is_error=False, severity="error" (the spatial-translation case) shows why. Renaming is_error to something like record_failed would make the two axes obviously distinct, but it touches call sites beyond this PR.`
> 
> This is funny to me; I think the spatial-translation "error" today should become a warning. It's not something that stops the dataset from loading, and I would consider that a warning, not an error.

That's a good point: reclassifying spatial-translation as a "warning" would fix the confusing combination, but the confusingly-named parameter pair would still exist. Any thoughts on renaming `is_error`?

jbrown-xentity:
> That's a good point: reclassifying spatial-translation as a "warning" would fix the confusing combination, but the confusingly-named parameter pair would still exist. Any thoughts on renaming `is_error`?

Hmm. I understand the naming convention is pretty confusing. In general we have `critical` and `non-critical` errors, which generally break down into `critical` = job level errors (ie can't pull source, can't compare against our previous pull, something is horribly wrong and we need to abort) and non-critical = individual dataset level errors (not valid metadata, transform failure, sync failure to opensearch, etc).

For most of our non-critical exceptions/errors they are logged in the app, a new error record is created, and the status of the dataset record in our DB is marked as an error. https://github.com/GSA/datagov-harvester/blob/8df3572acc7c16ca8521336e94ec6c8f02ac0b47/harvester/exceptions.py#L60-L82. For the spatial use case we created the ability to stop the update of the dataset record to mark it as a failure without actually changing the DB structure and process to denote it as a warning. So we basically cheated.

I think the functionality as-is shows the need for this feature. When we come back and fix the spatial item to be a warning and not an error, we can remove the `is_error` functionality and have a different log function called `log_warning` or some such that does the appropriate thing. But we don't have to do that today when building the space and infrastructure to apply warnings. That could be part of a future implementation ticket; creating an exception class that isn't used would be a bit silly in this PR.

jbrown-xentity (approved):
LGTM. Coordinate with @rshewitt and/or @FuhuXia on release and DB migration as we've had trouble with that in the past. Might need to stop harvesting in order to release.

PR #767, opened 2026-07-16
Rename datagov-harvest-admin to datagov-harvest

FuhuXia:
correct error in readme.
There is only one app datagov-harvest. No datagov-harvest-admin or datagov-harvest-runner.

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -248,17 +248,14 @@ Note: we prefer that you deploy to the development environment by pushing to the
 ### datagov-harvest-proxy
 This is an nginx app which owns the public route and proxies traffic to the internal Flask app route.
 
-### datagov-harvest-admin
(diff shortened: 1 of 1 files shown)


jbrown-xentity (approved):
LGTM

Aguiardavidm (approved):
This change looks good, as someone who hasn't used this and just read this document it wasnt clear to me what the commands would effect. I am not sure about a dev prod difference on login.gov

FuhuXia:
> This change looks good, as someone who hasn't used this and just read this document it wasnt clear to me what the commands would effect. I am not sure about a dev prod difference on login.gov

Yeah for apps we have dev, staging and prod. for login.gov we can dev (they call sandbox) and prod. I will help you to clear it up on 1-on-1.

PR #768, opened 2026-07-16
integrate data access repo as replacement for db and os functionality

rshewitt:
# Pull Request

Related to [6141](https://github.com/GSA/data.gov/issues/6141)

## About
swap data access functionality with data access repo. description of work [here](https://github.com/GSA/datagov-databases/pull/2). 
- all read/write functionality for opensearch derives from the associated classes. this could be overboard since our db-search compare flask command did batch processing for opensearch but it centralizes the logic and responsibility of data access to a specific place rather than several.
- the major swaps are in the harvester, the models, flask commands, and interfaces (postgres and opensearch)

- [job on dev](https://harvest-dev.data.gov/harvest_job/bf3cc3ec-05ab-44a4-8d62-4a803b437033) with swap
- [sync action](https://github.com/GSA/datagov-harvester/actions/runs/29847714826/job/88692478785)

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


database/cf_connect.sh
jbrown-xentity, line 1: I can't believe you saved this file, but I appreciate it!

database/models.py
```diff
@@ -1,26 +1,20 @@
...
+    HarvestJob,
+    HarvestJobError,
+    HarvestRecord,
+    HarvestRecordError,
+    HarvestSource,
+    HarvestUser,
+    Locations,
+    Organization,
+    ResourceViewCount,
 )
```
jbrown-xentity, line 16: Interesting approach. This will minimize the number of changes in this PR I assume? It will be a strange future artifact; a file that lives to pull in other files for the most part...

pyproject.toml
```diff
@@ -52,6 +52,7 @@ flask-talisman = "^1.1.0"
 newrelic = ">=12.1,<14.0"
 idna = ">=3.15,<4.0"
 mako = ">=1.3.12,<2.0.0"
+datagov-data-access = {git = "https://github.com/GSA/datagov-databases", rev = "base-release"}
```
jbrown-xentity: Before merging I would make sure upstream is released, and target 1.0.0 tag. Could be convinced to just use `main` as dependency will change.

jbrown-xentity (commented):
A few notes, and I see a few failing tests in github actions, but overall looking very good @rshewitt !

database/cf_connect.sh
rshewitt: yeah this kinda highlights my intention of bringing all "data access" functionality over despite us maybe not needing something like this. I didn't want anything lost in the transition. we can further improve once this base release gets out.

database/models.py
rshewitt: yes exactly. just update the entry point instead of all the import statements. we can change that later though if we want.

pyproject.toml
rshewitt: yeah that's my TODO i mentioned in the PR writeup. I would prefer pinning to an exact known "good" version instead of main branch to avoid unintended breaking changes but i'm open to whatever

FuhuXia:
Is the DB models.py in datagov-databases `1.0.0` up to date? I see a few missing indexes such as ix_harvest_record_error_harvest_record_id. There are might be other things missing.

FuhuXia:
Tests are passing. models.py matches what we have before. Good to go.

FuhuXia (dismissed):

FuhuXia (approved):

PR #769, opened 2026-07-17
6127 DCAT-US v3 Warning Detection Logic

akuny:
# DCAT-US v3 Warning Detection Logic

See https://github.com/GSA/data.gov/issues/6127

## Overview
Adds content-quality warning detection for DCAT-US v3 records. These "warnings" are semantic checks that sit above schema validation (likely data-quality issues, incomplete v1.1→v3 migrations) without failing a record. Implements every rule in the agreed spec: https://github.com/GSA/data.gov/issues/6093#issuecomment-4799323747

This is detection only: `detect_dcat_warnings(data)` takes a Dataset dict and returns warnings. It stores nothing and doesn't touch `harvest_record_error`/`log_non_critical_error` because wiring warnings into storage and display is out of scope for this issue.

## Change Summary

- `harvester/utils/dcat_warnings.py`: exposes an entry point and contains one small pure checker per rule
- `harvester/utils/reference_data/*.json`: bundled snapshots for the vocabulary-backed rules (ISO 639-1, IANA charsets, IANA media types, NARA restriction lists). No network calls in the harvest path. Note that NARA lists are hand-curated (HTML-only source).
- `scripts/refresh_dcat_reference_data.py`: this utility regenerates the three machine-readable snapshots on demand. 
- `tests/unit/test_dcat_warnings.py`: 63 unit tests that run independent of any infrastructure, one class per rule.

## Decisions Made

- Use static snapshots over live lookups for reference data since non-blocking warnings should not add per-record network calls
- landingPage URL resolution excluded: the spec flags it out of scope (possible post-harvest job).

## Notes for Reviewers

- `conformsTo`: string entries are IRI-checked; Standard-object entries fall to the universal `@id` check. Matches spec intent, but deviates from a literal read.
- Duplicate-keyword detection is case-sensitive ("Climate" ≠ "climate").
- An Address with only `@id`/`@type` is reported as empty.

## Checklist

- [x] Code well documented
- [x] Files linted


cody-seibert-gsa:
any clue how often this data needs to be refreshed? will this eventually be a cron job github action that runs weekly or daily and creates a PR for us if stuff changed? right now It seems like a manual process, but I'm curious who / what owns running it periodically.

scripts/refresh_dcat_reference_data.py
```diff
@@ -0,0 +1,134 @@
...
+re-running this script.
+
+Machine-readable sources fetched here:
+  - ISO 639-1 language codes ....... id.loc.gov (JSON)
+  - IANA character set names ....... iana.org (CSV)
+  - IANA media types ............... iana.org (one CSV per top-level type)
+
+The NARA restriction authority lists are HTML-only (no CSV/JSON/API) and hold
+~5 stable terms each, so they are hand-curated in nara_restrictions.json rather
+than scraped here. See that file's header for provenance.
```
jbrown-xentity, line 17: This was not actually my intention. The intention was to implement these into the harvester itself and validate/warn at run-time.

There are future warnings that we may want to be non-blocking that will have customized reports (for example resource link checks). These type of reports may need to be run periodically because of metadata getting stale (ie the links were good at harvest time but now they are bad), and we don't want to use them in-harvest because they would significantly slow things down (ie external http calls are expensive from a waiting/time perspective). Nothing I see here seems to be in that vein, it's all self contained logic and will only slow the harvesting imperceptibly.

@akuny can you implement this into the harvest process after validation?

akuny: @jbrown-xentity In-harvest integration is accounted for in this subsequent issue: https://github.com/GSA/data.gov/issues/6128. This PR is just the detection logic, and the refresh script is only for regenerating the vocabulary snapshots, not for running detection out-of-band. If you'd rather the integration land in this PR, let me know and I'll pull in the 6128 work.

akuny:
> any clue how often this data needs to be refreshed? will this eventually be a cron job github action that runs weekly or daily and creates a PR for us if stuff changed? right now It seems like a manual process, but I'm curious who / what owns running it periodically.

Good point. The sources drift slowly as best I can tell, so staleness just means a rare false-positive warning and nothing blocking. But agreed it shouldn't be manual long-term: a GH Action that runs the script weekly and opens a PR if anything changed would be straightforward. If that sounds appropriate, I'll add it. @cody-seibert-gsa

jbrown-xentity (approved):
Looks good, will wait for implementation ticket to see it working/integrated with harvester.

scripts/refresh_dcat_reference_data.py
jbrown-xentity: My misunderstanding of what this script was doing, all good.

PR #771, opened 2026-07-17
5885 zero downtime opensearch

cody-seibert-gsa:
I tested two jobs:

with the alias switch logic: https://github.com/GSA/datagov-harvester/actions/runs/29762229408/job/88419394851

skip alias switch logic: https://github.com/GSA/datagov-harvester/actions/runs/29762881914/job/88421629420
# PR description: Zero-downtime OpenSearch rebuild

## Summary

- Add a manual GitHub Actions workflow for rebuilding the OpenSearch dataset index without interrupting search traffic.
- Set `HARVEST_RUNNER_MAX_TASKS=0`, rolling-restart the harvester, and wait for active harvest tasks to drain before starting the rebuild.
- Build and validate a versioned physical index from PostgreSQL, then optionally switch the `datasets` alias.
- Always restore the previous max-tasks value and rolling-restart the harvester after success or failure.
- Add command, Cloud Foundry task, alias-management, monitoring, and test support for the rebuild process.

## Why

The existing mapping reset process deletes and recreates the live `datasets`
index. That can interrupt search traffic and leaves no validated replacement
ready for cutover. This change keeps the current index serving requests while a
new physical index is built from PostgreSQL, validates the candidate, and only
then switches readers and writers through an alias when the schema is compatible.

## Rebuild sequence

```mermaid
sequenceDiagram
    autonumber
    actor Operator
    participant GHA as GitHub Actions
    participant CF as Cloud Foundry
    participant DB as PostgreSQL
    participant Harvest as Harvest Tasks
    participant OS as OpenSearch

    Operator->>GHA: Run Rebuild OpenSearch Index
    GHA->>CF: Save current HARVEST_RUNNER_MAX_TASKS
    GHA->>CF: Set HARVEST_RUNNER_MAX_TASKS to 0
    GHA->>CF: Rolling restart harvester
    CF-->>GHA: Web instances restarted with capacity 0

    loop Until no active tasks for 30 seconds
        GHA->>CF: List active harvest-job tasks
        CF->>Harvest: Check PENDING, RUNNING, and CANCELING tasks
        Harvest-->>CF: Current task states
        CF-->>GHA: Active task count
    end

    GHA->>CF: Run rebuild-index task
    CF->>CF: Verify HARVEST_RUNNER_MAX_TASKS is 0
    CF->>CF: Verify no harvest tasks are active
    CF->>OS: Create datasets-run-id-attempt
    OS-->>CF: Candidate mapping
    CF->>CF: Validate candidate mapping

    loop PostgreSQL keyset batches
        CF->>DB: Read next dataset batch
        DB-->>CF: Dataset rows
        CF->>OS: Bulk index into candidate
        OS-->>CF: Per-document results
    end

    CF->>OS: Refresh candidate and read document count
    CF->>DB: Recheck PostgreSQL dataset count

    alt Backfill or validation fails
        CF-->>GHA: Fail without changing the live alias
    else Candidate valid and alias switch enabled
        CF->>OS: Atomically switch datasets alias
        OS-->>CF: Alias update acknowledged
        CF-->>GHA: Rebuild completed
    else Candidate valid and alias switch disabled
        CF-->>GHA: Leave datasets alias unchanged
    end

    GHA->>CF: Always restore saved HARVEST_RUNNER_MAX_TASKS
    GHA->>CF: Rolling restart harvester
    CF-->>GHA: Capacity restored and queued jobs can start
```

## Alias behavior

- The logical index name remains `datasets`.
- Physical indexes use unique names such as
  `datasets-<github-run-id>-<github-run-attempt>`.
- On the first cutover, the atomic alias request removes the legacy concrete
  `datasets` index and adds the `datasets` alias to the validated candidate.
- On later cutovers, the same atomic request removes the alias from the old
  physical index and adds it to the new one.
- Previous versioned indexes are retained for recovery after subsequent
  rebuilds.

## Breaking schema changes

When the new mapping is incompatible with the running Catalog (for example a
field rename), disable **Switch the datasets alias to the rebuilt index**:

1. Build and validate the physical index without switching the alias.
2. Deploy Catalog hardcoded to that physical index.
3. After a successful Catalog deploy, point the `datasets` alias at the new
   physical index.
4. Redeploy Catalog back to the logical `datasets` name when ready.

There is no standalone workflow/CLI yet that only points the alias at an
existing physical index; alias cutover today happens inside
`flask search rebuild-index --switch-alias` via `OpenSearchInterface.switch_alias()`.

## Validation and failure safety

- The rebuild refuses to run unless `HARVEST_RUNNER_MAX_TASKS` is `0` and no
  harvest tasks are active.
- The candidate mapping must match the application mapping.
- Every bulk indexing result is checked; any failed document stops the rebuild.
- PostgreSQL counts are checked before and after the backfill.
- The candidate document count must match the final PostgreSQL count.
- Failures before cutover leave the current index or alias unchanged.
- Partially built candidates are retained for diagnosis.
- The workflow always attempts to restore capacity and restart the harvester.
- Per-environment concurrency prevents rebuild and synchronization workflows
  from overlapping.

## Operational notes

- Deploy the application before running the workflow.
- Consider taking an OpenSearch snapshot before the first alias conversion
  because the legacy concrete `datasets` index is removed during that cutover.
- Run the workflow from **Actions → Rebuild OpenSearch Index** and select the
  target Cloud Foundry environment.
- Operator runbook: [`docs/ops/rebuild-opensearch-index.md`](rebuild-opensearch-index.md)
- Design notes: [`docs/opensearch-index-rebuild.md`](../opensearch-index-rebuild.md)


.github/workflows/delete_opensearch_index.yml
```diff
@@ -0,0 +1,68 @@
+permissions:
+  contents: read
+
+name: Delete OpenSearch Physical Index
```
jbrown-xentity, line 4: 👍 This is necessary in the case where we pause and create a new index, but want both catalog and harvester to switch to a new index on pull request build and release (and not before that occurs). Then we will need to go back and manually delete the old index, once everyone is swapped to the new.

app/commands/search.py
```diff
@@ -56,10 +65,112 @@ def _normalize_mapping_for_comparison(value):
...
+    target_index: str,
+    batch_size: int,
+    total_datasets: int,
+) -> tuple[int, int]:
+    total_indexed = 0
+    total_failed = 0
+    last_dataset_id = None
+    batch_number = 0
+    total_batches = (total_datasets + batch_size - 1) // batch_size
+    started_at = monotonic()
```
jbrown-xentity: Correct me if I'm wrong, but this duplicates https://github.com/GSA/datagov-harvester/blob/main/app/commands/search.py#L107-L157 generally. Let's try to merge these 2 together so we don't have 2 functions doing the same thing very differently.

app/commands/search.py
```diff
@@ -70,16 +181,185 @@ def reset_opensearch_mapping():
...
+    default=True,
+    show_default=True,
+    help="Switch the logical datasets alias after a successful backfill.",
+)
+def rebuild_opensearch_index(
+    target_index: str | None,
+    batch_size: int,
+    allow_legacy_index_removal: bool,
+    switch_alias: bool,
+):
```
jbrown-xentity, line 330: I think this needs to exist from a wrapper perspective, but doesn't need to do the actual indexing. We still need to do the checks of turning off harvesting jobs, and making sure we're ready to deploy. However the rest of the steps are in https://github.com/GSA/datagov-harvester/blob/main/app/commands/search.py#L103-L105. Can't we setup the new alias, setup the correct client, and then use the same function for compare?

docs/ops/rebuild-opensearch-index.md
```diff
@@ -0,0 +1,147 @@
...
+Actions.
+
+Design details, validation rules, and failure modes live in
+[docs/opensearch-index-rebuild.md](../opensearch-index-rebuild.md).
+
+## What the workflow does
+
+In short:
+
+1. Saves the current `HARVEST_RUNNER_MAX_TASKS` value.
```
jbrown-xentity: This is a weird thing. Why does it save this? We're manipulating prod, which should always be defined in code in the prod variables. It should just re-deploy on main with the normal variables.

harvester/lib/load_manager.py
```diff
@@ -63,11 +67,16 @@ def _handle_failed_job(self, job):
...
         in_progress_jobs = interface.get_in_progress_jobs()
-        running_tasks = self.handler.get_running_app_tasks()
-        running_harvest_ids = set(self.handler.job_ids_from_tasks(running_tasks))
+        active_tasks = self.handler.get_active_harvest_tasks()
+        if active_tasks is None:
+            logger.warning(
+                "Not cleaning in-progress jobs because tasks could not be listed"
+            )
+            return
+        active_harvest_ids = set(self.handler.job_ids_from_tasks(active_tasks))
```
jbrown-xentity: Please describe more why you changed this. What behavior did you observe that caused the need for a new function with a new handling that if this new function returns "None" we shouldn't clean up the jobs? What does "None" mean in this case, as it seems different from an empty set?

harvester/lib/load_manager.py
```diff
@@ -106,6 +119,8 @@ def _start_new_jobs(self, check_from_task=False):
         jobs = interface.get_new_harvest_jobs_in_past(limit=slots)
         for job in jobs:
             self.start_job(job.id, job.job_type)
+            if interface.get_harvest_job(job.id).status != "in_progress":
+                continue
```
jbrown-xentity: I'm assuming this is trying to fix some bug, but it's unclear what. This is skipping scheduling new jobs for jobs that are not "in_progress". Why? I don't understand what this has to do with the zero down time ticket...

scripts/opensearch_status.sh
```diff
@@ -0,0 +1,213 @@
...
+
+exec python -
+'
+
+cf ssh "$app_name" -i "$app_instance" -T -c "$remote_command" <<'PYTHON'
+from __future__ import annotations
+
+import os
+import sys
+from datetime import datetime, timezone
```
jbrown-xentity: I understand this is a quick one off, but having python code embedded as a string in a shell script is just weird. Suggest splitting this out, and the shell script can just take a parameter for what python script it should run, and make this it's own python script.

jbrown-xentity (commented):
Ok, lots of thoughts and comments here @cody-seibert-gsa , sorry.

Please try to keep the "blast radius" of the code minimal. We're responsible for all created code, even utility tools, and the more we create the more we have to manage and the more we have to review.

Obviously this touches the openSearch code which is on pause with @rshewitt 's work. Will need to coordinate with him on integration and release. Might need to be a team discussion on where these tools/processes even lives. I'm not sure if in the shared interface repo, in harvester, in catalog, or in data.gov as a coordination script.

.github/workflows/rebuild_opensearch_index.yml
```diff
@@ -0,0 +1,114 @@
...
+          cf_username: ${{ secrets.CF_SERVICE_USER }}
+          cf_password: ${{ secrets.CF_SERVICE_AUTH }}
+
+      - name: Start index rebuild
+        uses: cloud-gov/cg-cli-tools@main
+        with:
+          command: >
+            cf run-task datagov-harvest --name "$REBUILD_TASK" -k 4G -m 2G
+            --command "flask search rebuild-index --target-index $TARGET_INDEX
+            --allow-legacy-index-removal $ALIAS_OPTION"
```
akuny: To make sure I'm on the same page:

The `flask search rebuild-index` command (which this workflow runs as a Cloud Foundry task) has a built-in safety check for one specific situation. Right now, `datasets` is a normal index: when the search app asks OpenSearch for `datasets`, it gets one specific index directly. The changes in this PR require deleting that index once, so the name `datasets` can be reused as a re-aimable pointer to whichever index is current.

This command does not delete the normal `datasets` index unless you explicitly pass `--allow-legacy-index-removal`, so the one-time deletion is framed as a deliberate decision. However, this workflow hardcodes the flag.

Instead of passing that flag automatically on every run, could we bind it to a checkbox so that the person invoking the action has to opt into deleting the old index?

akuny (commented):
This looks good to me overall: the docs are particularly thorough. Given the scope of these changes, I'll defer approval to a reviewer with more context, but FWIW this seems solid.

docs/ops/rebuild-opensearch-index.md
cody-seibert-gsa: A full deployment seems broader than necessary (redeploys staging, proxies, etc) for restoring one temporary environment override. Instead of storing the existing variable, I could just read the vars.<environment>.yml and restart the harvester services.

harvester/lib/load_manager.py
cody-seibert-gsa: yeah I think I saw this as a potential bug, but agreed this is not related to the original zero downtime ticket.  I will go through and revert changes that do not align with the original zero downtime goal.

harvester/lib/load_manager.py
cody-seibert-gsa: None means CF task listing failed, while an empty list means it succeeded and found no active tasks. Previously, an API failure was converted into an empty job-ID list and could incorrectly mark every in-progress job as failed. The new helper also includes PENDING and CANCELING.  I can pull these out into a separate branch.

.github/workflows/rebuild_opensearch_index.yml
cody-seibert-gsa: yeah this is honestly a one time thing, after we've used the alias as datasets and delete the old physical index with the duplicate name, we won't even need this functionality anymore.

PR #776, opened 2026-07-20
Potential fix for code scanning alert no. 39: URL redirection from remote source

Aguiardavidm:
Potential fix for [https://github.com/GSA/datagov-harvester/security/code-scanning/39](https://github.com/GSA/datagov-harvester/security/code-scanning/39)

Best practice is **validate untrusted redirect targets immediately before redirecting**, even if they were validated earlier. This creates defense in depth and prevents open redirects if session state is polluted or set from another path.

In `app/main/auth.py`, update the block around lines 125–128 so that after `next_url = session.pop("next", None)`, you only call `redirect(next_url)` if `_is_safe_redirect_target(next_url)` is true. Otherwise, fall back to `redirect(url_for("main.index"))`. No new imports or dependencies are required since `_is_safe_redirect_target` already exists in this file.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/app/main/auth.py
+++ b/app/main/auth.py
@@ -123,7 +123,7 @@ def login():
                 request.headers.get("X-Forwarded-For", request.remote_addr),
             )
             next_url = session.pop("next", None)
-            if next_url:
+            if next_url and _is_safe_redirect_target(next_url):
                 return redirect(next_url)
             return redirect(url_for("main.index"))
         flash("Invalid username or password.", "danger")


Aguiardavidm:
For context we added the safe redirect last week so this looks to me like expanding it to cover more cases. That is why this is new and caught now

jbrown-xentity (approved):
No problem including this. Should never hit this case in practice, this is only for local development (direct login without login.gov is only in local dev environment).

PR #777, opened 2026-07-20
use shared source type enum in harvest source form

FuhuXia:
For
- https://github.com/GSA/data.gov/issues/6186

Changes:
- show `waf-collection` in the source form source type drop down.
- show `collection_parent_url` for `waf-collection` type


app/util.py
```diff
@@ -24,6 +24,10 @@
 
 # Helper Functions
 def make_new_source_contract(form):
+    collection_parent_url = None
+    if form.source_type.data == "waf-collection":
+        collection_parent_url = form.collection_parent_url.data
+
```
Aguiardavidm, line 30: There currently isnt a test that checks if a stale collection_parent_url is violating the DB's wafcollectionparenturl CHECK constraint when a source is edited away from waf-collection. 
Recommendation: add a test in tests/integration/app/test_forms.py that edits an existing waf-collection source, changes source_type to document, and asserts collection_parent_url is None after save.

app/templates/edit_data.html
```diff
@@ -15,7 +15,12 @@ <h1>{{action}} {{data_type}}</h1>
         </p>
         {% for field in form if field.widget.input_type != 'hidden' %}
         {% set field_class = 'usa-textarea' if field.type == 'TextAreaField' else ('usa-select' if field.type == 'SelectField' else 'usa-input') %}
-        <div class="usa-form-group">
+        {% set show_collection_parent_url = form.source_type is defined and form.source_type.data == 'waf-collection' %}
```
Aguiardavidm, line 18: Nit: This template is shared with OrganizationForm (see app/main/organizations.py:43,164), but this block hardcodes a check for field.id == 'collection_parent_url', coupling a generic template to one specific HarvestSourceForm field. Not blocking for a single field, but if we need another conditional field later this won't scale cleanly — might be worth a more generic field.render_kw / data-attribute-driven approach down the line.

app/static/package-lock.json
```diff
@@ -2212,9 +2212,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
```
Aguiardavidm, line 2225: Nit: This diff looks like incidental libc metadata churn from a different local npm version, unrelated to the feature. Can we revert this file or regenerate it cleanly so the PR diff stays focused?

Aguiardavidm (changes requested):
Looks good only one test case I think we should add and a couple of nits.

app/util.py
FuhuXia: test added to cover an existing waf-collection source changed to source_type=document case.

app/templates/edit_data.html
FuhuXia: yes, will need attention if to get more fields like this.

app/static/package-lock.json
FuhuXia: noice removed.

Aguiardavidm (approved):

PR #778, opened 2026-07-21
Address Vulnerabilities

akuny:
To address npm audit findings:

- Pin transient dependency shell-quote to 1.10.0 in `app/static/package.json` overrides to work around concurrently's unpatched shell-quote dependency: https://github.com/GSA/datagov-harvester/security/dependabot/176
- Update root `package-lock.json`

Local build works as expected.

### DIFF
--- a/app/static/package-lock.json
+++ b/app/static/package-lock.json
@@ -23,7 +23,7 @@
         "@uswds/compile": "^1.3.2",
         "babelify": "^10.0.0",
         "browserify": "^17.0.1",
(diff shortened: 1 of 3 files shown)


FuhuXia:
any info about the audit finding?

akuny:
> any info about the audit finding?

You can see more information about the vulnerabilities here: https://github.com/GSA/datagov-harvester/security/dependabot

FuhuXia (approved):
LGTM. looking much better than pinning to an old version.

PR #779, opened 2026-07-22
5885 - adding support to disable harvester from github action

cody-seibert-gsa:
## Summary

Adds a manually triggered GitHub Actions workflow for enabling or disabling
scheduled harvesting in development, staging, or production.

- Disabling sets the existing `HARVEST_RUNNER_MAX_TASKS` variable to `0`.
- Enabling sets it to a positive value that defaults to `3`.
- Uses `cf set-env`, so bound application secrets are not read or modified.
- Removes the variable from the manifest and vars files so later deployments
  preserve the selected state.
- Performs a rolling restart and verifies that every application instance
  loaded the expected value.
- Leaves the existing Python scheduling logic unchanged.

Tasks that are already running are not stopped. Manual harvest triggers do not
use the scheduler's task-limit setting and are not disabled by this workflow.


FuhuXia:
This is good to merge. To make it even more complete:
1. add a short sentence to let user aware that existing harvest jobs are not stopped unless you log in to harvest and manually stop running jobs.
2. add a Check action to show current HARVEST_RUNNER_MAX_TASKS
3. add the step to send notification to slack channel on toggle action like we did [here](https://github.com/GSA/catalog.data.gov/blob/main/.github/workflows/maintenance.yml#L56).

FuhuXia (dismissed):

akuny (approved):
Looks good.

PR #780, opened 2026-07-24
5885 - adding the workflow to do the zero downtime alias deploy

cody-seibert-gsa:
# Zero-downtime OpenSearch index rebuild

## Summary

Adds a zero-downtime way to rebuild the OpenSearch `datasets` index. Today, a
mapping change requires `reset-mapping` (which deletes and recreates the index,
causing a **search outage** until `compare --force-update` repopulates it). This
PR builds a fresh index alongside the live one, backfills it from PostgreSQL, and
atomically swaps a `datasets` **alias** so catalog and harvester never see a
missing or empty index.

Achieved with **no changes to production runtime code** and **no dependency
changes** — the OpenSearch client lives in the pinned `datagov-data-access`
package, so all new index/alias work goes through the raw `opensearch-py` client
it already exposes.

## How it works

The `Rebuild OpenSearch Index` action orchestrates existing pieces:

1. **Disable harvester** — reuses `toggle_harvester.yml` (now callable via
   `workflow_call`, so the disable/enable logic isn't duplicated).
2. **Drain harvest jobs** — `bin/wait_for_harvest_tasks.sh` polls the CF task API
   until in-flight jobs finish; `force_kill_running_jobs` cancels stragglers after
   15 minutes.
3. **Rebuild** — `flask search rebuild-index` creates `datasets-<timestamp>` with
   the current mapping, backfills every document **from PostgreSQL**, validates the
   doc count against the DB, then atomically switches the `datasets` alias.
4. **Re-enable harvester** — always runs, even if the rebuild failed.
5. **Optional cleanup** — `delete_old_index` removes the previous index after a
   successful switch (or use the standalone `Delete OpenSearch Physical Index`
   action later).

### Why backfill from PostgreSQL instead of OpenSearch `_reindex`?

`_reindex` copies the old `_source`, so it can only reproduce the **old document
shape** — it would silently miss new/changed computed fields, which is often the
whole reason for a rebuild. Regenerating from the DB (via the same
`DatasetDocument` builder used in production) picks up document-shape **and**
mapping changes, and mirrors the existing `compare --force-update` repair path.

## Changes

**Additive only** — existing `reset-mapping` / `compare` commands are untouched:

- `app/commands/search.py` — new `flask search rebuild-index` and
  `flask search delete-index` commands.
- `.github/workflows/rebuild_opensearch_index.yml` — orchestration workflow.
- `.github/workflows/delete_opensearch_index.yml` — manual index-cleanup escape hatch.
- `.github/workflows/toggle_harvester.yml` — added a `workflow_call` trigger
  (existing manual dispatch unchanged).
- `bin/wait_for_harvest_tasks.sh` — CF task drain/force-kill helper (pure `cf curl`).
- Tests: `tests/unit/test_search_commands.py`, `tests/scripts/test_wait_for_harvest_tasks.py`.

## Notes for reviewers / operators

- The alias is named **`datasets`** (not `datasets-alias`) so catalog and harvester
  keep working with zero code changes.
- The **first** production run needs `--allow-legacy-index-removal` to convert the
  current concrete `datasets` index into an alias in one atomic step. The workflow
  always passes this flag.
- Search stays available the entire time: the old index serves reads until the
  atomic cutover.

### DIFF
--- a/.github/workflows/delete_opensearch_index.yml
+++ b/.github/workflows/delete_opensearch_index.yml
@@ -1,0 +1,73 @@
+name: Delete OpenSearch Physical Index
+
+# Manually delete an unused physical dataset index (for example, one left behind
+# by a rebuild that was run with --no-delete-old-index). The command refuses to
+# delete an index still attached to the datasets alias, so live search cannot be
+# removed by mistake.
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Cloud Foundry environment"
+        required: true
+        default: development
+        type: choice
+        options:
+          - development
+          - staging
+          - prod
+      index_name:
+        description: "Exact unused physical index name (for example, datasets-20260723152900)"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: opensearch-maintenance-${{ inputs.environment }}
+  cancel-in-progress: false
+
+jobs:
+  delete:
+    name: Delete OpenSearch index (${{ inputs.environment }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    environment: ${{ inputs.environment }}
+    env:
+      CF_SPACE: ${{ inputs.environment }}
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v7
+
+      - name: Validate physical index name
+        env:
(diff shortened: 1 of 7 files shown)


akuny (approved):
This looks good, no notes from me for what that's worth.

PR #786, opened 2026-07-27
use body-file instead. template expects string.

rshewitt:
# Pull Request

Related to [6191](https://github.com/GSA/data.gov/issues/6191)

## About
`--template` expects a string not a file object. swapping to use `--body-file`. here's my local [test](https://github.com/GSA/datagov-catalog/pull/361)

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM

PR #787, opened 2026-07-27
Add SESSION_IDLE_TIMEOUT_SECONDS to environment variables

FuhuXia:
local box idle time after 8 hours, not 15 mins.


rshewitt (approved):
sure, the convenience of not having to login as frequently when running locally is nice

jbrown-xentity (approved):
LGTM

PR #790, opened 2026-07-29
5885 zero downtime opensearch timeouts

cody-seibert-gsa:
# Zero-downtime OpenSearch cluster migration

Automates the whole OpenSearch cluster migration, and then wires it into the release so a
mapping change rebuilds the index **on merge** instead of needing someone to remember.
Resizing or re-mapping the `datasets` index no longer means a hand-run sequence of `cf`
commands with search down for the duration.

Previously a rebuild happened **in place**: the index was dropped and backfilled on the
live cluster, so search returned nothing until the backfill finished. And a resize was
not possible at all without a new instance — the Cloud Foundry broker cannot resize an
existing OpenSearch instance in place.

## How it works

The migration builds a **second cluster** alongside the live one, fills and verifies it,
then moves both apps onto it. The live cluster serves queries at full speed the entire
time.

```
disable harvesting
  → provision replacement cluster, bind both apps
  → drain in-flight harvest jobs
  → rebuild the `datasets` index into the replacement (backfilled from PostgreSQL)
  → verify it matches PostgreSQL exactly          ← gate
  → promote: rename the cluster, roll both apps
  → verify again                                  ← gate
  → delete the replaced cluster
re-enable harvesting (always, even on failure)
```

Three ideas do most of the work:

**Promotion is a rename, not a repoint.** Both apps resolve one canonical service name
(`datagov-catalog-opensearch`). The promote renames `<canonical>` → `<canonical>-old`,
then `<next>` → `<canonical>`, so the cluster moves *underneath* a name that never stops
resolving. Neither app is repointed with `cf set-env`. The subsequent rolling restarts
are mandatory, not cosmetic: a rename is metadata-only, and `.profile` resolves the host
exactly once at container start, so a running instance would otherwise stay on the old
cluster indefinitely.

**The replacement is reached via `--cluster next`.** `flask search rebuild-index` and
`compare` take a `--cluster live|next` flag. `next` resolves the replacement's
credentials from `OPENSEARCH_NEXT_*`, which `.profile` exports when
`OPENSEARCH_NEXT_SERVICE_NAME` is set. It refuses to run if `next` resolves to the same
host as `live`, so a misconfigured run can't silently hit production while reporting
"next".

**The backfill reads PostgreSQL, not OpenSearch.** Rebuilding from the source of truth
(rather than a server-side `_reindex`) means the rebuild picks up document-shape changes
as well as mapping changes, and reuses the same path as `compare --force-update`.

### Safety properties

- **Nothing touches the live cluster before the promote.** Every stage up to that point
  is free to fail; on failure the half-built replacement is removed and the live cluster
  is exactly as it was.
- **The verification gates actually gate.** `compare --fail-on-discrepancy` exits
  non-zero on any missing/extra/stale document. Without that flag `compare` only reports
  and exits 0, so an unverified cluster would sail through.
- **The replaced cluster is deleted only after both gates pass**, so rollback stays
  available through every risky step. `delete_opensearch_cluster.sh` independently reads
  each app's resolved host and refuses to delete anything still being served.
- **Harvesting is always re-enabled**, including on failure or cancellation.
- **A shared concurrency lock** keeps Rebuild / Sync / Cutover / Delete *and* the
  10-minute restart cron from running against the space mid-migration.

## Automatic reindex on merge

Label a PR **`force re-index recommended`** and the index rebuilds itself after each
space deploys — development on a merge to `develop`, staging then prod on a merge to
`main`. This is the half that makes the automation reachable without a hand dispatch.

Nothing else detects the need. `MAPPINGS` carries no version or hash, and
`_ensure_index()` only creates an index when one is *absent*, so deploying a mapping
change against an existing index is a **silent no-op** and search quietly serves the old
shape.

### One reusable pipeline, three spaces

`release-space.yml` is the whole per-space release, called once by `commit.yml`
(development) and twice by `deploy.yml` (staging, then prod), so all three spaces run
byte-identical logic:

```
pause harvesting → create services → push → await rollout → reindex → network policies
resume harvesting  ← always()
```

**Harvesting is paused before the deploy, not inside the rebuild.** This is the ordering
fix from #834. `migrate_opensearch_cluster.yml` opens with its own `disable-harvester`,
which was correct when that workflow was only ever dispatched *after* a deploy had
landed — but wrapping a deploy in front of it moved the pause from "first in its own
workflow" to "third in the pipeline". The window is not brief: the drain waits up to
**2h** for in-flight jobs, and every one of them indexes per dataset, so every write in
that window went into an index whose mapping no longer matched the document. Additive
mapping changes tolerate that; a retyped or renamed field does not.

Only labelled releases pause. A plain deploy is unaffected — rolling deploys are already
safe for harvest jobs.

The resume is gated `!= 'skipped'` rather than `== 'success'`, deliberately:
`set_harvest_runner_capacity.sh` does `cf set-env` and *then*
`cf restart --strategy rolling`, so a disable whose restart times out reports failure
**having already written 0**. `== 'success'` would strand harvesting off in exactly that
case. `skipped` is still precisely the plain-deploy case, so a hand-paused harvester is
left alone.

### Detection measures from the last successful run

Not from the push's own commit range. A release queue allows one pending run and a third
merge *cancels* the pending one, so a labelled merge could be superseded and its reindex
silently lost. Instead `detect-reindex.yml` measures from the `head_sha` of the last
**successful** run of the calling workflow, and a cancelled or failed run never advances
that watermark — the obligation stays detectable until a run actually completes.

Detection lives in the caller, not in `release-space.yml`, because a workflow invoked via
`uses:` produces no workflow run of its own (the API attributes it to the caller), so a
reusable file has no run history to measure from.

`detect-reindex-label.sh` **fails closed**: no watermark, rewritten history, or a range
truncated past the compare API's 250-commit cap all stop the deploy rather than guess.
Nothing is lost when it refuses — dispatch manually with `reindex: force` or
`reindex: skip` to state the intent.

### Two backstops

- **`preflight-clusters`** refuses to start a labelled release if
  `datagov-catalog-opensearch-next` already exists in a target space, which means a
  migration is in flight or parked. It runs before staging *and* prod, so prod is checked
  before staging deploys rather than hours later.
- **`reindex-owed`** opens a tracking issue from `.github/reindex_owed.md` if a run
  decided a reindex was needed and did not finish one. The watermark already keeps the
  obligation detectable; the issue makes it visible to a human rather than only to the
  next run.

### Each release path is a queue

| Path | Queue | Scope |
| --- | --- | --- |
| `develop` → development | `release-develop` (job level) | that one release |
| `main` → staging + prod | `release-main` (workflow level) | **both** spaces in one hold |

`release-main` is workflow-level because that run spans two spaces: a job-level group
would be released between staging and prod and let a second merge's staging deploy slip
into the gap. `release-develop` is job-level on purpose — `commit.yml` also runs
lint/test on *every* branch, and a workflow-level hold there would serialize unrelated
PRs' CI.

A `cf push` landing mid-migration is what broke a staging run on 2026-08-10, redeploying
the app and removing the `--cluster` flag the in-flight rebuild depended on. Consequence
worth knowing: **a labelled merge blocks later merges to that branch for the length of
its migrations** — hours, for a prod `es-large`. Dispatch manually with `reindex: skip`
to jump that queue.

`release-space.yml` deliberately holds **no** workflow-level concurrency group: it calls
`migrate_opensearch_cluster.yml`, which takes `opensearch-maintenance-<space>` at its own
workflow level, and a hold above would queue the child behind its own ancestor forever.
The cf-touching jobs take that group at *job* level instead.

The 10-minute **restart cron** now shares the same per-space lock. A rolling restart
landing mid-rename either fails `.profile`'s empty-host guard or supersedes the
migration's own deployment — making the migration's blocking `cf restart` return
non-zero right after the renames, the worst possible place to stop.

### 1. Normal rebuild (no code changes)

For a resize, or a mapping change already deployed and backwards-compatible. One
dispatch does everything; search stays up throughout.

| Input | Value |
| --- | --- |
| `environment` | target space |
| `start_at` | `provision` |
| `stop_after` | `decommission` |
| `next_plan` | blank to match the live plan, or e.g. `es-large` to resize |

Dispatch it and walk away. Both apps end up on the new cluster under the canonical
name, and the replaced cluster is deleted.

### 2. Coordinated rebuild (needs a harvester or catalog deploy)

For a **schema-breaking** change — one where the new mapping or document shape only
works with new application code. Both apps must move together and must not read the new
mapping before their code ships, so the migration splits in two around the promote.

**Run 1 — build and park:**

| Input | Value |
| --- | --- |
| `start_at` | `provision` |
| `stop_after` | `verify` |

Stops with the replacement built, bound, verified, and **inert**. Both apps are still
serving from the live cluster; nothing moved. Safe to leave for days.

**Then deploy** the harvester and catalog code carrying the new schema.

**Run 2 — promote and clean up:**

| Input | Value |
| --- | --- |
| `start_at` | `cutover` |
| `next_service_name` | same name as run 1 |
| `stop_after` | `decommission` |

Both apps move together in run 2 and never disagree about the schema. The split is
*before* the promote rather than inside it because catalog's `.profile` resolves
`${APP_NAME}-opensearch` with no override — the rename is the only thing that can move
catalog, so splitting it from the harvester's move would leave catalog reading a
different cluster than the harvester writes.

### Resuming

A prod `es-large` provision can take **~2 hours** (AWS quotes 15–30 min per node), so
`start_at` exists to skip work already done. Every stage is idempotent and detects
existing state, so a re-dispatch is safe.

> [!IMPORTANT]
> For a long prod provision, dispatch with **`on_build_failure: keep`**. The default is
> `delete`, which discards the half-built cluster on failure — right for development
> where re-provisioning is cheap, wrong when it costs hours. `keep` is also the
> precondition for resuming with `start_at: rebuild`.

Note that GitHub's **"re-run failed jobs" reuses the original inputs**, so it cannot
change `start_at` and will not resume mid-migration. Recover with a *fresh dispatch*
using `start_at: rebuild` (replacement already built) or `cutover` (already verified).

### Other inputs

| Input | Purpose |
| --- | --- |
| `keep_old_cluster` | promote but keep the replaced cluster as a rollback path |
| `force_kill_running_jobs` | cancel harvest jobs still running after 15 minutes |
| `max_tasks` | `HARVEST_RUNNER_MAX_TASKS` to restore when re-enabling |

## Known limitations

**The automated label path stops the writes, not the reads.** From the moment the deploy
lands until the promote, the new code — and catalog — still query the *old* index. On a
schema-breaking change that means wrong results for the full provision + rebuild
duration: ~20 min on `es-medium`, hours on a prod `es-large`.

Closing that needs two-phase mode, which the label path **cannot currently reach**:
`release-space.yml` hardcodes `stop_after: decommission` and exposes no `stop_after`
input. Follow-up work is to thread it through and select it from a second label. Until
then, for a change that breaks reads, deploy with `reindex: skip` and drive the migration
by hand in two phases.

**Catalog is not verified automatically.** Before a labelled merge reaches prod, confirm
`GSA/datagov-catalog` is running code that matches the new mapping. `compare` checks id
sets and `last_harvested_date`, never document shape, so it cannot see a mismatch. Each
promote says this in Slack and probes catalog's `/search` afterwards, warning if a common
term returns zero results.

## Also in this PR

- `rebuild-index` tolerates **individually-rejected documents** up to
  `--max-failed-records` (default 50) instead of discarding a whole multi-hour backfill
  over one malformed record. Every failed id is printed in full for investigation, and
  the count is carried into the final document-count assertion so a document lost for
  any *other* reason still fails. `compare --fail-on-discrepancy` takes the same flag and
  default, so a rejection the rebuild tolerates does not then fail verification —
  and it applies only to *missing* documents; extra or stale ones always fail.
- **Index creation timeout** raised to 300s. `indices.create` waits for shards to become
  active, which can exceed the client's 60s socket timeout on a loaded cluster; the
  client then retries and fails with `resource_already_exists_exception`. That case is
  now treated as success, making the retry idempotent.
- **`ensure_index=False`** on the OpenSearch client, so `rebuild-index` owns index
  creation rather than having the constructor create it on connect (ported from
  GSA/datagov_data_access#11).
- Dropped the `datasets` **alias** in favor of one rebuild per replacement cluster —
  index names are scoped to a cluster, so there is nothing to coordinate. `datasets` is
  now a plain index on every cluster; the rebuild still detects and clears a leftover
  alias from an older release, so a cluster in either state ends up equally clean.
- Mapping comparison normalizes the `dynamic` flag, which OpenSearch echoes back as the
  string `"false"` where the application mapping declares a bool.
- **`bin/wait_for_deployment.sh`** gates the reindex on the rollout actually draining.
  The deploy pushes with `--no-wait`, and the rebuild runs via `cf run-task` against the
  app's *current* droplet — if that is still the old one it writes documents in the OLD
  shape into the NEW cluster and both gates pass anyway, because `compare` never checks
  document shape.
- **`bin/report_opensearch_cluster.sh`** prints each app's resolved cluster host, which
  is the fastest way to see where things actually landed after a promote or a failure.
- **`shellcheck` in `make lint-check`** (skipped when not installed). The `bin/` scripts
  rename and delete production service instances, so a shell slip there is expensive.
- **Local two-cluster testing**: an opt-in `next-cluster` docker-compose profile plus
  `OPENSEARCH_NEXT_HOST`, so the `--cluster next` path can be exercised without cloud.gov.
- The replacement cluster is deliberately **not** listed in `manifest.yml` — a manifest
  cannot reference a service instance that does not exist, so listing it would break
  deploys in every space not yet provisioned. Manifest application is additive, so
  `cf bind-service` survives deploys on its own.

## Docs

`docs/ops/migrate-opensearch-cluster.md` is the operator runbook: the automated label
path, both manual run modes, resuming, rollback, the emergency rename-service swap, and
the catalog-side changes. `docs/ops/migrate-opensearch-cluster-sequence.html` is a
sequence diagram of the promote.

## Testing

- **~1,500 lines of new script tests** covering `.profile` host resolution, provision,
  promote, cluster deletion, `wait_for_deployment`, and label detection — plus a
  substantially expanded `test_search_commands.py`.
- The migration has been **run end-to-end against staging** — run `31432510233`, all
  green: provision → rebuild → verify → promote → verify → decommission, with harvesting
  toggled around it. (Dispatched via a temporary scratch runner, since a `workflow_call`
  workflow cannot be dispatched from an unmerged branch; the scratch file is deleted.)
- The label-driven release path was exercised in development, first **unlabelled** —
  confirming the deploy path and that detection correctly skips — and then labelled.

### Before merging to `main`

`main` has **none** of this pipeline, so `deploy.yml` and `commit.yml` arrive here in
their new form. The regression to watch on the first merge is a **plain, unlabelled
deploy**: `create-services` now depends on a job that skips on the common path, and a job
whose `needs` were skipped is itself skipped *before* its own `if:` is evaluated — hence
the explicit `contains(fromJson('["success","skipped"]'), ...)`. If that condition were
wrong it would break **every** deploy, not just labelled ones.


cody-seibert-gsa:
here is the successful run on staging https://github.com/GSA/datagov-harvester/actions/runs/31432510233

cody-seibert-gsa:
deploying to develop without the label and verified it doesn't automatically run the reindex https://github.com/GSA/datagov-harvester/actions/runs/31518201599

cody-seibert-gsa:
here is an update on the latest test https://github.com/GSA/datagov-harvester/actions/runs/31528075293 to develop after merging this PR with the label https://github.com/GSA/datagov-harvester/pull/833

cody-seibert-gsa:
https://github.com/GSA/datagov-harvester/pull/842

PR #792, opened 2026-07-29
Commit Signing for Dependency Bumps

akuny:
Create dependency-bump commits via GitHub API so they are signed.


rshewitt (approved):
i imagine `DATA_ACCESS_RELEASE_TOKEN` could be updated with the necessary permission(s) to do what `secrets.GITHUB_TOKEN` is doing to avoid this change? maybe not. i like the use of `createCommitOnBranch` instead of setting up an ssh key.

PR #793, opened 2026-07-29
Add ON DELETE CASCADE to harvest source/job/record foreign keys

cody-seibert-gsa:
# Pull Request

Related to the staging 502 when deleting a harvest source.

Pairs with GSA/datagov_data_access#9.

## About

Deleting a harvest source on staging returned 502 and the source was never removed.

The ORM was doing the cascade row-by-row: every child relationship had `passive_deletes=False`, so SQLAlchemy loaded each child `harvest_record` — including `source_raw` — into the web worker and emitted one `DELETE` per row. On staging source `99c4e78b` that is **44,694 records / 582 MB** (measured) into a 1 GB instance with 3 workers. The kernel OOM-killed the worker, confirmed in staging logs as `Worker (pid:175) was sent SIGKILL!` with **no preceding `WORKER TIMEOUT`** (ruling out gunicorn's 120s timeout). Nothing committed, so the source remained and the dropped connection surfaced as a 502 with `x-cache: Error from cloudfront`.

This migration pushes the cascade into the database. The matching `passive_deletes=True` change lives in datagov_data_access#9 — **this migration is safe to deploy on its own** (it just makes the schema cascade correctly), but it does not fix the OOM until the ORM side ships.

### Lock-safety

Each constraint is dropped and re-added `NOT VALID`, then validated in a separate statement. A plain FK add holds `ACCESS EXCLUSIVE` while it scans the entire child table; `VALIDATE CONSTRAINT` only needs `SHARE UPDATE EXCLUSIVE`. That matters because `harvest_record` is large in prod. The new index uses `CONCURRENTLY` inside an `autocommit_block`, following the pattern in `1f6d2c9a8b3e`.

### `harvest_record_error.harvest_record_id` → SET NULL, not CASCADE

Record errors intentionally outlive the record they describe (`test_harvest_record_error_remains` in the data-access repo asserts this). They are still removed when the owning job or source goes away, via `harvest_job_id`'s cascade.

### Also adds `ix_harvest_job_harvest_source_id`

The one cascade lookup path without an index. To be clear, this was **not** a cause of the 502 — `EXPLAIN ANALYZE` measured that seq scan at 3.4 ms — but an unindexed cascade would be a real problem at prod scale.

## Testing

FK constraint names were confirmed against a real migration-built database before writing the migration, not guessed.

Round-trip on a clean DB built from the full migration chain — `upgrade` → `downgrade` → `upgrade`, all clean. Resulting state:

```
harvest_source_organization_id_fkey           c  (validated)
harvest_job_harvest_source_id_fkey            c  (validated)
harvest_record_harvest_job_id_fkey            c  (validated)
harvest_record_harvest_source_id_fkey         c  (validated)
harvest_job_error_harvest_job_id_fkey         c  (validated)
harvest_record_error_harvest_job_id_fkey      c  (validated)
harvest_record_error_harvest_record_id_fkey   n  (validated)   <- SET NULL
```

End-to-end cascade, in a rolled-back transaction — **one statement**:

```
before | jobs 1 | recs 500 | jerr 1 | rerr 1 | ds 1
DELETE 1
after  | jobs 0 | recs 0   | jerr 0 | rerr 0 | ds 0
```

And the SET NULL semantics verified both ways: deleting only the record keeps its error (orphaned, `harvest_record_id IS NULL`); deleting the source then removes it.

No regressions. Both suites have pre-existing failures, baselined by removing this migration and re-running:

| suite | baseline | with migration |
|---|---|---|
| `tests/integration` | 13 failed, 152 passed | identical |
| `tests/unit` | 14 failed, 263 passed | identical |

Pre-existing failures are CF task mocking and an unrelated OpenSearch `dcat.modified` mapping error. Ruff clean.

## Deploy sequencing

1. Merge + tag datagov_data_access `1.2.0`.
2. Bump the pin here (`pyproject.toml:55`, currently `1.0.0`), `poetry lock`, commit the lockfile.
3. Deploy — `app-start.sh` runs `flask db upgrade` on instance 0.

This PR is step 0 and can merge independently. **Not** included: the pin bump, since Poetry resolves the tag from the remote and it would break the build before step 1.

Separately, staging source `99c4e78b-34e1-4c4b-a8fe-920489f4e2d1` is still stuck and does not need to wait for this — it can be removed now with explicit child-first deletes in a transaction, followed by `VACUUM (ANALYZE) harvest_record;`.

## Follow-ups not in scope

- **`records_warned` drift.** data-access `1.1.0` added `harvest_job.records_warned` and there is no migration for it here. Whoever bumps the pin past `1.0.0` must add that column or `HarvestJob` queries will fail.
- **Background worker.** Even a correct cascade blocks an HTTP worker for the duration. A CF task, like the existing harvest jobs, is the right long-term shape.
- **The 409 guard.** Left as-is deliberately. It counts only latest-non-deleted records, so a cleared source reports 0 while still holding 44k rows — which is what let this delete through. Harmless now that the DB does the cascade quickly. Note `synced=True` at `harvest.py:242` is silently ignored; nothing reads that kwarg despite the docstring.

## PR TASKS

- [x] Code well documented
- [x] Files linted


pyproject.toml
```diff
@@ -52,7 +52,7 @@ flask-talisman = "^1.1.0"
 newrelic = ">=12.1,<14.0"
 idna = ">=3.15,<4.0"
 mako = ">=1.3.12,<2.0.0"
-datagov-data-access = {git = "https://github.com/GSA/datagov_data_access.git", rev = "1.1.0"}
+datagov-data-access = {git = "https://github.com/GSA/datagov_data_access.git", rev = "1.2.0"}
```
Aguiardavidm, line 55: I know that this is already behind main as Im pretty sure we are on 1.2.1 on main. I just had to update my pr

cody-seibert-gsa:
will remake in a new pr

PR #795, opened 2026-07-30
6180: persist DCAT-US3 catalog metadata during harvest

Aguiardavidm:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/6180

## About

Adds support for retaining DCAT-US 3.0 catalog-level metadata during a harvest, per the ticket's sketch:

- Adds an Alembic migration for the new `dcatus_catalog` JSONB column on `harvest_job`.
- `acquire_minimum_external_data` now captures the full downloaded catalog for `dcatus3.0` sources, recursively extracts every dataset (including datasets nested in sub-catalogs, however deep), and persists a stripped copy of the catalog metadata (with `dataset`/`service`/`record` removed, recursively) onto the job via `update_harvest_job`.
- New helpers in `harvester/utils/general_utils.py`: `strip_dcatus3_catalog_objects` and `extract_dcatus3_catalog_datasets`.
- New fixture (`dcatus3_0_nested_catalog.json`) and tests covering both the flat and nested-catalog cases.

**Blocked on / depends on:** [datagov_data_access#10](https://github.com/GSA/datagov_data_access/pull/10), which adds the `dcatus_catalog` column to the shared `HarvestJob` model. This PR's `pyproject.toml` currently pins `datagov-data-access` to that PR's branch commit directly (`rev = "fb5fd8b..."`), **not** a released version — once that PR merges and a new version is tagged/released, this pin needs to move to `tag = "<version>"` and the lockfile regenerated before this is ready to merge.

Manually verified end-to-end against a real DCAT-US 3.0 source (OpenTopography, `https://portal.opentopography.org/csw/catalog30.json`) — harvested all 798 datasets successfully and confirmed `dcatus_catalog` persisted the expected stripped catalog metadata.

## PR TASKS

- [x] Code well documented
- [x] Files linted


Aguiardavidm:
Superseded by #812 — rewritten now that datagov_data_access is deprecated and its models are vendored directly into this repo, so the git-pinned dependency this PR relied on no longer applies.

PR #796, opened 2026-07-30
add "geospatial" to theme in ISO records

rshewitt:
# Pull Request

Related to [6162](https://github.com/GSA/data.gov/issues/6162)

## About
adds "geospatial" to the DCATUS "theme" array when not present

## TODO
- notify chris at geosplatform of the update and force-harvest all waf and waf-collection sources

## PR TASKS

- [x] Code well documented
- [x] Files linted


harvester/harvest.py
```diff
@@ -960,6 +960,20 @@ def add_parent(self) -> None:
...
             self.transformed_data["isPartOf"] = self.parent_identifier
 
+    def add_geospatial(self) -> None:
+        """
+        adds the term "geospatial" in the "theme" array when not present
+        """
+        if not self.transformed_data.get("theme"):
+            self.transformed_data["theme"] = ["geospatial"]
+        else:
+            has_spatial_theme = any(
```
Aguiardavidm, line 970: Nit: maybe rename to has_geospatial_theme? Just to match the term used elsewhere.

harvester/harvest.py
```diff
@@ -960,6 +960,20 @@ def add_parent(self) -> None:
         if self.parent_identifier:
             self.transformed_data["isPartOf"] = self.parent_identifier
 
+    def add_geospatial(self) -> None:
+        """
+        adds the term "geospatial" in the "theme" array when not present
+        """
+        if not self.transformed_data.get("theme"):
+            self.transformed_data["theme"] = ["geospatial"]
```
Aguiardavidm, line 968: I think theme can technically be a string per the schema, not just a list — might be worth an isinstance(..., list) check just in case, though this may never actually happen in practice

tests/integration/harvest_job_flows/test_harvest_job_full_flow.py
```diff
@@ -223,6 +223,12 @@ def test_harvest_waf_iso19115_2(
             record for record in harvest_job.records if record.status == "success"
         ]
 
+        # all ISO records must have "geospatial" in theme
+        assert all(
```
Aguiardavidm, line 227: Nit: This mostly covers the "no theme yet" case, I think. Might be nice to have a small unit test for the append/dedup branches too

jbrown-xentity (approved):
While I think some of @Aguiardavidm 's comments are valid and worth considering for test cases (ie the de-duplicate), having duplicate themes shouldn't have any impact on the UI experience or filtering, which is the main concern. I think it's ok without it, but wouldn't mind if it was validated.

Aguiardavidm (approved):

PR #799, opened 2026-07-31
Warning count and message

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/6204
- https://github.com/GSA/data.gov/issues/6208

## About
- displayed warning count on job summary
- fixed some error/warning count issue

Deployed to dev
https://harvest-dev.data.gov/harvest_job/5b8b027b-52f4-452c-bc58-ff4a9eed7772

image
image
image

## PR TASKS

- [ ] Code well documented
- [ ] Tests written, run and passed
- [ ] Files linted

Issue #797, opened 2026-07-30
Guard against model/migration drift between datagov-data-access and the harvester


cody-seibert-gsa:
Models live in `datagov-data-access`; Alembic migrations live here. Nothing enforces that a model column added there gets a migration here, so a version bump can land a column the database does not have. Any query selecting the model then fails with `column ... does not exist`.

## This has now happened twice with the same column

`harvest_job.records_warned` was added to the model in data-access 1.1.0.

1. PR #794 bumped the pin to 1.1.0 and added the migration (`b3f4a9c1d7e2`).
2. Independently, PR #793 hit the missing column while testing and wrote its own migration for it (`f4a2b8c15d93`).

Both landed. After #793 merged main, the branch had **two Alembic heads and two migrations adding the same column**. `flask db upgrade` runs on every deploy (`app-start.sh:9`) and fails on multiple heads, so the branch would have broken deploys. Fixed in 9a4f5c9 by dropping the duplicate and relinking to a single head.

So the failure mode is not only "migration missing" — it is also "two people write the same migration and collide."

## Suggested guards

- **CI check for multiple heads.** Cheap and catches the collision class directly:
  ```
  flask db heads   # fail the build if it reports more than one
  ```
- **CI check that models match migrations.** `alembic check` (or autogenerate-and-diff) fails when a model column has no corresponding migration, catching the drift at the bump rather than in production.
- **Checklist item on data-access bump PRs:** state explicitly whether the release added or changed columns, and link the migration if so.

The first two are the durable fixes; the multiple-heads check alone would have caught this instance in CI.

## Reference

- Column: `harvest_job.records_warned`, `server_default="0"`, added in data-access 1.1.0
- Migration on main: `migrations/versions/b3f4a9c1d7e2_add_records_warned_to_harvest_job.py`
- Related: #793, #794

rshewitt:
interesting, so even if the `data-access` repo just contained db models, for example, we'd still run into this situation where...
1. a db change occurs following by a new release (or no release and simply merge into main if ditching releasing all together)
2. multiple developers working on harvester simultaneously who integrate that change run migration locally causing multiple separate migrations leading to multiple heads.

[...]

PR #807, opened 2026-08-03
6209 - Reintegrate DB and OpenSearch models with harvester

Aguiardavidm:
## Summary

Reverses PR #768: pulls DB models, the harvester DB interface, and OpenSearch client/writer/reader/query code back into this repo from `datagov_data_access`, so models and the migrations that describe them are co-located again — this is issue #6209's "Option A" baseline, vendored at library tag `1.1.0` (the version this repo already pinned).

- `database/models.py`, `database/interface.py`, `database/configs.py`, `database/decorators.py` — vendored from `datagov_data_access.db.*`
- new top-level `search/` package — vendored from `datagov_data_access.search.*` (client/writer/reader/documents/transforms/spatial/queries). Lands at the top level rather than under `harvester/` to avoid a circular import (`database/interface.py` needs OpenSearch only inside `update_dataset_slug`, so that import is function-local)
- `database/cf_connect.sh` / `database/load_locations.sh` restored
- `harvester/opensearch.py` shim deleted; its 3 consumers now import `search.*` directly
- the `datagov-data-access` git dependency is dropped from `pyproject.toml`/`poetry.lock` entirely
- restores the DB and OpenSearch tests deleted in #768, ports the library's search/query test coverage, and dedupes `query_filter_builder` (previously defined twice)
- adds two CI guards for the drift class described in GSA/datagov-harvester#797 (model/migration disagreement invisible to a test suite that builds schema via `db.create_all()`): a `flask db check` step, and an `import database.interface` smoke check for the circular-import class of bug

**Not vendored:** `db/interfaces/catalog.py` and catalog-only tests/fixtures — `datagov-catalog` still depends on `datagov_data_access` for those, so the library is not deleted, only unused by this repo going forward.

**Deliberately out of scope** (per the issue's Option A): the 1.1.0→1.2.1 cascade-delete model changes (PR #793's territory) and the severity-query interface changes (PR #799's territory) are excluded. Both currently pin the library past 1.1.0 and will need to rebase their model/interface diffs directly into this repo's vendored code. #798/#800 (dependency bump PRs) have nothing left to bump and should be closed.

## Test plan

- [x] `make lint-check` — clean
- [x] `make test-ci` (unit + integration + scripts) — 668 passed
- [x] `flask db upgrade` reaches single head `b3f4a9c1d7e2`; `flask db check` reports no drift (verified it *does* fail loudly by deliberately introducing an unmigrated column, then reverted)
- [x] `flask search compare` / `reset-mapping` exercised against real Postgres + OpenSearch
- [x] confirmed no remaining `datagov_data_access` references in code, lockfile, or exported `requirements.txt`
- [ ] confirm `datagov_data_access`'s own `notify-consumers.yml` drops `datagov-harvester` from its dispatch matrix (separate repo, coordination note only)

Related: GSA/data.gov#6209, GSA/datagov-harvester#797


Aguiardavidm:
I put all of the copied code into commit [29a538b](https://github.com/GSA/datagov-harvester/pull/807/changes/29a538b0c1d75db89bca97fc1fd8f93e1da13ba3) and all of the new code in the following  commits especially [2855a32](https://github.com/GSA/datagov-harvester/pull/807/changes/2855a32a717ab5f51334cf2707ee1ccbf6889750). 

For review I would look at this from a commit level.

.github/workflows/commit.yml
```diff
@@ -70,6 +70,12 @@ jobs:
       - name: Run Pytest
         run: make test-ci
 
+      - name: Verify database/interface.py imports standalone
+        run: poetry run python -c "import database.interface"
```
cody-seibert-gsa, line 74: not sure where these came from or their use case.

cody-seibert-gsa (approved):

cody-seibert-gsa:
deployed to develop and it all passed https://github.com/GSA/datagov-harvester/actions/runs/30925968474, also we ran a few harvest jobs and they also passed on develop:

Screenshot 2026-08-04 at 11 55 12 AM

https://harvest-dev.data.gov/harvest_job/cad2a77b-abd9-46a3-894b-f47071acfce6
https://harvest-dev.data.gov/harvest_job/7f966e25-429c-416c-b423-1ce958521d7d

rshewitt (approved):
looks like you brought everything. hopefully we find the decomposition of the model tests, os client, os reader, os writer, etc... easier to maintain over time

PR #809, opened 2026-08-03
6210 - Run catalog's tests against harvester's schema on every PR

cody-seibert-gsa:
# Pull Request

Related to https://github.com/GSA/data.gov/issues/6210

> [!IMPORTANT]
> **Merge GSA/datagov-catalog#375 first.** This job pulls `ghcr.io/gsa/datagov-catalog:main`, which that PR's workflow publishes. Until it merges *and* the GHCR package is made public (a one-time manual step described there), the `Catalog Contract` job here will fail to pull the image.

## About

Harvester owns the Postgres schema (`migrations/`), the OpenSearch index mapping (`search/mappings.py`), and the indexed document shape (`search/documents.py`). Catalog reads all three and owns none of them — it has no migrations of its own. So a harvester change to any of them can break catalog, and until now nothing caught it before deploy.

This adds a `Catalog Contract` workflow that runs on every PR: it provisions Postgres and OpenSearch with **harvester's own** migrations and mapping, then runs **catalog's** test suite against them.

### Why the overlay exists

Both repos have now vendored the same `datagov_data_access` 1.1.0 code and dropped the dependency — harvester into `database/` + `search/` (#807), catalog into `app/models.py` + `app/search/` (catalog#376). **There is no longer a shared package anywhere.** Two independent copies, currently identical apart from import prefixes, with nothing enforcing that.

So before running pytest, the job overlays harvester's vendored `search/` modules onto catalog's `app/search/` and rewrites the import prefixes. Without it the test is a closed loop: catalog's own copy of the writer would produce the documents catalog then reads, so a harvester-side change to `MAPPINGS`, `DatasetDocument`, the writer, or the filter registry would go completely undetected.

Two deliberate limits:

- **Copies module-for-module, doesn't replace the tree.** Catalog's `app/search/` owns files harvester has no counterpart for (`__init__.py`'s public façade, `url_helpers.py`); clobbering them breaks catalog's imports. If harvester ever gains a module catalog lacks, the overlay **fails loudly** rather than silently skipping it. It also greps for unrewritten paths and fails hard, so a silent fallback to catalog's own copy is impossible.
- **`database/models.py` is not overlaid.** Catalog's fixtures must exercise the schema harvester's migrations actually built; substituting harvester's model definitions would make that check tautological. `flask db check` (added by #807) covers the models↔migrations direction.

### Relationship to existing guards

Additive, not duplicative. #807's guards are intra-repo (models↔migrations, importability). Catalog#376 added a **non-blocking** `continue-on-error` text-diff of harvester's `search/mappings.py` whose own commit message says *"the actual fix is a shared schema artifact or harvester-side contract tests (GSA/data.gov#6210). This just makes drift visible instead of silent."* This is that blocking version, and it covers the whole module set plus the schema rather than just the mapping file.

### What it proves

1. **Schema contract** — catalog's queries work against the schema `flask db upgrade` produces.
2. **Mapping contract** — catalog's searches, aggregations, and filters work against the index `flask search reset-mapping` creates.
3. **Document-shape contract** — catalog's fixtures index through the `OpenSearchWriter`, exercising `DatasetDocument` writer → index → reader → template.

Ordering is load-bearing and enforced by the target chain: catalog's `create_app()` connects to OpenSearch at import and will create the index with *its* mapping if provisioning hasn't run, which would quietly defeat the test. Migrations also terminate other database backends, so the steps must not overlap.

The stack uses its own compose project name and publishes **no host ports**, so it never collides with `make up`.

### It already found two real bugs

Running this against catalog exposed two incompatibilities that catalog#376's slimming had introduced, both invisible against catalog's own FK-less `create_all()` schema. Both are fixed on the catalog side (`EnumAsString` for the psycopg `::VARCHAR` bind cast against real enum columns, and parent-before-child flushes for real FKs). Catalog#376 had also already fixed a third of the same class — the `harvest_job` FK — explicitly citing this PR.

### Also fixes a latent bug in `flask search reset-mapping`

OpenSearch echoes `dynamic` back as the string `"false"` while the mapping declares a bool, so the command's own verification step **always failed against a real cluster** — it deleted and correctly recreated the index, then reported `Created index mapping does not match application mapping.` and exited non-zero.

Vendoring made this live rather than latent: `search/mappings.py` is now in-repo with `"dynamic": False`. The two existing tests mock `client.MAPPINGS` with only `type`/`analyzer` keys, so they never exercised the round trip. Also affects the scheduled `synchronize_opensearch_index.yml`, independent of #6210. Fixed in `app/commands/search.py` (5 lines) with a regression test using the real mapping shape.

Catalog's own docs already reference this bug as a known papercut when regenerating their harvester snapshot — that goes away with this fix.

### Also bumps cryptography

Snyk began failing on this branch with no dependency change of its own: it passed on the previous head and fails on every open PR that doesn't bump `cryptography`, matching GSA/inventory-app@2076f03. Raised past the old `<50.0.0` cap so 50.0.0 resolves; the lock diff is confined to that one package.

`pyOpenSSL` is deliberately **not** pinned alongside it as it was for inventory — harvester doesn't depend on it. The only occurrence in `poetry.lock` is an unselected `pyopenssl` extras marker on `google-auth`.

## Verification

Full local run against rebased `main` and a locally built catalog image:

| Check | Result |
| --- | --- |
| Full clean contract run | **261 passed in 15.7s**; 1m20s end-to-end |
| Schema preserved | `alembic_version` at the migration head after the run; tables truncated, not dropped |
| Overlay applied | log confirms harvester's modules overlaid onto `/app/app/search` |
| Harvester `tests/unit` | 343 passed (2 pre-existing `test_compare_update_*` failures, unchanged from `main`) |
| `make lint-check` | clean |
| All CI on this PR | green, Snyk included |

Deliberate-break checks, as the issue requests:

| Break | Result |
| --- | --- |
| Edit harvester's `search/mappings.py` (`spatial_centroid` → `geo_shape`) | 20 failed ✅ |
| Drop `dataset.popularity` (a column catalog reads) | 4 failed, 69 errors ✅ |
| Replace the OpenSearch mapping with a stub | 39 failed ✅ |
| Unset `CATALOG_TEST_EXTERNAL_SCHEMA` | **261 pass despite a dropped column** ✅ — confirms the env gate is load-bearing |

## Running it locally

```bash
make test-catalog-contract
```

Against a specific catalog commit or a local build:

```bash
CATALOG_IMAGE=ghcr.io/gsa/datagov-catalog:sha-<full-sha> make test-catalog-contract
```

See [docs/catalog-contract-test.md](docs/catalog-contract-test.md) for the full rationale and failure-mode guide.

## Follow-up after merge

Per the issue's final acceptance step, add `Catalog Contract` to this repo's branch protection **required checks** so failures are a hard stop. That's a repo settings change, not a code change.

One known caveat: the job tests against whatever `:main` currently points at, so a catalog merge mid-flight changes what a given harvester PR tested. `workflow_dispatch` accepts a `catalog_image` input for pinning a specific `sha-` tag, though it won't appear in the UI until this file reaches the default branch.

Once the two vendored copies are reconciled for real (a shared artifact, or catalog importing harvester's), the overlay in `docker-compose.catalog-contract.yml` can be deleted and this becomes a plain "run catalog's tests" job.

## PR TASKS

- [x] Code well documented
- [x] Files linted


docker-compose.catalog-contract.yml
```diff
@@ -0,0 +1,109 @@
...
+        set -e
+        echo "Image ships datagov-data-access: $$(python -c 'import importlib.metadata as m; print(m.version("datagov-data-access"))')"
+        echo "Harvester pins datagov-data-access: $${DATA_ACCESS_REF}"
+        # Run catalog's code against harvester's data_access, not its own pin.
+        # Without this the test is a closed loop: catalog's OpenSearchWriter would
+        # write the documents catalog then reads, so a harvester-side change to
+        # models / MAPPINGS / DatasetDocument would go undetected.
+        # A tarball (not git+) because the Alpine image has no git binary.
+        pip install --no-deps --force-reinstall --quiet \
+          "datagov-data-access @ https://github.com/GSA/datagov_data_access/archive/refs/tags/$${DATA_ACCESS_REF}.tar.gz"
```
cody-seibert-gsa: we need to determine which one gets merged first, if we manage to merge https://github.com/GSA/datagov-harvester/pull/807, then we can refactor this away, but if not, we will probably need to do rework after merging 807

jaredb96 (changes requested):
This is causing a snyk issue that I actually fixed for inventory today: 
https://github.com/GSA/inventory-app/pull/949/changes/0ace8f69c05b72e56eccd7811ad612074b2604e6
https://github.com/GSA/inventory-app/pull/949/changes/2076f030838c80428672def879a8b0c007bce022

Set `cryptography==50.0.0` and `pyOpenSSL==26.4.0` and that should fix snyk. Otherwise this LGTM.

jaredb96 (dismissed):
LGTM

jaredb96 (dismissed):

cody-seibert-gsa:
as a test, I commited https://github.com/GSA/datagov-harvester/actions/runs/31028127026/job/92381692785 which demos the contract tests failing after changing the mapping

akuny (approved):

PR #817, opened 2026-08-05
Fix job record error title

rshewitt:
# Pull Request

Related to [6183](https://github.com/GSA/data.gov/issues/6183)

## About
fix issue surrounding "source_raw" being expected for pulling dcatus titles in harvest record errors on job pages. this update now supports using source_transform instead of source_raw so the exception isn't thrown as often. deploying on dev now.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM

PR #813, opened 2026-08-05
6175: add has_download filter for datasets without distribution links

Aguiardavidm:
## Summary
- Ports [GSA/datagov_data_access#13](https://github.com/GSA/datagov_data_access/pull/13) into harvester's vendored `search/` package now that `datagov_data_access` was deprecated (never made it into the vendor snapshot before deprecation)
- Adds a `has_download` boolean field to the OpenSearch dataset document/mapping, computed from `dcat.distribution[].downloadURL` (DCAT-US 1.1/3.0 compatible; `accessURL`-only distributions do not count, per DCAT-US 3.0 semantics where `accessURL` is explicitly indirect access)
- Registers a new opt-in `has_download` search filter (single checkbox, off by default) rendered via the existing generic `checkbox_group` sidebar renderer

Addresses GSA/data.gov#6175. Companion PR in datagov-catalog: (link once opened).

## Test plan
- [x] `poetry run pytest tests/unit/search/test_search_filter_registry.py tests/unit/search/test_mappings.py tests/integration/search/test_documents.py` (74 passed)
- [x] `poetry run ruff check` / `poetry run black --check` on changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #819, opened 2026-08-06
Make catalog contract test one-directional, drop search/ tree parity

Aguiardavidm:
## Summary
Per discussion with Cody Seibert and James Brown on harvester #813: the "Catalog Contract" job was checking that harvester's and catalog's vendored `search/` trees match file-for-file, which is the wrong invariant. That overlay+parity check:

- Penalized pure additions (a new filter module) exactly like a breaking/removed one, since it can't tell the two apart.
- Forced the two repos' releases into lockstep — harvester couldn't ship a `search/` addition until catalog had already merged and shipped a matching file to `main`, even when catalog's current behavior wasn't at risk.

This PR removes the overlay and file-parity check entirely. The job now runs **catalog's own, unmodified code and test suite** against the Postgres schema and OpenSearch mapping a harvester PR is proposing — i.e. "does catalog's code, as deployed today, still work against this harvester change," not "do the two repos' internal `search/` copies match."

## What's unchanged
- Schema contract (catalog's queries work against `flask db upgrade`'s output) and mapping contract (catalog's searches/filters work against `flask search reset-mapping`'s index) are still checked, and still catch a bad migration or breaking mapping change.

## What's removed
- The `search/` → `app/search/` file overlay, import-prefix rewrite, and the hard failure when harvester has a `search/` module catalog lacks. The two vendored trees may now diverge; `search/` drift between the repos is visible via catalog's existing non-blocking text-diff in `commit.yml`, not blocking here.

## Verification
Ran `make test-catalog-contract` locally against harvester's `has_download.py` filter addition (the change that originally triggered the false failure on #813) — catalog's `:main` image doesn't have a matching file, and all 265 of catalog's `tests/unit` pass unmodified.

### DIFF
--- a/.github/workflows/catalog-contract.yml
+++ b/.github/workflows/catalog-contract.yml
@@ -1,8 +1,10 @@
 name: Catalog Contract
 
 # Verifies that harvester changes to the Postgres schema or the OpenSearch mapping
-# don't break catalog, which reads both and owns neither. Provisions services with
-# harvester's migrations and mapping, then runs catalog's test suite against them.
+# don't break catalog's own, unmodified code, which reads both and owns neither.
+# Provisions services with harvester's migrations and mapping, then runs catalog's
+# real test suite against them -- catalog's search/ code is not touched, and the
+# two repos' vendored search/ trees are not required to match file-for-file.
 # See GSA/data.gov#6210 and docs/catalog-contract-test.md.
 #
 # No branch filter: harvester PRs target both main and develop, and DB changes tend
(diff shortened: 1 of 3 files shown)


akuny (approved):
Makes sense to me, thank you for addressing this.

Aguiardavidm:
Superseded by #820, which makes the same functional change (drop the search/ overlay and file-parity check, run catalog's tests/unit directly against harvester's provisioned schema/mapping) but targets `main` directly and has more thorough verification (confirmed against catalog's actual deployed SHA) plus a clearer writeup of the tradeoffs and a dual-support release sequence. Closing this one in favor of that.

PR #821, opened 2026-08-06
handle no tranformation exception

rshewitt:
# Pull Request

Related to [6183](https://github.com/GSA/data.gov/issues/6183)

## About
made an initial improvement but noticed JSON parsing exceptions were still occurring after merge. this handles transformation exceptions where source_transform is none so source_raw is used. the json parse error is because `json.loads("<xml>value</xml")`

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


akuny (approved):
LGTM

PR #823, opened 2026-08-07
Revert "6210 - Run catalog's tests against harvester's schema on every PR"

cody-seibert-gsa:
Reverts GSA/datagov-harvester#809


tests/unit/test_search_commands.py
```diff
@@ -37,43 +37,6 @@ def test_reset_mapping_recreates_empty_index(app):
...
-    }
-
-    with patch(
-        "app.commands.search.OpenSearchClient.from_environment",
-        return_value=client,
-    ):
-        result = app.test_cli_runner().invoke(args=["search", "reset-mapping"])
-
-    assert result.exit_code == 0
-    assert "Mapping reset successfully. The index is empty." in result.output
```
jbrown-xentity, line 74: I think this test is valid without the catalog-contract file, but I may be wrong...

jbrown-xentity (approved):
LGTM. @Aguiardavidm @cody-seibert-gsa cane we evaluate the test below for re-inclusion in the future? Should it be maintained or not?

PR #824, opened 2026-08-07
Restore dynamic-flag mapping normalization dropped by revert-809

Aguiardavidm:
## Summary
- Restores the `dynamic` bool-vs-string normalization in `_normalize_mapping_for_comparison` (`app/commands/search.py`) that PR #823 accidentally removed along with its revert of the catalog-contract-test feature.
- OpenSearch echoes the mapping's `dynamic` flag back as the string `"false"`, while `search/mappings.py` declares it as a Python bool — without normalization, `reset-mapping` and `rebuild-opensearch-index` both false-positive on "Created index mapping does not match application mapping."
- This was actively breaking the `search-rebuild` GitHub Action in production.

## Test plan
- [x] `pytest tests/unit/test_search_commands.py` — 17/17 passed, including restored `test_reset_mapping_accepts_stringified_dynamic_flag`

### DIFF
--- a/app/commands/search.py
+++ b/app/commands/search.py
@@ -53,6 +53,11 @@ def _normalize_mapping_for_comparison(value):
             "search_analyzer"
         ) == normalized.get("analyzer"):
             normalized.pop("search_analyzer")
+        # OpenSearch echoes `dynamic` back as a string ("false"), while the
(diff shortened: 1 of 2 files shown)


akuny (approved):
LGTM

PR #825, opened 2026-08-07
6130: Incorporate Warnings in API

### LINKED ISSUE #6130: Handle DCAT-US v3.0 Warnings in Harvester API
There are several API routes that will need to be updated to account for the new warnings.

 Rough layout of changes:
  - Update API response schemas to include records_warned in job summary
  - Allow for severity filtering when applicable

akuny:
Closes [6130](https://github.com/GSA/data.gov/issues/6130), part of epic [5970](https://github.com/GSA/data.gov/issues/5970).

DCAT-US v3.0 warnings are already detected, written, and counted: #6126 added `harvest_record_error.severity`, #6127 added the detection engine, #6128 wired it into `Record.validate`. The JSON API didn't describe warnings or let anyone filter on them. This PR closes that gap.

`records_warned` and `severity` were already in every JSON response, since `HarvesterDBInterface._to_dict` reflects over all mapper columns. They were just missing from `/openapi.json`. `/api/harvest_record_errors/` returned errors and warnings mixed together under a name that promises errors only. Most of this diff documents behavior that already existed. One commit, `39ce0b9e`, changes behavior, and it's isolated so it can be reverted alone.

## Changes

- `records_warned` on `JobInfo`, `severity` on `ErrorInfo`, both now in the published spec.
- A `severity` query param (`error` | `warning`) on both record-issue read routes, documented on both.
- `/api/harvest_record_errors/` defaults to errors only.
- `/api/harvest_record/<id>/errors` now accepts `severity`. The interface function has supported it since #6126; the route never passed it through.

No interface-layer changes. Every DB function already took the parameter.

## Decisions

Judgment calls made, called out so they can be challenged individually if necessary.

### 1. Errors-only default on `/api/harvest_record_errors/` is a breaking change

This route never filtered on severity before. `pget_harvest_record_errors`, which defaults to `severity="error"`, was never called from it; the route went straight to `pget_db_query`. So #6128 didn't regress anything, it just introduced the first `severity="warning"` rows, which exposed a query that was always under-specified.

I'm changing the default to errors-only because #6126 already established that pattern everywhere else, and named the all-issues accessor `get_harvest_record_issues` specifically so an "errors" function would never return warnings. Warning rows currently only exist for `dcatus3.0` sources, so this is the cheapest point to fix the naming mismatch.

Alternative: keep returning everything by default and treat `severity` as additive. Non-breaking, but leaves the naming wrong. Reverting `39ce0b9e` accomplishes this.

### 2. An explicit `severity` facet overrides the default instead of ANDing with it

Precedence: a `severity` facet wins outright; otherwise `?severity=` applies; otherwise errors only.

The naive approach appends `severity eq error` to whatever `facets` the caller supplied. That silently breaks `?facets=severity eq warning`: two contradictory conditions get ANDed and the response is always empty. I hit this bug first and two tests caught it. The fix is a small helper, `_facets_on_severity`, which parses facets the same way `query_filter_builder` does, so it stays in sync with what actually reaches the query.

Alternative: let them AND and document the contradiction. Simpler, but a documented feature would quietly return nothing.

### 3. `severity` takes exactly `error` or `warning`, and anything else is a 400

No `all` sentinel, since that would put a vocabulary in the API that doesn't exist in the database; to get everything in one request, facet on severity or query twice. Invalid values are rejected rather than silently falling back to the default: they reach `query_filter_builder`, which builds filters from a string split on spaces and commas, so passing unvalidated input through is both a broken query and unnecessary injection surface. The check is case-sensitive, so `?severity=Warning` is a 400, not a silent zero-result match. On `/api/harvest_record/<id>/errors` the validation sits outside the route's existing broad `try`, or the 400 would get swallowed and re-emitted as a 404.

Alternative: add an `all` value, or ignore unknown values and fall back to the default. Both are friendlier but hide typos or add API surface the database doesn't back.

## Deliberately out of scope

Left for #6129 and PR #799:

- **CSV export** (`/api/harvest_job/<id>/errors/<type>`). Until #799 lands, the CSV keeps exporting warnings unlabeled, as if they were errors. #799 already adds a severity column there.
- **`get_record_errors_summary_by_job`** groups by `type` only, so it mixes warning types into the HTML job page's error summary. Rendering concern, not an API route.
- **`for_view` positional-tuple contract** in `app/util.py`. Untouched, which is what keeps this diff safe.
- Templates, the notification email, the source-history chart.

## PR Tasks

- [x] Code well documented
- [x] Files linted


akuny:
Closing, will reconcile with #799

PR #826, opened 2026-08-07
6130: Warnings in API

### LINKED ISSUE #6130: Handle DCAT-US v3.0 Warnings in Harvester API

akuny:
Closes [6130](https://github.com/GSA/data.gov/issues/6130), part of epic [5970](https://github.com/GSA/data.gov/issues/5970).
 
DCAT-US v3.0 warnings are already detected, written, and counted (#6126 added `harvest_record_error.severity`, #6127 added the detection engine, #6128 wired it into `Record.validate`). The JSON API never described warnings or let anyone filter on them. This PR closes that gap.
 
`records_warned` and `severity` were already in every JSON response, since `HarvesterDBInterface._to_dict` reflects over all mapper columns. They just weren't in `/openapi.json`, so they looked unsupported. Most of this diff documents existing behavior, plus adds a filter.
 
### Changes
 
- `records_warned` on `JobInfo`, `severity` on `ErrorInfo`, both now in the published spec.
- A `severity` query param (`error` | `warning`) on both record-issue read routes, documented on both.
- `/api/harvest_record/<id>/errors` couldn't reach warnings before. The interface function accepted `severity` since #6126; the route never passed it.
No interface-layer changes. Every DB function already took the parameter.
 
### Both routes follow #799's convention
 
With no `severity` param, both return every issue and let callers distinguish via the `severity` field on each row. This matches what PR #799 established at the job level: `get_harvest_record_errors_by_job` defaults to all issues, and #799 surfaced `severity` through the view rows, CSV export, `make_new_record_error_contract`, the job template, and `get_record_errors_summary_by_job` (now grouped by severity).
 
An earlier revision defaulted these routes to errors-only, on the theory that an endpoint named `harvest_record_errors` shouldn't serve warnings. That predated #799. Keeping that default while #799 surfaces severity everywhere else would leave two opposite conventions in the same epic, so this branch follows #799 instead. The errors-only argument is in the history for anyone who wants to revisit it, but consistency won.
 
One asymmetry to note: `get_harvest_record_errors_by_record` and `pget_harvest_record_errors` still default to `severity="error"` at the interface layer (#799 only flipped the job-level reads). Both routes therefore pass `severity=None` explicitly. That's deliberate; `test_record_route_defaults_to_all_issues` guards it, since dropping the explicit `None` would silently reintroduce errors-only filtering.
 
### 1. Invalid severity is a 400, not a silent fallback
 
**Decision:** `?severity=bogus` returns `{"error": "Invalid severity. Must be one of: error, warning."}`.
 
Values reach `query_filter_builder`, which builds filters from a string split on spaces and commas. Passing an unvalidated value through is both a broken query and needless injection surface. It's also case-sensitive: the DB enum is lowercase, so `?severity=Warning` is a 400, not a silent zero-result match.
 
The message deliberately doesn't echo the submitted value (the CodeQL "information exposure through an exception" finding, fixed in `128c90e7`). `test_collection_invalid_severity` asserts the value is absent from the response, so the test would catch a regression, not just pass by omission.
 
On `/api/harvest_record/<id>/errors`, validation sits outside that function's existing broad `try`, or the 400 would be swallowed and re-emitted as a 404.
 
**Alternative:** ignore unknown values and fall back to no filter. Friendlier, but hides typos behind plausible-looking results.
 
### 2. Shared `ErrorInfo` schema left as-is
 
Job errors and record errors are separate tables sharing one `ErrorInfo` schema in the spec, so `ErrorInfo` now advertises `severity` even though only `harvest_record_error` has that column. Job-error responses just omit the field.
 
Splitting the schema would describe this honestly, but adds a schema definition, which `tests/playwright/test_openapi.py` pins by count. Documented as a known wart in `docs/routes.md` Notes instead.
 
**Alternative:** split it and bump the assertion. Cleaner contract, larger diff, and it drags the playwright test into an API-docs PR.
 
## PR TASKS
 
- [x] Code well documented
- [x] Files linted


tests/integration/app/test_routes.py
```diff
@@ -837,3 +837,150 @@ def test_autoescape_disabled_raises_exception(
...
+
+        Job-level reads default to every issue and surface `severity` on each
+        row rather than filtering; this endpoint matches that. The severity of
+        each row is in the response, so nothing is ambiguous.
+        """
+        res = client.get("/api/harvest_record_errors/?paginate=False")
+
+        assert res.status_code == 200
+        assert {e["severity"] for e in res.json} == {"error", "warning"}
+        assert len(res.json) == 17
```
jbrown-xentity, line 921: Not sure I love hardcoded values like this; this probably relates to other input data and test documents and will drift.

tests/integration/app/test_routes.py
```diff
@@ -837,3 +837,150 @@ def test_autoescape_disabled_raises_exception(
...
+        self, client, interface_with_warning
+    ):
+        """?severity=error excludes warnings."""
+        res = client.get(
+            "/api/harvest_record_errors/?severity=error&paginate=False&count=True"
+        )
+
+        assert res.status_code == 200
+        # 16 seeded errors; the added warning is filtered out
+        assert res.json["count"] == 16
```
jbrown-xentity, line 956: Same as above comment. I like the test on line 899, it makes sure everything is a warning without checking a specific count.

jbrown-xentity (dismissed):
LGTM. Small quibbles about test cases, but good to move forward.

jbrown-xentity (approved):
LGTM

PR #827, opened 2026-08-07
Add /api/v1 URI versioning for harvest admin API

Aguiardavidm:
## Summary
- Adds URI versioning to the harvest admin API: the `api` blueprint is now also registered under `/api/v1`, so every existing route is reachable at both the un-prefixed `/api/...` path and the versioned `/api/v1/...` path with identical behavior.
- Unversioned `/api/...` paths stay live as an alias so existing consumers (e.g. CGI Federal) aren't broken.
- Updates `docs/routes.md` to document the new alias.
- Fixes the dynamic route-table test's endpoint whitelist regex to recognize the new `api_v1` blueprint name alongside `main`/`api`.

Addresses [GSA/data.gov#5128](https://github.com/GSA/data.gov/issues/5128).

## Test plan
- [x] `poetry run pytest tests/unit` — 347 passed
- [x] `poetry run pytest tests/integration/app/` — 97 passed (pre-existing, unrelated `test_load_manager.py` failures reproduce identically on `main`)
- [x] Manually verified against the running app: `/api/organizations/` and `/api/v1/organizations/` both return 200 with identical data, both appear as separate paths in `/openapi.json`, and auth-required routes correctly 401 under `/api/v1/...` too


cody-seibert-gsa (approved):

tests/playwright/test_openapi.py
```diff
@@ -12,5 +12,5 @@ class TestOpenAPI:
     def test_swagger(self, page):
         expect(page.locator("h2.title")).to_have_text("Datagov Harvester 0.1.0 OAS 3.1")
         expect(page.locator("h3")).to_have_text("Api")
-        expect(page.locator(".opblock-get")).to_have_count(15)
+        expect(page.locator(".opblock-get")).to_have_count(30)
```
rshewitt, line 15: doubled because we're duplicating routes by versioning them (e.g. `/api/harvest_source` + `/api/v1/harvest_source`)

PR #829, opened 2026-08-10
6181: Support harvesting DCAT-US3 CatalogRecord objects

### LINKED ISSUE #6181: Support harvesting DCAT-US3 CatalogRecord objects
## User Story

In order to harvest DCAT-US3.0 metadata from non-dataset records, OMB wants CatalogRecord objects in the [record field](https://resources.data.gov/standards/catalog/dcat-us-3/catalog/#record) supported end-to-end in the harvest pipeline.

## Acceptance Criteria

- [ ] GIVEN a valid DCAT-US3 CatalogRecord example is present in the source metadata, WHEN a harvest runs, THEN CatalogRecord objects are detected and persisted in the harvest database.

## Background

This story splits out the CatalogRecord portion of #6000. The documentation of what a catalogRecord looks like is [here](https://resources.data.gov/standards/catalog/dcat-us-3/quality-governance/#catalog-record).
At this time there are no plans to make this data searchable, however it remains to be seen how data providers will use this metadata. In order to limit the catalog object storage and make sure everything is captured, we want to make sure any objects are tracked and maintained so we could use them in the future.

Reference material:
- #6000
- https://github.com/GSA/dcat-us/blob/main/jsonschema/examples/Catalog/good/complete_example.json

None.

## Sketch

- Add processing for the list of objects in the `record` field of the catalog object
  - This processing should track objects for adds, updates, and deletes
  - This processing should assume `@id` is present even though it is optional, and throw an error if it doesn't exist.
  - This should go into the harvestRecord table (https://github.com/GSA/datagov_data_access/blob/main/datagov_data_access/db/models.py#L207), with the new enum of the catalogRecord type detailed in https://github.com/GSA/data.gov/issues/6000#issuecomment-5107673275

This should not cover making things viewable or discoverable in openSearch. This may mean that we need to update the Dataset table updates or other processes to exclude these record types for now.

jaredb96:
PR for [Issue #6181](https://github.com/GSA/data.gov/issues/6181). With DCAT-US 3.0 we're adding / modifying database models and interfaces to support storing CatalogRecord objects from DCAT-US3.0 catalogs. These new objects have to be added to `HarvestRecord` and persisted in the harvester db during harvesting.


jaredb96:
Work will be done here: https://github.com/GSA/datagov-harvester/pull/835

PR #830, opened 2026-08-10
add normalized theme keyword for exact matching

rshewitt:
# Pull Request

Related to [6161](https://github.com/GSA/data.gov/issues/6161)
Related [catalog PR](https://github.com/GSA/datagov-catalog/pull/386)

## About
updates opensearch mapping "theme" to add exact keyword matching. search document transformation already supports writing "theme" so this should just be a mapping update.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jbrown-xentity (approved):
LGTM. Would love for this to use new index 0 down time deployment, might wait until then to deploy.

PR #831, opened 2026-08-10
6178 Support harvesting DCAT-US3 DataService objects

Aguiardavidm:
## Summary
- Adds a `record_type` column to `harvest_record` (dataset / data_service / data_series / catalog_record), folded into the "latest successful records" index and query so a Dataset and a DataService sharing an identifier don't collide. `data_series`/`catalog_record` are defined now but unused, to avoid a second enum migration when GSA/data.gov#6180 (data_series) and #6181 (catalog_record) land.
- Extracts DataService objects from a DCAT-US 3.0 catalog's `service` array (mirrors the existing dataset extractor), validates each against the existing `DataService.json` schema, and persists them as `HarvestRecord`s — independently of `Dataset` objects in the same source. DataService records don't get a `Dataset` table row; there's no downstream consumer for that yet, so persistence stops at the HarvestRecord level per the parent ticket's scoping.
- `internal_records` is now keyed by `(identifier, record_type)` throughout the harvest pipeline (compare, dedup, deletions) instead of identifier alone.

Addresses [GSA/data.gov#6178](https://github.com/GSA/data.gov/issues/6178), a sub-ticket of [GSA/data.gov#6000](https://github.com/GSA/data.gov/issues/6000).

## Commits
1. Schema: `record_type` column + index
2. DB query updates: dedup/scope by `record_type`
3. Harvest pipeline: DataService extraction, validation, persistence
4. New DCAT-US 3.0 DataService fixture data
5. Tests for DataService extraction/validation/persistence
6. Fixes to pre-existing tests broken by the new column/field
7. Unrelated cleanup: generalize repeated harvest-source test setup into a shared fixture (`tests/integration/harvest/conftest.py`)

## Test plan
- [x] Full `pytest tests/unit tests/integration` — 666 passed; the 12 failures present are pre-existing and unrelated (confirmed via `git stash` against `main`, before this branch's changes)
- [x] Manually ran migration up/down locally against a fresh dev DB

### DIFF
--- a/harvester/utils/general_utils.py
+++ b/harvester/utils/general_utils.py
@@ -471,17 +471,35 @@ def strip_dcatus3_catalog_objects(catalog: dict) -> dict:
     return cleaned
 
 
+def _extract_dcatus3_catalog_objects(catalog: dict, field: str) -> list:
+    """
+    recursively collect every entry of [field] ("dataset" or "service") from a
+    DCAT-US3 Catalog dict, including entries nested arbitrarily deep within
+    its "catalog" (sub-catalog) field.
+    """
+    objects = list(catalog.get(field) or [])
+
+    for sub_catalog in catalog.get("catalog") or []:
+        objects.extend(_extract_dcatus3_catalog_objects(sub_catalog, field))
+
+    return objects
+
+
 def extract_dcatus3_catalog_datasets(catalog: dict) -> list:
     """
     recursively collect every dataset from a DCAT-US3 Catalog dict, including
     datasets nested arbitrarily deep within its "catalog" (sub-catalog) field.
     """
-    datasets = list(catalog.get("dataset") or [])
+    return _extract_dcatus3_catalog_objects(catalog, "dataset")
 
-    for sub_catalog in catalog.get("catalog") or []:
-        datasets.extend(extract_dcatus3_catalog_datasets(sub_catalog))
 
-    return datasets
+def extract_dcatus3_catalog_services(catalog: dict) -> list:
+    """
+    recursively collect every DataService from a DCAT-US3 Catalog dict,
+    including services nested arbitrarily deep within its "catalog"
+    (sub-catalog) field.
+    """
+    return _extract_dcatus3_catalog_objects(catalog, "service")
 
 
 def make_record_mapping(record):
@@ -496,6 +514,7 @@ def make_record_mapping(record):
         "action": record.action,
         "ckan_id": record.ckan_id,
         "parent_identifier": record.parent_identifier,
+        "record_type": record.record_type,
     }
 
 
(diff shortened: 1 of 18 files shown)


harvester/utils/general_utils.py
```diff
@@ -471,17 +471,35 @@ def strip_dcatus3_catalog_objects(catalog: dict) -> dict:
     return cleaned
 
 
+def _extract_dcatus3_catalog_objects(catalog: dict, field: str) -> list:
```
cody-seibert-gsa, line 474: feel free to ignore this comment, but it might be worth adding some type of max depth to this recursive call in case a dataset provider has a really recursively nested object and this causes call stack exceeded errors and crashes harvester, but honestly I don't think this will be an issue.

cody-seibert-gsa (dismissed):

harvester/utils/general_utils.py
cody-seibert-gsa: Python's default recursion limit is ~1000, so yeah this is probably fine honestly

jbrown-xentity (dismissed):
This looks good to me.

harvester/utils/general_utils.py
```diff
@@ -471,17 +471,35 @@ def strip_dcatus3_catalog_objects(catalog: dict) -> dict:
...
-        datasets.extend(extract_dcatus3_catalog_datasets(sub_catalog))
 
-    return datasets
+def extract_dcatus3_catalog_services(catalog: dict) -> list:
+    """
+    recursively collect every DataService from a DCAT-US3 Catalog dict,
+    including services nested arbitrarily deep within its "catalog"
+    (sub-catalog) field.
+    """
+    return _extract_dcatus3_catalog_objects(catalog, "service")
```
jbrown-xentity, line 502: Actually recognizing a problem. This pulls out the dataService's. But dataService can house it's own list of datasets: https://resources.data.gov/standards/catalog/dcat-us-3/quality-governance/#data-service--servesDataset. So we need to loop through those, sending that array to be extracted just like datasets, but with a way to track the relationship of the dataService to the dataset (tracking on the dataset makes sense to me).
For datasetSeries there is something similar, so might want to think about this as a repeatable process, not just a one-off for dataService.

Aguiardavidm: Good catch, thanks. Fixed in 6845c45a: added `extract_dcatus3_nested_datasets`, a generic helper that pulls Dataset objects out of one or more fields on a list of parent objects (here `DataService.servesDataset`), tagging each with `parent_identifier` set to the parent's own identifier so the relationship survives once the dataset is harvested on its own via the existing `HarvestRecord.parent_identifier` column (same mechanism already used for WAF-collection parents).

Wrote it generically over a list of fields (`extract_dcatus3_nested_datasets(parents, *fields)`) specifically because DatasetSeries has the identical shape via `seriesMember`/`first`/`last` — that's coming in #6179 and will reuse this same helper.

Datasets served this way get merged into the normal `self.external_records` pipeline, so they're validated and get a real `Dataset` row like any other dataset, rather than staying nested inside their DataService.

jbrown-xentity (approved):
LGTM, reviewed latest changes and all looks good. Ready to merge.

PR #835, opened 2026-08-11
Support harvesting DCAT-US3 CatalogRecord objects

Aguiardavidm:
## Summary

Stacked on #831 (#6178, DataService). Generalizes that PR's DataService-specific handling into a registry (`NON_DATASET_RECORD_TYPES`) so each additional DCAT-US 3.0 record type is one entry instead of a duplicated field/loop, then adds CatalogRecord as the second entry.

- Extracts CatalogRecord objects from a DCAT-US 3.0 catalog's `record` array (mirrors the existing dataset/service extractors), validates each against the existing `CatalogRecord.json` schema, and persists them as `HarvestRecord`s — independently of Dataset objects in the same source. Same persistence scope as DataService: no downstream table yet, per this ticket's note that OpenSearch/display integration isn't planned for CatalogRecord.
- CatalogRecord has no `identifier` field at all — only an optional top-level `@id`. Per the ticket's sketch ("assume `@id` is present even though it is optional, and throw an error if it doesn't exist"), the identifier-reading helpers (`find_indexes_for_duplicates`, `describe_identifier_error`, and the dispatch throughout `HarvestSource`) now take the identifier field name as a parameter instead of hardcoding `"identifier"`.
- Replaced the DataService-only `external_service_records` field and `.validator`/`.dataservice_validator` properties with a generic `external_records_by_type` dict and a `validator_for(record_type)` method, so the pattern scales to the next type (`data_series`, #6179) without another copy-pasted field/loop.

No new migration — `record_type` (from #6178) already includes `catalog_record`.

Addresses [GSA/data.gov#6181](https://github.com/GSA/data.gov/issues/6181).

**Note on scope overlap:** there's also an independent open PR (#829) implementing #6181 with its own competing `record_type` migration. Flagging so whoever reviews is aware — this PR and #829 will conflict if both land as-is.

## Commits
1. Generalize non-dataset record handling (registry-based) + wire up CatalogRecord extraction/validation
2. New CatalogRecord fixture data
3. Tests for CatalogRecord extraction/validation/persistence

## Test plan
- [x] Full `pytest tests/unit tests/integration` — 678 passed; the 12 failures present are pre-existing and unrelated (same failures as on the base #6178 branch, confirmed unchanged)
- [x] Manually verified no new migration needed; `record_type` enum already includes `catalog_record` from #6178


jbrown-xentity (approved):
This all looks good to me, and improves upon the dataService implementation.
Still needs to be rebased cleanly before merging I think.

PR #840, opened 2026-08-12
Support harvesting DCAT-US3 DatasetSeries objects

Aguiardavidm:
## Summary

Stacked on #835 (#6181, CatalogRecord), which is itself stacked on #831 (#6178, DataService).

- Adds `extract_dcatus3_catalog_dataset_series` and wires `data_series` into the `NON_DATASET_RECORD_TYPES` registry established in #835 — same "extract from a catalog field, validate, persist as HarvestRecord only" shape as CatalogRecord.
- The interesting part: a DatasetSeries's `seriesMember`/`first`/`last` fields embed full inline `Dataset` objects (same shape as `DataService.servesDataset` from #831), not references. These are pulled out via `extract_dcatus3_nested_datasets` (already built generically for exactly this in #831) and merged into the normal dataset pipeline — so series members get real `Dataset` rows, appear in the catalog/OpenSearch, and are tagged with `parent_identifier` = the series' own `@id` (reusing the existing `HarvestRecord.parent_identifier` column, same mechanism as WAF-collection parents).
- `extract_dcatus3_nested_datasets` gained two things to make this work correctly:
  - `parent_identifier_field`, since DatasetSeries (like CatalogRecord) has no `identifier` field, only `@id`.
  - intra-parent dedup by dataset identifier, since a series' `first`/`last` are typically also present in `seriesMember` — that's redundancy in the source data, not a harvest-time duplicate-identifier error, and would otherwise get flagged as one.
- Found and fixed a real bug along the way: `HarvestSource._validators` had no `data_series` entry, and `validator_for()` silently fell back to the Dataset validator for any unrecognized `record_type` — so a DatasetSeries record would have validated (and failed) against the Dataset schema instead of its own DatasetSeries schema. Validators are now built generically off `NON_DATASET_RECORD_TYPES` (which gained a `schema_ref` field) so the two registries can't drift apart again, and `validator_for()` now raises on an unknown `record_type` instead of silently mis-validating.

New migration (`c3e5a7b9d1f3`) adds `data_series` to the `record_type` enum, chained on top of #835's `catalog_record` migration — tested both upgrade and downgrade against the live dev DB.

Addresses [GSA/data.gov#6179](https://github.com/GSA/data.gov/issues/6179).

## Commits
1. `record_type` enum migration for `data_series`
2. DatasetSeries extraction/validation (+ the validator-fallback bug fix)
3. New DatasetSeries fixture (with the seriesMember/first/last redundancy case)
4. Tests for extraction/validation/persistence

## Test plan
- [x] Full `pytest tests/unit tests/integration` — 692 passed; the 12 failures present are pre-existing and unrelated (same failures as on the base #6181 branch)
- [x] Manually tested the new migration's upgrade and downgrade paths against the live dev DB

### DIFF
--- a/example_data/dcatus/dcatus3_0_series_with_members.json
+++ b/example_data/dcatus/dcatus3_0_series_with_members.json
@@ -1,0 +1,80 @@
+{
+  "@type": "Catalog",
+  "@id": "https://example.gov/data-series-with-members.json",
+  "conformsTo": {
+    "@type": "Standard",
+    "title": "DCAT-US 3.0"
+  },
+  "title": "Example DCAT-US 3.0 Catalog With A Dataset Series",
+  "description": "A sample DCAT-US 3.0 catalog where a DatasetSeries's seriesMember/first/last embed full Dataset objects, used for harvest testing.",
+  "datasetSeries": [
+    {
+      "@type": "DatasetSeries",
+      "@id": "https://example.gov/series/annual-report",
+      "title": "Annual Report Series",
+      "description": "A series of annual reports, one dataset per year.",
+      "seriesMember": [
+        {
+          "@type": "Dataset",
+          "title": "Annual Report 2023",
+          "description": "The first dataset in the series.",
+          "identifier": "https://example.gov/datasets/annual-report-2023",
+          "publisher": {
+            "@type": "Organization",
+            "name": "Test Agency"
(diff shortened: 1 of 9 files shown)


jbrown-xentity (approved):
LGTM, ready for release.

PR #844, opened 2026-08-12
5919 - Return early on harvest source delete to avoid 504s

cody-seibert-gsa:
#https://github.com/GSA/data.gov/issues/5919

## Summary
- Schedule harvest-source CASCADE deletes as a CF/local background task instead of running them inside the HTTP request (which often hits gateway/gunicorn timeouts).
- Keep the clear-first check synchronous; on accept, flash/return *“This harvest source may take some time to delete.”* (UI redirect + API 202).
- Add a dedicated `harvester/delete_source.py` task entrypoint and cover enqueue/entrypoint behavior with unit tests; update Playwright expectations to wait for async completion.


harvester/lib/source_delete.py
```diff
@@ -0,0 +1,52 @@
...
+    task_id = f"delete-harvest-source-{source_id}"
+    command = f"python harvester/delete_source.py {source_id}"
+    try:
+        _task_handler_for_delete().start_task(command=command, task_id=task_id)
+    except Exception as e:
+        err = f"Failed to schedule harvest source delete :: {repr(e)}"
+        logger.error(err)
+        return err, 500
+
+    logger.info("Scheduled harvest source delete task %s", task_id)
```
Aguiardavidm, line 55: Once this returns 202, the request path is done — but the background task re-checks `can_delete_harvest_source` and can still fail (409/404) if state changed between the web-request precheck and task execution (e.g. new records added to the source in that window). That failure only gets logged server-side; the user already saw "may take some time to delete" and was redirected to the list, with no way to know the delete never happened. Worth at least a code comment noting this tradeoff, or consider surfacing failed deletes somewhere visible (flash on next list load, a status column, etc.).

app/main/harvest_sources.py
```diff
@@ -297,15 +298,12 @@ def update_harvest_source_actions(source_id: str):
...
             _log_mutation("delete", "harvest_source", source_id, status=status)
             flash(message)
-            if status == 409:
-                return redirect(
-                    url_for("main.view_harvest_source", source_id=source_id)
-                )
-            else:
+            if status == 202:
                 return redirect(url_for("main.harvest_source_list"))
+            return redirect(url_for("main.view_harvest_source", source_id=source_id))
```
Aguiardavidm: This fallback got inverted from the old logic (only 409 → view page, everything else → list) to only 202 → list, everything else → view page. That means a 404 (source not found) or 500 (task scheduling failed) now redirects back to `view_harvest_source(source_id=source_id)` instead of the list — for a 404 in particular, that page is rendering a source that (by definition) does not exist. Worth explicitly branching on 404/500 back to the list, similar to the old default case.

harvester/lib/source_delete.py
cody-seibert-gsa: Agreed. I added an explicit note to `enqueue_harvest_source_delete` documenting that the task repeats the precheck, logs a second-check failure, and leaves the source in the list because deletes do not currently have a durable task-status UI. Adding persisted delete status would be a larger feature, so I kept this PR scoped to documenting the tradeoff. Addressed in fc8ff9e8.

app/main/harvest_sources.py
cody-seibert-gsa: Agreed. I restored the prior redirect rule: only a 409 returns to the source detail page; 202, 404, 500, and other statuses return to the source list. I also added parameterized route coverage for 202/404/409/500. Addressed in fc8ff9e8.

jbrown-xentity (approved):
I think this looks good. One final caveat: the datasets in openSearch won't sync until overnight. I don't think that's a problem, but may confuse users that think that the "clear" is complete but still shows up in catalog.
However right now we're the only users, so it's not really important.

FuhuXia:
> I think this looks good. One final caveat: the datasets in openSearch won't sync until overnight. I don't think that's a problem, but may confuse users that think that the "clear" is complete but still shows up in catalog. However right now we're the only users, so it's not really important.

@jbrown-xentity OpenSearch index happens realtime during dataset deletion. We dont need sync job overnight. The sync job was there mostly for catching any glitches.

PR #846, opened 2026-08-13
6084: Add code_repo_url and code_repo_exempt to the organization model and database schema

### LINKED ISSUE #6084: Add `code_repo_url` and `code_repo_exempt` to the Organization model and database schema
**What:** Add two new columns to the `Organization` model in both `database/models.py` (datagov-harvester) and `app/models.py` (datagov-catalog), and write the Alembic migration.

- `code_repo_url`: nullable String, stores the URL of the agency's source code repository
- `code_repo_exempt`: Boolean, default False, indicates OMB has approved an exemption for this agency

**Why:** The shared Postgres database is the single source of truth for organization metadata. The catalog reads from it but does not write. Adding the fields here makes them available to both applications. The comment at the top of both model files already says "Keep this file in sync between datagov-harvester and datagov-catalog" -- this ticket follows that convention.

**Done When:**
- `code_repo_url = db.Column(db.String)` is added to the `Organization` class in `database/models.py` in datagov-harvester
- `code_repo_exempt = db.Column(db.Boolean, default=False, nullable=False, server_default="false")` is added to the same class
- Both lines are added identically to `app/models.py` in datagov-catalog
- An Alembic migration in datagov-harvester adds both columns to the `organization` table
- Migration runs cleanly against the staging database
- No existing organization records are affected (`code_repo_url` is nullable, `code_repo_exempt` defaults to False)

**Notes:**
- `code_repo_url` is a plain `String` (not `Text`) consistent with `logo` and `slug` on the same model.
- `code_repo_exempt` should have a `server_default` of `"false"` so existing rows get the correct value without a data migration.
- The two fields are independent. An org should not have both `code_repo_url` set and `code_repo_exempt` True. No database constraint is needed for MVP but the admin UI should make this clear (see sub-ticket 2).
- The `.github/workflows/check-db-models.yml` workflow in datagov-catalog exists to catch model drift. Both files must be updated in the same PR or the CI check will fail.
- **Test case:** Use the GSA organization record (`slug: gsa`) to validate the migration. Set `code_repo_url` to `[https://github.com/GSA`](https://github.com/GSA) and confirm it is stored and readable. Confirm `code_repo_exempt` defaults to False on all existing rows.

**Security Considerations:** No sensitive data. Both fields are admin-entered, not user-supplied from the public.

jaredb96:
PR for [Issue #6084](https://github.com/GSA/data.gov/issues/6084). For [Epic #6083](https://github.com/GSA/data.gov/issues/6083). This implements the SHARE IT Act requirement for Data.gov to display federal agency source code repositories. Two new Organization model fields: code_repo_url - link to agency's code repository and code_repo_exempt - flag for OMB exemptions.


jbrown-xentity (approved):
Good and simple. Should work with @rshewitt and review status of https://github.com/GSA/datagov-harvester/pull/847 before making DB migration happen...

PR #847, opened 2026-08-13
closes scoped_session at end of function with idling transaction

rshewitt:
# Pull Request

Related to [6223](https://github.com/GSA/data.gov/issues/6223). 

## About
- relevant [comment](https://github.com/GSA/data.gov/issues/6223#issuecomment-5270027423). basically, `get_new_harvest_jobs_in_past` produces an `idle in transaction` process in postgres lasting quite a bit (almost 30 minutes in prod). i pushed this update to dev yesterday and so far there's no longer issues with that query. here's the sql command you can use to check pg activity. it'll store the results locally in a csv.
- i'd prefer to update load manager to accept a session and manage it's own db interface instead of relying solely on the harvester session and db interface but that would require more change.

```sql 
\copy (
SELECT
    state,
    now() - xact_start as duration,
    now() - query_start as query_duration,
    wait_event_type,
    wait_event,
    query
FROM pg_stat_activity
WHERE datname = current_database()
ORDER BY xact_start NULLS LAST
) TO 'processes.csv' WITH CSV HEADER;
```

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/harvester/lib/load_manager.py
+++ b/harvester/lib/load_manager.py
@@ -79,34 +79,40 @@ class LoadManager:
         task before it stops so we adjust the running_tasks calculation and
         only schedule at most one new job.
         """
-        running_tasks = self.handler.num_running_app_tasks()
-        if running_tasks is None:
-            # None here indicates that tasks couldn't be listed with the API
-            # so be safe by not doing anything.
-            logger.warning("Not starting new jobs because tasks could not be listed")
-            return
+        try:
+            running_tasks = self.handler.num_running_app_tasks()
+            if running_tasks is None:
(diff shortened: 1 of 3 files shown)


FuhuXia:
This change addresses one of several potential causes of hanging connections.

For this particular case, we can make the cleanup unconditional rather than limiting it to the happy path. The DB session should be closed even when `_start_new_jobs()` returns because task listing failed or the maximum number of tasks is already running. I would handle the cleanup with a `try/finally` block.

jbrown-xentity (approved):
LGTM

PR #849, opened 2026-08-13
store DCAT warnings through non-critical error helper

FuhuXia:
# Pull Request

Related to
- https://github.com/GSA/data.gov/issues/6129

## About

DCAT-US 3 warnings are non-fatal harvest issues, the same category as SpatialTransformationException warnings. Store them through `log_non_critical_error` with is_error=False and severity=warning instead of hand-building harvest_record_error rows.

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -1098,15 +1098,14 @@ class Record:
         warning is not a failure and must not clobber an errored (or
         successful) record's status.
         """
-        self.harvest_source.db_interface.add_harvest_record_error(
-            {
-                "message": warning.message,
-                "type": warning.warning_type,
-                "date_created": get_datetime(),
-                "harvest_job_id": self.harvest_source.job_id,
-                "harvest_record_id": self.id,
-                "severity": "warning",
-            }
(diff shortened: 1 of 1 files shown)


harvester/harvest.py
```diff
@@ -1098,15 +1098,14 @@ def _report_warning(self, warning: DcatWarning) -> None:
...
-                "harvest_record_id": self.id,
-                "severity": "warning",
-            }
+        log_non_critical_error(
+            warning.message,
+            self.harvest_source.job_id,
+            self.id,
+            warning.warning_type,
+            is_error=False,
+            emit_log=False,
```
jbrown-xentity, line 1209: I would say there's no reason to suppress the log statement, but there's no harm in doing it either if it's persisted in the DB.

jbrown-xentity (approved):
LGTM. Minor quibble.

harvester/harvest.py
FuhuXia: I want to log it but I want to log it as logger.warning, not logger.error. More change needed to get it implemented. Maybe through another issue another PR.

PR #859, opened 2026-08-17
6245: persist and index DatasetSeries as a Dataset row

Aguiardavidm:
## Summary
- Adds `Dataset.type` (mirrors `HarvestRecord.record_type`); `Record.sync()` now persists a `Dataset` row for `data_series` records instead of HarvestRecord-only, so DatasetSeries are searchable/displayable like a dataset (paired with GSA/data.gov#6245).
- Threads `type` through the OpenSearch mapping/document so catalog can distinguish record kinds.
- Adds a "Type" column to the harvest source's datasets table in the admin UI.
- Fixes a false-positive duplicate-identifier error: a dataset listed both at a catalog's top level and as a `DatasetSeries` member (e.g. `seriesMember`) is the same dataset, not a duplicate. `merge_dcatus3_datasets` keeps the top-level copy and tags it with the series' `parent_identifier` instead of erroring on the second copy.
- Fixes `CatalogRecord` records with no `@id` being dropped with a hard error. `@id` is optional per the DCAT-US 3.0 schema (only `modified`/`primaryTopic` are required); `backfill_catalog_record_identifiers` now synthesizes a stable identifier from those two required fields so the record still harvests.

## Test plan
- [x] `make test-unit` / `make test-integration` pass
- [x] Verified locally against `example_data/dcatus/dcatus3_0_series_with_members.json`: harvesting produces 3 `dataset` rows (2 members + 1 series, `type=data_series`)
- [x] Migration applies cleanly upgrade/downgrade from an empty DB
- [x] New fixture `example_data/dcatus/dcatus3_0_series_member_also_top_level.json` covers the top-level/series-member overlap case end to end (no duplicate error, one row, tagged with the series' identifier)
- [x] Existing `dcatus3_0_record_no_id.json` fixture/test updated to assert the `CatalogRecord` now harvests successfully with a synthesized `@id`, instead of asserting the old drop-with-error behavior


jbrown-xentity (dismissed):
This all looks good to me, but will require a re-index. I think this should NOT be merged until  https://github.com/GSA/data.gov/issues/5885 is complete. Doing this in off hours isn't worth it at this time.

jbrown-xentity (approved):
LGTM

PR #861, opened 2026-08-17
6202 fix unexpected job failure source harvester

marcos-nieto-usds:
# Pull Request

Related to: 
- https://github.com/GSA/data.gov/issues/6202
- the referenced ticket originated from the failed harvest job: https://harvest.data.gov/harvest_job/a04294f5-afb7-484d-9012-7ed688566fb0

## Background:
The main purpose of this story is to improve how harvest jobs are selected and started by the scheduler in the harvester app.

Previously, _start_new_jobs() selected eligible "new" jobs and then attempted to start them, which left room for duplicate selection under concurrent scheduler/task activity. In addition, jobs for a source that already had an active in_progress job could be repeatedly claimed and bounced back, creating a potential for an unwanted loop.

The goal of this change is to make job pickup safer and more predictable by claiming jobs and marking them in progress, before task startup, while also respecting the rule that **only one job per source should run at a time**.

This is important because in the referenced failed harvest job linked above, we found that the job was picked up by both app instances. This caused the error because each task canceled the other task that already picked up the job.

## Changes:
In order to prevent duplicate selection of the same queued harvest job, the main changes revolve around adding a FOR UPDATE SKIP LOCKED to avoid duplicate claim attempts across concurrent schedulers.

There was a change introduced, which was recommended by James. Basically, we want to update the status of a job to in_progress as early as possible during start_job so that there is less of a chance for a race condition since we still check if we already have a job in progress in this method. The update is done in the beginning of start_job which is called after fetching the harvest_job with the new update_harvest_job_if_status method. 

In start_job, if another job in progress is found for the same job source, then the job that was changed to in_progress will go back to new since another job already in progress was found for that source. 

As a result:

concurrent scheduler runs are less likely to pick up the same queued job,
busy-source jobs remain queued instead of being repeatedly reprocessed,
eligible jobs from other sources can still be started,
and the existing manual start behavior remains intact.

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


rshewitt (dismissed):
this seems overly complicated. why can't you just simply add `.with_for_update(skip_locked=True)` after the order_by in `get_new_harvest_jobs_in_past`? the query would look like...
```python
return (
            self.db.query(HarvestJob)
            .filter(
                HarvestJob.date_created < datetime.now(timezone.utc),
                HarvestJob.status == "new",
            )
            .order_by(asc(HarvestJob.date_created))
            .with_for_update(skip_locked=True)
            .limit(limit)
            .all()
        )
```
it would hold onto the lock until the session closes in the finally block

marcos-nieto-usds:
> this seems overly complicated. why can't you just simply add `.with_for_update(skip_locked=True)` after the order_by in `get_new_harvest_jobs_in_past`? the query would look like...
> 
> ```python
> return (
>             self.db.query(HarvestJob)
>             .filter(
>                 HarvestJob.date_created < datetime.now(timezone.utc),
>                 HarvestJob.status == "new",
>             )
>             .order_by(asc(HarvestJob.date_created))
>             .with_for_update(skip_locked=True)
>             .limit(limit)
>             .all()
>         )
> ```
> 
> it would hold onto the lock until the session closes in the finally block

I agree. I originally had something like this but ran into a looping issue, but it could've been related to my data. Let me try an approach like this again.

marcos-nieto-usds:
I got the simpler .with_for_update(skip_locked=True)to work with one caveat.

There's a bit of contention with the line .options(lazyload(HarvestJob.source)).

The reason for that line is to avoid the error:
ERROR [harvest_runner.check_for_more_work:1613] Error checking for more work: NotSupportedError('(psycopg.errors.FeatureNotSupported) FOR UPDATE cannot be applied to the nullable side of an outer join')
This error makes the app exit immediately.

This happens because of the query built behind the scenes via this relationship:
this is from models.py
records = relationship(
        "HarvestRecord",
        backref="job",
        cascade="all, delete-orphan",
        passive_deletes=True,
        lazy=True,
    )
jobs has this backref relationship with its parent, which is source:
jobs = relationship(
        "HarvestJob",
        backref=backref("source", lazy="joined"),
        cascade="all, delete-orphan",
        passive_deletes=True,
        lazy=True,
    )
That's why I got a LEFT OUTER JOIN appended to my query, because it must backref the source (the parent of a job), even if you are not using it at that moment. Because of the backref, a trip to the db for a HarvestJob will get me source data.

I've been testing my changes locally, with test data where I have a source and I add a new job while that source has an existing in_progress job, my new job does not get picked up because there exists an in_progress job for that source. When I move my new job to a separate source, then the job can complete via a full harvest.
INFO [harvest_runner.harvest_job_starter:1598] Harvest job completed for JobId: 11069a6d-2513-4c54-963f-4e957162b20c

cody-seibert-gsa (approved):

PR #866, opened 2026-08-18
add new relic db monitor script

rshewitt:
# Pull Request

Related to [6252](https://github.com/GSA/data.gov/issues/6252)

## About
adds a script for sending db "idle_in_transaction" data to new relic. this occurs on flask app startup after the last db operation. flask-sqlalchemy shapes session management around the request lifecycle so sessions are closed following a response. it also occurs at the end of a harvest. 

another thing we could look into is setting this for our scheduled scripts (e.g. "update_dataset_view_count") but for now let's try to see what's happening during harvesting and the load manager (called on flask app startup) 

[image of event on prod](https://github.com/GSA/data.gov/issues/6252#issuecomment-5331516391)
## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -17,6 +17,7 @@ from app.local_dev_auth import is_running_on_cloud_foundry
 from app.startup_validation import validate_required_env_vars
 from config.logger_config import LOGGING_CONFIG
 from database.models import db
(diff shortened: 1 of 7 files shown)


jbrown-xentity (approved):
LGTM. Will help track this problem in the future.

PR #867, opened 2026-08-18
#6243: count a dataset quality issue once, as either an error or a warning

cody-seibert-gsa:
Related to https://github.com/GSA/data.gov/issues/6243

Harvesting counted one dataset-quality defect twice: a schema `ValidationError` and a content warning. The reported example is `@id: "qwer"`, which produced a `format: iri` error and an `invalid_iri` warning.

**Fix:** JSON Schema owns type, format, length, and enum. `dcat_warnings` only runs content checks on values JSON Schema already accepts.

### Removed IRI checks

JSON Schema already requires an IRI on these fields, so the extra warning was a second ticket for the same defect.

- `_is_valid_iri`, `_warn_iri`, `_warn_iris` — deleted. Every DCAT-US 3 `@id` is `{"type": "string", "format": "iri"}`; `relation`, `image`, `isReferencedBy`, and `conformsTo` are schema-enforced IRIs too.
- `detect_dcat_warnings` — no longer walks every object for `@id`. That walk was the source of the `invalid_iri` warning on `"qwer"`.
- `_dataset_warnings`, `_distribution_warnings`, `_dataservice_warnings` — no longer call `_warn_iris` / `_warn_iri` for `relation`, `image`, `isReferencedBy`, or `conformsTo`.

### Content warnings that still run

These methods still flag schema-valid data that looks wrong. Each one now returns no warning for the listed values, because JSON Schema already reports those values as errors.

- `_warn_duplicate_keywords` — still warns when a keyword appears more than once. Empty `""` entries are skipped: `keyword` items have `minLength: 1`, so an empty string is already a schema error.
- `_warn_spatial_resolution` — still warns when a string is not a number greater than zero. Non-string values are skipped: `spatialResolutionInMeters` is `["null", "string"]`.
- `_warn_temporal_resolution` — still warns when a string is not an xsd:duration. Non-string values are skipped: `temporalResolution` is `["null", "string"]`.
- `_warn_byte_size` — still warns when a string does not parse as a number. Non-string values are skipped: `byteSize` is `["null", "string"]`. A list such as `[]` used to reach `is_number()` and raise `TypeError`.
- `_warn_expected_data_type` — still warns when a string does not start with `xsd:`. Non-string values are skipped: `expectedDataType` is a scalar string.
- `_warn_spatial_unresolved` — still warns when a schema-valid `bbox` or `geometry` cannot be resolved to a place. Skipped when the value is not a string or object, when an object is missing `type`/`coordinates`, or when a `bbox` object’s `type` is not `"Polygon"` (already a schema `const` error).
- `_warn_empty_address` — still warns when every Address field is `None` or `""`. Wrongly typed values such as `0` or `[]` are no longer treated as empty: those are already schema type errors.
- `_warn_language` — still warns when a 2-character code is missing or not ISO 639-1. Codes longer than 2 characters are skipped: the schema already enforces `maxLength: 2`.
- `_warn_character_encoding` — still warns when a list entry is not an IANA character set. A bare string is skipped: `characterEncoding` must be `null` or an array of strings.
- `_parse_dcat_date` (used by the `date_out_of_order` warnings) — still compares dates that JSON Schema accepts. Padded strings, missing UTC offsets, and other schema-invalid dates are no longer parsed, so `_warn_created_after` / `_warn_issued_after_modified` cannot emit a second warning for the same defect.

`detect_dcat_warnings` also skips a non-string `@type` instead of looking that value up in `_TYPE_RULES`. A list or dict `@type` is already a schema type error, and used to raise `TypeError: unhashable type`.

### Also found while removing the overlap

`assemble_validation_errors` stored no message for some schema failures (`created: ["2025"]` and similar). The assembler now records one error at the failing field instead of dropping every validator note.

`tests/unit/test_dcat_warning_error_overlap.py` pins the boundary: a warning must not fire on a schema-invalid value, and a schema error must not also produce a warning.

## PR TASKS

- [x] Code well documented
- [x] Files linted

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,7 +76,7 @@ test-e2e-ci: re-up test-playwright test-functional ## All e2e/expensive tests. R
 
 test-ci: up test-unit test-integration test-scripts ## All simulated tests using only db and required test resources. Run on commit.
 
-re-up: clean up sleep-5 load-test-data ## resets system to clean fixture status
+re-up: clean up wait-for-app load-test-data ## resets system to clean fixture status
 
 re-up-debug: clean up-debug load-test-data ## resets system to clean fixture status for flask debugging
 
@@ -103,6 +103,21 @@ clean: ## Cleans docker images
 	
 sleep-5:
 	sleep 5
+
+wait-for-app: ## Wait until flask has finished db upgrade and is serving
+	@echo "Waiting for the app to finish migrations and start..."
+	@i=0; \
+	while [ $$i -lt 60 ]; do \
+		code=$$(curl -sS -o /dev/null -w '%{http_code}' http://localhost:8080/ 2>/dev/null || true); \
+		case $$code in \
+			2*|3*) echo "App is up (HTTP $$code)."; exit 0 ;; \
+		esac; \
+		i=$$((i + 1)); \
+		sleep 1; \
+	done; \
+	echo "App did not become ready in 60s"; \
+	docker compose logs --tail=80 app; \
+	exit 1
 
 lint-check:  ## Lints wtih ruff, isort, black
 	poetry run ruff check .
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -17,7 +17,6 @@ from app.local_dev_auth import is_running_on_cloud_foundry
 from app.startup_validation import validate_required_env_vars
 from config.logger_config import LOGGING_CONFIG
 from database.models import db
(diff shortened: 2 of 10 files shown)


Aguiardavidm (dismissed):
LGTM 
Nit - The comments are very verbose and I would take a pass to get rid of any implicit context or target just technical readers but I understand that ai likes comments more than I do.

Aguiardavidm (approved):

PR #871, opened 2026-08-19
5991: Test WAF parsing across server index formats

cody-seibert-gsa:
Related to GSA/data.gov#5991

## Overview

WAF sources can be served through Apache, nginx, or Microsoft IIS, each of which generates different directory-index HTML.

### Before

`get_waf_datetimes()` searched the listing for dates independently of the XML anchors. For `<pre>` listings, the parser split the text on `.xml` and assumed the resulting dates remained in the same order as the discovered XML files.

A real nginx listing showed that `.html` and `.txt` entries between XML files broke that assumption and could assign a timestamp to the wrong XML file.

A broader catalog audit found another edge case in the BTS listing. Its timestamps use a weekday and full month name, such as `Friday, July 31, 2026 4:53 PM`. That format was not recognized, so all 143 XML files received the fallback timestamp.

### After

The parser iterates over the XML anchors directly. For each anchor, it parses the date from the containing `<tr>` or the adjacent text surrounding the link. This keeps every XML path associated with its own modification date across the three server layouts.

The supported date formats now also include the weekday/month-name format used by BTS. Small captures from real Apache, nginx, IIS, and BTS directories document and exercise each layout and date format.


jaredb96 (approved):
LGTM

PR #872, opened 2026-08-19
Skip Playwright apt deps in e2e CI

cody-seibert-gsa:
## Summary
- `--with-deps` is not installing Chromium libraries on `ubuntu-latest` (those are already present). It is downloading ~21MB of CJK/unifont packages from `azure.archive.ubuntu.com`.
- On some GitHub-hosted runners that mirror crawls at ~16kB/s, so the "Install Playwright Browsers" step stalls for 20–50 minutes. The Chromium zip itself finishes in seconds.
- Drop `--with-deps` and keep `playwright install --only-shell chromium`.

## Test plan
- [ ] Confirm the e2e workflow's Install Playwright Browsers step finishes in seconds, not tens of minutes
- [ ] Confirm Playwright e2e tests still pass without the extra apt fonts
- [ ] If a run fails with a missing `.so`, restore `--with-deps` or install only the missing lib


jbrown-xentity (approved):
LGTM, less external dependencies always a good optimization.

PR #873, opened 2026-08-19
5375: put the harvest calendar on the source

cody-seibert-gsa:
Related to https://github.com/GSA/data.gov/issues/5375 and https://github.com/GSA/data.gov/issues/5374

## What this is about

Data.gov regularly pulls catalogs from agencies. Each pull is a **harvest job**. Each agency feed is a **harvest source** (daily, weekly, or “only when someone clicks Harvest”).

This PR changes **where we remember “when should this source run next?”** and **stops lying about job dates**.

## Before

To remember that a source should run tomorrow, we created a job row *today* with a timestamp in the future. That one timestamp, `date_created`, was doing three jobs:

1. “Run me at this time”
2. “Am I late? If this time is in the past, start me”
3. “When did work actually start?” — and when the job started, we **overwrote** the timestamp with now, so the original plan was gone

On the metrics page, queued jobs and running jobs both showed that field as **Started**, even though for a queued job it meant “scheduled for.”

If you clicked Harvest, we still created a real run immediately. The future “appointment” job stayed in the table next to it. Changing how often a source runs meant deleting that appointment job and making a new one.

## After

The source itself stores “next run at …”. A job row is only created when it is time to actually run (or when someone clicks Harvest).

So for a daily source you typically see:

- Past jobs (finished or failed) — the history
- At most one job that is waiting to start or currently running
- On the source: **next run tomorrow**

What the dates mean now:

- **Next run** (on the source) — when the scheduler should create the next job
- **Created** (on the job) — when we inserted this job row; we never overwrite it
- **Started** (on the job) — when a worker actually picked it up
- **Finished** (on the job) — when it ended (this field already existed)

Clicking Harvest still means “run once, right now.” It does **not** change when the next automatic run is due. If a job is already waiting or running for that source, Harvest is refused.

Changing the schedule (daily → weekly, etc.) throws away a waiting job and picks a new next-run time. Editing the source’s name or URL does not.

## What we did not change

Jobs still slide later in the day over time (`now + 1 day` after each scheduled run). We did not add a separate “scheduled at” date on the job.

## Migration Risks

I'm not worried about this migration on prod — it only touches ~33k harvest jobs and 793 sources (30 MB total). Runs in ~2 seconds before gunicorn even starts, so no downtime. Only caveat: it's one-way, since it deletes 788 placeholder jobs that downgrade() won't restore.


migrations/versions/d7e4b8c1a2f0_add_date_next_run_and_date_started.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+    op.add_column(
+        "harvest_job",
+        sa.Column("date_started", sa.DateTime(), nullable=True),
+    )
+
+    # Old in-progress/complete/error rows used date_created as the start time.
+    op.execute(
+        """
+        UPDATE harvest_job
```
cody-seibert-gsa: I'm worried some of these command will take a long time with how much data we have.  we may need the migration script to do chunking or just allow downtime so that we can just run this with the catalog and harvester disabled.

cody-seibert-gsa:
deployed to develop and it passed https://github.com/GSA/datagov-harvester/actions/runs/33108622128

docs/ops/README.md
```diff
@@ -52,12 +52,13 @@ specified amount of time in the future (see the `-v` option to the Mac OS X
 $ export API_TOKEN=...
 $ export id=.....  # harvest source id
 $ curl -s -H "X-API-Key: ${API_TOKEN}" \
-  https://datagov-harvest-admin-dev.app.cloud.gov/harvest_job/add \
-  --json "{\"harvest_source_id\": \"$id\", \
-           \"status\": \"new\", \
-           \"date_created\": \"$(date -Iseconds -j -v +45M)\"}"
+  https://datagov-harvest-admin-dev.app.cloud.gov/api/harvest_source/edit/$id \
```
jbrown-xentity: This is now just harvest-dev.data.gov

scripts/README.md
```diff
@@ -17,11 +17,12 @@ $ cf env datagov-harvest | grep HARVEST_API_TOKEN
 $ export API_TOKEN=...
 $ cat harvest_source_ids.txt | while read id; do \
     curl -s -H "X-API-Key: ${API_TOKEN}" \
-    https://harvest-dev.data.gov/api/harvest_job/add \
-    --json "{\"harvest_source_id\": \"$id\", \"status\": \"new\", \"date_created\": \"$(TZ=UTC date -Iseconds -j -v +45M)\"}" ;
+    https://harvest-dev.data.gov/api/harvest_source/edit/$id \
```
jbrown-xentity, line 20: Same, should be harvest-dev.data.gov

jbrown-xentity (dismissed):
LGTM, small quibbles.

docs/ops/README.md
cody-seibert-gsa: Fixed in d8a7911c — swapped in harvest-dev.data.gov.

scripts/README.md
cody-seibert-gsa: This line already reads harvest-dev.data.gov (has since 6b78fcf back in April) — nothing to change here, only the docs/ops/README.md example still had the old cloud.gov host.

jbrown-xentity (approved):
LGTM

PR #875, opened 2026-08-21
Write the organization type as its value, not the enum member

### LINKED ISSUE #6207: Organization API rejects valid organization_type values
`POST /api/organization/add` fails with a 400 for any request that includes a valid `organization_type` value (e.g. `"Non-Profit"`), even though the value is one of the documented valid options.

- **Component:** `datagov-harvester` admin app, `app/api/organizations.py` / `app/api_schemas.py`
- **Environment:** reproduced locally (`flask run` against a local Postgres), branch `6180-catalog-metadata`, but the root cause is unrelated to that branch and appears on `main` as well.

**Error logs:**

```
ERROR:root:Error: (builtins.LookupError) 'OrganizationType.Non-Profit' is not among the defined enum values. Enum name: organization_type_enum. Possible values: Federal Gov.., City Govern.., State Gover.., ..., Non-Profit
[SQL: INSERT INTO organization (name, logo, description, slug, organization_type, aliases, id) VALUES (%(name)s::VARCHAR, %(logo)s::VARCHAR, %(description)s::VARCHAR, %(slug)s::VARCHAR, %(organization_type)s, %(aliases)s::VARCHAR[], %(id)s::VARCHAR)]
[parameters: [{'slug': 'opentopography', 'name': 'OpenTopography', 'organization_type': <OrganizationType.Non-Profit: 'Non-Profit'>, 'logo': None, 'description': None, 'aliases': None}]]
```

## How to reproduce

1. Start the admin app locally against a local Postgres instance.
2. Send:
   ```bash
   curl -s -X POST http://localhost:8080/api/organization/add -H "X-API-Key: <token>" -H "Content-Type: application/json" -d '{"name": "OpenTopography", "slug": "opentopography", "organization_type": "Non-Profit"}'
   ```
3. Observe the response and server log.

## Expected behavior

The organization is created successfully with `organization_type` set to `"Non-Profit"`, matching the value sent in the request (same as it works today when `organization_type` is omitted from the payload).

## Actual behavior

The request returns `{"error":"Failed to add organization."}` (HTTP 400). The server log shows the actual underlying `LookupError` above — the insert is rejected by Postgres because the value being bound is the string `"OrganizationType.Non-Profit"` instead of `"Non-Profit"`.

Root cause: `app/api_schemas.py`'s `OrgCreate.organization_type` field is built via `_to_enum("OrganizationType", ORGANIZATION_TYPE_VALUES)`, which deserializes the incoming JSON string into an actual Python `Enum` **member** (`OrganizationType.Non-Profit`), not a plain string. That enum member then flows straight through `deps.add_organization()` → `Organization(**org_data)` without ever calling `.value` on it. The `Organization.organization_type` database column is a plain string-based SQLAlchemy `Enum(*ORGANIZATION_TYPE_VALUES, name="organization_type_enum")` (defined in `datagov-databases`), so when SQLAlchemy binds the Python `Enum` member as a query parameter, it stringifies it via the enum's default `__str__`, producing `"OrganizationType.Non-Profit"` — not a valid value in the Postgres `organization_type_enum` type.

Any request that omits `organization_type` succeeds today, since that fallback path never encounters this conversion at all. This means organizations can currently only be created via this API by leaving `organization_type` unset, or by inserting directly into the database.

## Sketch

- Convert the deserialized enum member back to its `.value` before handing `org_data` to `Organization(...)` — likely in `add_organization_api()` in `app/api/organizations.py`, or inside `deps.add_organization()` — so a plain string reaches the SQLAlchemy `Enum` column, consistent with how the raw column is defined.
- Check whether `edit_organization_api()` / `update_organization()` have the same issue when updating `organization_type` on an existing org.
- Confirm whether the web UI's organization form goes through the same `OrgCreate` schema/path, or a different one — if the same, it's likely affected too and should be covered by the same fix.
- Add a regression test asserting an organization can be created via the API with each valid `organization_type` value, not just with the field omitted.

arpitjain099:
Fixes GSA/data.gov#6207.

`OrgCreate.organization_type` is an `Enum` field over the enum that `_to_enum` builds (`api_schemas.py:40`), and marshmallow deserializes that to the **member**, not to the string. `add_organization_api` passes the loaded dict straight into `Organization` (`app/api/organizations.py:40` then `:50`), and `Organization.organization_type` is declared with the plain values (`database/models.py:63-69`), so the member stringifies to `OrganizationType.Non-Profit` and the insert is rejected:

```
LookupError: 'OrganizationType.Non-Profit' is not among the defined enum values.
Enum name: organization_type_enum. Possible values: Federal Gov.., City Govern.., State Gover.., ..., Non-Profit
```

That is every valid `organization_type`, not just `Non-Profit`.

I checked the two halves separately rather than assuming. What the field returns today:

```
default   -> <OrganizationType.Non-Profit: 'Non-Profit'> | str(): OrganizationType.Non-Profit
by_value  -> <OrganizationType.Non-Profit: 'Non-Profit'> | str(): OrganizationType.Non-Profit
```

and what the column type does with each:

```
enum member    -> LookupError: 'OrganizationType.Non-Profit' is not among the defined enum values ...
plain string   -> bind ok: 'Non-Profit'
```

Worth noting for anyone reaching for it first: `by_value=True` is **not** the fix. It changes which side of the enum the incoming data is matched against and still hands back the member.

Since `_to_enum` builds these with the name and the value the same, unwrapping to `.value` in a `post_load` returns exactly the string that came in, so nothing else has to change. I kept it on `OrgCreate` rather than making it generic: `OrgCreate().load` is the only `.load` call anywhere in `app/`, and the other enum-bearing schemas (`SourceInfo` and friends) are dump-only today, so widening it would be speculative. The same shape would bite them the moment one of them is loaded, which may be worth a follow-up.

Four tests in a new `tests/unit/test_api_schemas.py`. Two cover the fix, one asserts the loaded value survives the actual column's bind processor rather than just checking the Python type, and one asserts an unknown type is still a `ValidationError` so this cannot pass by loosening validation. The first two fail on `main` with the `AssertionError` and the `LookupError` above.

`ruff check`, `ruff format --check` and `black --check` are clean on both files. I ran the new file with `--noconftest`, since the session fixtures in `tests/conftest.py` want a live Postgres and OpenSearch that I do not have here; CI covers the rest.


arpitjain099:
Heads up on the red E2E check, since it is not this change.

The Playwright suite itself passes, including the organization ones this PR touches:

```
tests/playwright/test_organization.py ...........                        [ 75%]
tests/playwright/test_organization_list.py ....                          [ 81%]
```

The job then moves on to `make test-functional`, which fails at collection time before running anything:

```
ERROR collecting tests/functional/test_cf_tasks.py
E   RuntimeError: CF_API_URL, CF_SERVICE_USER, and CF_SERVICE_AUTH must be set in GitHub Actions
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
make: *** [Makefile:65: test-functional] Error 2
```

Those are repository secrets, and GitHub does not expose them to a `pull_request` run from a fork, so the check cannot pass on an outside contribution regardless of the diff. Every one of the last ten internal PRs passes this check; this is the only fork PR among them and the only failure.

Nothing for me to fix here as far as I can tell, but say the word if you would rather I restructure something. If it is worth it on your side, skipping or xfailing `tests/functional/test_cf_tasks.py` when those variables are absent would let outside PRs get a meaningful signal from the rest of the job.

PR #888, opened 2026-08-24
Drop falsy isPartOf before persisting dataset dcat

### LINKED ISSUE #5862: fix collections where isPartOf is null
[dataset](https://catalog.data.gov/dataset/baby-names-from-social-security-card-applications-national-data) apart of a collection which points to the root of the app.

## How to reproduce
go to that dataset

## Expected behavior
not apart of a collection so that section under "Find Related Datasets" shouldn't be there

## Actual behavior
indication the dataset is apart of a collection

## Sketch
ignore isPartOf when null

cody-seibert-gsa:
DCAT-US 1.1 permits `isPartOf: null`, so records with an explicit null pass validation and get persisted as-is, making downstream consumers (search facets, the catalog UI) treat the dataset as part of a collection it isn't actually part of.

Strips a falsy `isPartOf` in `Record._dataset_payload` before it's stored, so the DB, search index, and catalog never see the misleading null/empty value.

Fixes GSA/data.gov#5862


cody-seibert-gsa:
Closing — investigated further and this doesn't actually fix the reported bug. The isPartOf: null this cleans up is stored on the raw upstream record, and the harvester's change-detection hashes the *raw* untransformed source content, not the persisted payload. If a publisher's catalog.json is unchanged (typical for weeks between harvests on a source's normal cadence), the record's action stays None and this code path never re-runs — so it wouldn't touch already-harvested datasets like the one in the report, and provides no benefit until/unless a record is re-created or force-harvested. The actual fix belongs in datagov-catalog (see GSA/datagov-catalog#337), which guards the read/display side and fixes existing data immediately with no re-harvest dependency.

PR #889, opened 2026-08-24
6255: Index parent_identifier as a top-level OpenSearch field

Aguiardavidm:
## Summary
- Add a top-level `parent_identifier` keyword field to the OpenSearch mapping, sourced from `HarvestRecord.parent_identifier` (already populated for ISO WAF collections, DCAT-US 1.1 explicit parents, and DCAT-US 3.0 DataService/DatasetSeries members).
- Purely additive: `dcat.isPartOf` is untouched, existing queries keep working.

## Dependency
`datagov-catalog` PR #423 switches its `collection` filter to this new field — that PR requires this one to merge and the index to be rebuilt first (`flask search rebuild-index`), otherwise the filter has nothing to match on.

Part of GSA/data.gov#6255

## Test plan
- [x] `make test-unit` / `make test-integration` equivalent (`poetry run pytest tests/unit/search tests/unit/test_opensearch_transform.py tests/integration/search`, 155 passed)
- [x] `make lint-check` (clean on changed files; pre-existing unrelated node_modules isort noise)

### DIFF
--- a/search/documents.py
+++ b/search/documents.py
@@ -108,12 +108,17 @@ class DatasetDocument:
 
         popularity = dataset.popularity if dataset.popularity is not None else None
 
+        parent_identifier = getattr(
+            getattr(dataset, "harvest_record", None), "parent_identifier", None
+        )
+
         document = {
             "_index": self.INDEX_NAME,
             "_id": dataset.id,
             "title": index_fields["title"],
             "slug": dataset.slug,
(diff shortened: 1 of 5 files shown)


cody-seibert-gsa:
I think this would require a re-index label as we added a new mapping.

cody-seibert-gsa (approved):

PR #891, opened 2026-08-25
Queue deploys, drop watermark reconstruction from OpenSearch migration detection

cody-seibert-gsa:
**Before:** the OpenSearch migration label detector reconstructed "the last successful release" by querying GitHub Actions run history, then rescanned every commit since that watermark for a `force re-index recommended` label. It needed to do this because GitHub's default `concurrency` behavior cancels a *pending* run outright the moment a newer push queues behind it — so a labeled PR's dedicated run could vanish before it ever executed if another merge landed first, and the watermark walk was the only way to recover a label from a run that never ran. That reconstruction carried its own hand-rolled workarounds (a force-push walk, a compare-truncation guard) and was exactly the kind of guessed-history logic that has already caused a false-positive rebuild once.

**After:** `deploy.yml` and `deploy-development.yml` add `queue: max` to their `concurrency` blocks, so every push queues behind the one in progress instead of replacing it — nothing gets cancelled, and every merge gets its own dedicated run. With no runs ever dropped, the detector no longer needs history: it just checks whether the commit that triggered *this* run has an associated PR carrying the label. Deletes the watermark walk, the run-history query, the force-push handling, and the compare-truncation guard, along with the tests that covered them.

## Testing

`tests/scripts/test_detect_opensearch_migration_label.py` rewritten for the single-commit check (label found, no label, transient API retry, fails closed after repeated failures). Full `tests/scripts` suite (63 tests) passes; ruff/black/isort clean.

Note: `actionlint` (Homebrew, v1.7.12) doesn't yet recognize the `queue` concurrency key — confirmed as a known upstream gap ([rhysd/actionlint#657](https://github.com/rhysd/actionlint/issues/657)), not an error in this YAML. It isn't wired into any CI workflow in this repo.

Not exercised against a live queue-of-merges scenario — worth confirming on `develop` first that a normal single-commit push still detects/skips migration correctly.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -11,18 +11,15 @@ permissions:
 concurrency:
   group: release-main
   cancel-in-progress: false
+  queue: max
 
 jobs:
   detect-opensearch-migration:
     name: Detect OpenSearch migration
     permissions:
-      actions: read
       contents: read
       pull-requests: read
     uses: ./.github/workflows/detect-opensearch-migration.yml
-    with:
-      workflow_file: deploy.yml
-      branch: main
 
   release-staging:
     name: Release staging
--- a/docs/ops/opensearch-zero-downtime-deployment.md
+++ b/docs/ops/opensearch-zero-downtime-deployment.md
@@ -4,9 +4,10 @@ Pull requests carrying the `force re-index recommended` label use a replacement
 OpenSearch cluster when they are merged. This runs in development for merges to
 `develop`, then in staging and production for merges to `main`.
 
-The label detector scans from the last successful release rather than inspecting
-only the current push. This preserves the migration request when GitHub replaces a
-pending release with a newer merge.
+The label detector checks only the commit that triggered the current release run.
+`deploy.yml` and `deploy-development.yml` use `queue: max`, so GitHub queues every
+push behind the one in progress instead of replacing a pending run with a newer
+merge — each merge still gets its own dedicated run and its own label check.
 
 ## Release sequence
 
--- a/.github/scripts/detect-opensearch-migration-label.sh
+++ b/.github/scripts/detect-opensearch-migration-label.sh
@@ -4,11 +4,8 @@ set -euo pipefail
 
 : "${GITHUB_REPOSITORY:?GITHUB_REPOSITORY must be set}"
 
-head_sha=${1:?Usage: detect-opensearch-migration-label.sh <head_sha> <workflow_file> <branch> [label]}
-workflow_file=${2:?Usage: detect-opensearch-migration-label.sh <head_sha> <workflow_file> <branch> [label]}
-branch=${3:?Usage: detect-opensearch-migration-label.sh <head_sha> <workflow_file> <branch> [label]}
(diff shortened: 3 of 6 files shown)


Aguiardavidm (commented):
Two review notes on the queue-based approach — main concern is that the 'no runs are ever dropped' guarantee doesn't survive this repo's own disable-on-release-failure procedure.

docs/ops/opensearch-zero-downtime-deployment.md
```diff
@@ -4,9 +4,10 @@ Pull requests carrying the `force re-index recommended` label use a replacement
 OpenSearch cluster when they are merged. This runs in development for merges to
 `develop`, then in staging and production for merges to `main`.
 
-The label detector scans from the last successful release rather than inspecting
-only the current push. This preserves the migration request when GitHub replaces a
-pending release with a newer merge.
+The label detector checks only the commit that triggered the current release run.
```
Aguiardavidm, line 7: This "no runs ever dropped" guarantee doesn't hold against this repo's own disable-on-release-failure job (both deploy.yml and deploy-development.yml). When a release fails, `gh workflow disable deploy.yml` runs — and a disabled workflow doesn't create a run at all for push events, so `queue: max` never even gets a run to queue. Any PR carrying `force re-index recommended` that merges to main while deploy.yml is disabled is never inspected, and once an operator re-enables it, only the next single commit is checked — the backlog is silently lost. This is exactly the scenario the deleted watermark-walk logic protected against. Worth either keeping a bounded fallback scan for the disable→re-enable case, or calling this out here and adding a manual "check for pending migration labels" step to the re-enable procedure below.

.github/workflows/deploy.yml
```diff
@@ -11,18 +11,15 @@ permissions:
 concurrency:
   group: release-main
   cancel-in-progress: false
+  queue: max
```
Aguiardavidm, line 14: Per GitHub's docs, `queue: max` caps pending runs at 100 — it's not unlimited. If more than 100 pushes land on main while a run is in progress (e.g. a stuck release + a merge burst), runs beyond the cap still get cancelled, reintroducing the drop scenario this PR removes protection for. Unlikely at this repo's cadence, but the docs/PR description should say "queues every push (up to 100 pending)" rather than implying an unconditional guarantee.

docs/ops/opensearch-zero-downtime-deployment.md
codyseibert: I think for now we should just go with:
- notify the team if you're doing a reindex
- no one merge more pull requests with that label when a reindex is in progress
- if someone does and the workflow was disabled, just make a temp pr and add the label and merge to rekick it off (easiest fix)

.github/workflows/deploy.yml
codyseibert: yeah good to point out, but very unlikely

Aguiardavidm (approved):

cody-seibert-gsa:
https://github.com/GSA/datagov-harvester/actions/runs/32885810868 successful deploy to prod

PR #895, opened 2026-08-26
5455: add alembic migration tests

Aguiardavidm:
## Summary
- Add a test that walks every Alembic revision (base -> head) and verifies upgrade -> downgrade -> upgrade succeeds for each one, using a real Postgres instead of the ORM's `create_all()`.
- Fix a bug the new test caught: the initial migration's `downgrade()` dropped tables but never dropped the Postgres ENUM types those columns depended on, so downgrading and re-upgrading failed with "type already exists".

## Test plan
- [x] `make test-integration` passes locally, including the new migration test alongside the rest of the suite (no fixture leakage into other tests)
- [x] Confirmed the fix is needed: reverted it locally and watched the new test fail with `DuplicateObject: type "frequency" already exists`
- [x] Runs automatically on every push via the existing `test-ci` -> `test-integration` step, no new CI wiring required

Follow-up (not in scope here): a reviewer on GSA/data.gov#5455 suggested moving `flask db upgrade` out of `app-start.sh` (which has a startup timeout) into its own release step.


Aguiardavidm:
Retargeting: branch was cut from develop which is 25 commits ahead of main. Recreating the branch off main instead.

PR #897, opened 2026-08-27
6255: Fix WAF collection root incorrectly tagged as its own parent

Aguiardavidm:
## Summary
- The WAF collection root record was unconditionally tagged with `parent_identifier = collection_parent_url`, pointing at itself instead of having no parent.
- This broke the catalog's "am I a collection root with children" detection (`dataset_detail_by_slug_or_id` in datagov-catalog), since it relies on the root record having a null `parent_identifier`.

Part of GSA/data.gov#6255

## Test plan
- [x] Updated `test_harvest_waf_collection` to assert the root record has `parent_identifier is None` and only its children get tagged with `collection_parent_url`
- [x] `poetry run pytest tests/integration/harvest_job_flows/test_harvest_job_full_flow.py -k test_harvest_waf_collection`

### DIFF
--- a/harvester/harvest.py
+++ b/harvester/harvest.py
@@ -584,7 +584,10 @@ class HarvestSource:
                 elif self.source_type == "waf-collection":
                     record["content"] = download_file(record["identifier"], ".xml")
                     dataset = record["content"]
-                    parent_identifier = self.collection_parent_url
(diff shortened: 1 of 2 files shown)


Aguiardavidm:
Combined into #899 for easier review.

PR #898, opened 2026-08-27
6255: Read DCAT-US 1.1 isPartOf field into parent_identifier

Aguiardavidm:
## Summary
- DCAT-US 1.1 datasets declare their parent collection via a native `isPartOf` field, but the harvester only ever read `parent_identifier` off a record dict populated by the DCAT-US 3.0 series/service nested-extraction path (`extract_dcatus3_nested_datasets`). Plain "document" sources on DCAT-US 1.1 never had `isPartOf` read into `parent_identifier`, so `harvest_record.parent_identifier` stayed null even though the raw `isPartOf` value passed through into `dataset.dcat`.
- Fix: fall back to `record.get("isPartOf")` when no `parent_identifier` key is already present (dcatus3.0 nested records still take priority since they set it explicitly).
- Confirmed against catalog-dev: `isPartOf` showed correctly in the child dataset's raw metadata, but neither the child→parent link nor the parent→children list rendered, because `harvest_record.parent_identifier` was never populated.

Part of GSA/data.gov#6255

## Test plan
- [x] New fixture `example_data/dcatus/dcatus1_1_ispartof.json` (a DCAT-US 1.1 collection + one member with `isPartOf`)
- [x] New integration test `test_harvest_dcatus1_1_ispartof_persists_parent` asserting the child's `parent_identifier` resolves to the collection's identifier, and the collection's own `parent_identifier` stays null
- [x] `poetry run pytest tests/integration/harvest_job_flows/test_harvest_job_full_flow.py -k "ispartof or series_members or waf_collection"`

### DIFF
--- a/example_data/dcatus/dcatus1_1_ispartof.json
+++ b/example_data/dcatus/dcatus1_1_ispartof.json
@@ -1,0 +1,54 @@
+{
+  "@context": "https://project-open-data.cio.gov/v1.1/schema/catalog.jsonld",
+  "@id": "https://example.gov/data.json",
+  "@type": "dcat:Catalog",
+  "conformsTo": "https://project-open-data.cio.gov/v1.1/schema",
+  "describedBy": "https://project-open-data.cio.gov/v1.1/schema/catalog.json",
+  "dataset": [
+    {
+      "accessLevel": "public",
+      "bureauCode": ["000:00"],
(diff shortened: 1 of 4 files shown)


Aguiardavidm:
Combined into #899 for easier review.

PR #899, opened 2026-08-27
6255: Fix parent_identifier gaps (WAF collection root, DCAT-US 1.1 isPartOf)

Aguiardavidm:
## Summary
Combines two fixes to `parent_identifier` extraction found while investigating why catalog's parent/child collection links (GSA/data.gov#6255) weren't showing up for some record types on dev:

- **WAF collection root self-referencing its own parent.** The collection root record was unconditionally tagged with `parent_identifier = collection_parent_url`, pointing at itself instead of having no parent. This broke catalog's "am I a collection root with children" detection, since it relies on the root record having a null `parent_identifier`.
- **DCAT-US 1.1 `isPartOf` never read.** DCAT-US 1.1 datasets declare their parent collection via a native `isPartOf` field, but the harvester only ever read `parent_identifier` off a record dict populated by the DCAT-US 3.0 series/service nested-extraction path. Plain "document" sources on DCAT-US 1.1 never had `isPartOf` read into `parent_identifier`, so it stayed null even though the raw value passed through into `dataset.dcat`.

Part of GSA/data.gov#6255

## Test plan
- [x] `test_harvest_waf_collection` updated to assert the root record has `parent_identifier is None` and only its children get tagged with `collection_parent_url`
- [x] New `test_harvest_dcatus1_1_ispartof_persists_parent` with a dedicated DCAT-US 1.1 fixture
- [x] `poetry run pytest tests/integration/harvest_job_flows/test_harvest_job_full_flow.py -k "ispartof or series_members or waf_collection"`
- [x] `make lint-check`

### DIFF
--- a/example_data/dcatus/dcatus1_1_ispartof.json
+++ b/example_data/dcatus/dcatus1_1_ispartof.json
@@ -1,0 +1,54 @@
+{
+  "@context": "https://project-open-data.cio.gov/v1.1/schema/catalog.jsonld",
+  "@id": "https://example.gov/data.json",
+  "@type": "dcat:Catalog",
+  "conformsTo": "https://project-open-data.cio.gov/v1.1/schema",
+  "describedBy": "https://project-open-data.cio.gov/v1.1/schema/catalog.json",
+  "dataset": [
+    {
+      "accessLevel": "public",
+      "bureauCode": ["000:00"],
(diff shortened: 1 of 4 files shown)


jbrown-xentity (approved):
LGTM

PR #903, opened 2026-08-28
6264 Support WKT geometry transformation

cody-seibert-gsa:
## Summary

Closes GSA/data.gov#6264.

DCAT-US3's `Location` fields (`spatial`/`bbox`/`geometry`) accept Well-Known Text as an alternative to GeoJSON (e.g. `POLYGON((-125 24, -66 24, -66 50, -125 50, -125 24))`), but nothing in the harvester understood it. A WKT string fell through every existing translation strategy in `translate_spatial` (GeoJSON parse, DB place-name lookup, loose numeric munging) and came back empty, which raised a `SpatialTransformationException` warning during harvest and left the record without spatial search data in OpenSearch. This is a known, previously "out of scope" gap — the upstream `GSA/dcat-us` complete example itself contains an unconverted WKT polygon, called out in `test_dcat_warning_error_overlap.py` as a pre-existing false-positive tied to GSA/data.gov#6243.

- `translate_wkt_to_geojson` (new, in `harvester/utils/general_utils.py`) converts a WKT string to GeoJSON via `shapely`, guarded by a cheap regex so only strings that actually look like WKT pay the parse cost.
- `translate_spatial` now tries this conversion first, then hands the result through the *existing* `validate_geojson`/`geojson_validator` fix-up pipeline unchanged (antimeridian wrap-around and friends still apply).
- `shapely` is declared as a direct dependency; it was already resolved in `poetry.lock` as a hard transitive dependency of `geojson-validator`, so this adds no new install footprint.
- Every consumer of `translate_spatial` — `harvest.py`'s `_dataset_payload`, `dcat_warnings.py`'s `unresolvable_spatial_value` warning, and `scripts/sync_datasets.py` — is fixed by this one change.

Scope note: kept minimal per the issue's own sketch — this only adds WKT support. It does not attempt the larger "consolidate geojson_validator's validate/fix/antimeridian logic onto shapely and drop geojson-validator" refactor the issue also floats as an option.

## Test plan

- [x] `poetry run pytest tests/unit` — 506 passed
- [x] `poetry run pytest tests/integration` — 378 passed (one pre-existing, unrelated failure in `test_check_more_work`, confirmed present on `develop` via `git stash` before these changes)
- [x] New unit tests: `translate_wkt_to_geojson` (polygon, point, non-WKT string, malformed WKT-looking string) and `translate_spatial` WKT branch
- [x] New `TestLocationSpatial` case confirming the DCAT-US3 `unresolvable_spatial_value` warning no longer fires on WKT `geometry`/`bbox`
- [x] New end-to-end `test_can_translate_wkt_spatial` (mirrors `test_cant_translate_spatial`): harvests a WKT `spatial` value with zero warnings and a correctly populated `Dataset.translated_spatial`
- [x] Manual check: lowercase WKT (`polygon(...)`) also resolves correctly


jbrown-xentity (approved):
LGTM.

PR #911, opened 2026-09-01
6067 bug

akuny:
Part of GSA/data.gov#6067. Uploading a large catalog returned a bare gateway error that named neither the problem nor the limit. The limit was enforced inconsistently and was unreadable everywhere.
 
| Layer | Was | Now |
| --- | --- | --- |
| Flask `MAX_FORM_MEMORY_SIZE` | 500KB default, so a 2MB paste failed where the same 2MB upload passed | tracks `MAX_CONTENT_LENGTH` |
| nginx `client_max_body_size` | 10M, so multipart overhead on a 10MB file drew a bare 413 before Flask saw it | 12M, leaving the app room to answer |
| Browser | no advertised limit, no client check | `usa-hint`, blocks oversized input before submit |
| Flask 413 | APIFlask's bare JSON blob to browsers | `413.html` for HTML routes, `{"error": ...}` for `/api/`, both naming the limit |
 
Every Python path now reads `MAX_UPLOAD_BYTES` from `app/constants.py`, including the URL fetch path in `app/util.py`. `proxy/nginx.conf:42` cannot, and nothing in CI catches that drift. Code comments explain the 12M value and why nginx gets no `error_page 413`. Also here: `proxy_read_timeout 110s`, hardening rather than a fix.
 
## Also in this PR
 
**Deep nesting no longer 500s.** Both schemas have self-referential `$ref`s, which jsonschema resolves by recursion, so a small document can exhaust the stack. A 3.0 catalog fails at depth 200 (~17KB); a 1.1 `subOrganizationOf` chain at 300 (~20KB). `validate_records` now raises `CatalogTooDeeplyNested`, the form renders it as a field error, and the API returns a documented 422.
 
**Error assembly got slower with every error it assembled.** Each error added work to every subsequent error, so the cost of a validation rose with the square of the error count. Two independent causes. `finalize_validation_messages` ran on every recursive call and its return value was thrown away. `_count_messages` re-summed the whole dict twice per error. 1.1 escaped this because the assembler runs once per record; 3.0 validates the whole catalog in one call, so the dict grew with the dataset count. 4,000 datasets: 27s before, 0.02s after (dev laptop; treat the ratio as the result). Recursion moved to a private `_collect_validation_messages` returning its own count. `_forced` left the public signature; every call site passes one argument and the only other reference is a `@patch` in `test_harvest_job_full_flow.py`. The `messages` parameter is now dead and could reasonably be dropped.
 
**Both rode along and can be split out on request.** The assembler change touches `harvester/utils/general_utils.py`, which the harvest pipeline calls at `harvester/harvest.py:1332`.
  
## Tests
 
`make test-ci` 943 passed. New integration tests for the 500KB regression and both 413 shapes, new playwright tests for the client guards, two nesting tests using documents that genuinely exhaust the stack, and two assembler tests.


jaredb96 (approved):
LGTM

PR #912, opened 2026-09-01
6040: Handle DCAT-US 3.0 Location object spatial input

### LINKED ISSUE #6040: Update spatial handling for DCAT-US v3.0 Location object
**What:** DCAT-US v3.0 `spatial` is a Location object, not a plain string. The current `translate_spatial` and `translate_spatial_to_geojson` utilities produce GeoJSON strings or dicts. These need to handle the v3.0 Location object as both input (when reading a v3.0 record) and output (when producing a v3.0-compatible record from ISO or v1.1 source).

**Likely scope:**
- Update `translate_spatial` to accept a Location object dict as input and extract the geometry value from it for processing
- Update or wrap `translate_spatial_to_geojson` to produce a Location object dict as output when the target schema is v3.0
- Ensure the `translated_spatial` column in the dataset table stores the GeoJSON geometry (extracted from the Location object) for OpenSearch indexing
- Update `_dataset_payload` in `Record` to produce the correct spatial structure based on the source `schema_type`

**Done When:**
- v3.0 records with Location object `spatial` pass validation and store correctly
- v1.1 records with plain string `spatial` continue to work
- ISO records (which arrive via MDTranslator as v1.1 output) continue to work
- OpenSearch spatial indexing is unaffected

**Notes:**
- The DCAT-US v3.0 Location class definition is at [github.com/GSA/dcat-us/blob/main/jsonschema/definitions/Location.json](https://github.com/GSA/dcat-us/blob/main/jsonschema/definitions/Location.json)
- Do not change the internal GeoJSON representation used for OpenSearch -- only the DCAT-facing structure needs to change

Aguiardavidm:
https://github.com/GSA/data.gov/issues/6040

## Summary
- `translate_spatial` treated any dict as a bare GeoJSON geometry, so DCAT-US 3.0 `spatial` (a `Location` object or array of them) silently produced no `translated_spatial` and raised a spurious warning on every v3.0 harvest.
- Added `_unwrap_location` to extract the geometry-bearing value (`geometry` → `bbox` → `centroid`, first element if an array) before the existing WKT/GeoJSON logic runs. v1.1 strings and bare GeoJSON dicts are unaffected.
- Deferred scope: producing a `Location` object as *output* (v1.1/ISO → v3.0) — no code path converts a record to v3.0 today, so there's no consumer for it yet.

Closes #6040

## Test plan
- [x] `make test-unit` — 9 new cases (Location object, array, bbox/centroid fallback, no-geometry, non-mutation, real v3.0 fixture) plus full existing suite (519 passed)
- [ ] `make re-up` against a v3.0 test harvest source — confirm `translated_spatial` is non-null and OpenSearch `spatial_shape`/`spatial_centroid` populate for a Location-bearing dataset


jaredb96 (approved):

tests/unit/test_utils.py
```diff
@@ -307,6 +307,89 @@ def test_translate_spatial_wkt_point(self):
...
+                        [-76.909393, 38.791645],
+                        [-76.909393, 38.99538],
+                        [-77.119759, 38.99538],
+                        [-77.119759, 38.791645],
+                    ]
+                ],
+            },
+        }
+        geojson = translate_spatial_to_geojson(location)
+        assert geojson["type"] == "Polygon"
```
jbrown-xentity, line 357: Curious why this only checks the type and not the coordinates...

jbrown-xentity (approved):
LGTM, I'm surprised this wasn't already in place but looks good to me. Small quibble.

PR #913, opened 2026-09-02
Bump Python to a version Supported by the current cloud.gov buildpack

akuny:
cloud.gov recently updated its Python buildpack from 1.9.2 to 1.9.3, and that release dropped 3.12.13 in favor of 3.12.14. Harvester pins an exact Python version, 3.12.13, in three separate files. 

The buildpack now offers 3.10.21, 3.11.16, 3.12.14, 3.13.15, and 3.14.7, none of which match what we ask for, so it fails during the compile phase before installing anything.

This PR bumps python to 3.12.14.

development deployment: https://github.com/GSA/datagov-harvester/actions/runs/33669130016


.python-version
```diff
@@ -1 +1 @@
-3.12.13
+3.12
```
marcos-nieto-usds, line 1: Does this one need to be specific like the others, 3.12.14?

pyproject.toml
```diff
@@ -13,7 +13,7 @@ repository = "https://github.com/GSA/datagov-harvester"
 poetry-plugin-export = ">=1.8"
 
 [tool.poetry.dependencies]
-python = "3.12.13" # stay with 3.12 until *.apps.internal:61443 works for 3.13
+python = "~3.12" # stay with 3.12 until *.apps.internal:61443 works for 3.13
```
marcos-nieto-usds, line 16: What about this one?

runtime.txt
```diff
@@ -1 +1 @@
-python-3.12.13
+python-3.12.x
```
marcos-nieto-usds, line 1: Lastly, this one you opted for "x." Any specific reason?

jbrown-xentity (approved):
LGTM

runtime.txt
akuny: runtime.txt is the only one the buildpack reads, and cloud.gov deletes patch versions for CVEs (1.9.3 dropped both 3.12.13 and 3.12.10), so an exact pin there just breaks again next rotation.

The other two you flagged are local tooling only and do not affect the deploy, so pinning an exact patch in them would only force everyone to build that specific release locally.

.python-version
akuny: See comment for runtime.txt

pyproject.toml
akuny: See comment for runtime.txt

runtime.txt
marcos-nieto-usds: This makes sense, thank you for the explanation.

PR #914, opened 2026-09-02
6284: log WAF detection header on proxy access log

Aguiardavidm:
## Summary
- Adds `$http_x_amzn_waf_detection` to the `cloudfoundry` nginx log_format so flagged requests show *why* they got a 429, instead of just an unexplained status code.
- No behavior change to blocking: the existing detection-based 429 branch in `nginx-malicious.conf` is unchanged (confirmed this is the actual rate-limit enforcement mechanism, paired with the WAF's `count`-only action on the rate-limit rule).

Part of #6284.

## Test plan
- [ ] `make up` and confirm nginx config parses/starts without error
- [ ] Manually set the WAF detection header on a request against local proxy and confirm access log line includes the value


Aguiardavidm:
Closing in favor of #924, which is rebuilt cleanly off main with only the nginx.conf log_format change (this branch had picked up unrelated develop history).

PR #916, opened 2026-09-03
6271: add standalone harvest job report page

### LINKED ISSUE #6271: Develop DCAT US 3 Dev-Test-Report
Develop a "test" report so we can add agencies to the Dev environment, run their harvest source, and send them back a comprehensive email that shows if any errors or warnings were found, and what their files look like generally in the catalog (I'm not sure what they will look like so I dont have this nailed down yet)

As an agency harvest source manager, I would like a harvest source report from the dev system before it goes live to the public so I can see what the public will see, and decide if I need to make any changes before pushing it live to the public.

Develop a standardized report "product" so that we can plan out what this email will have for agencies and send it the same way every time.

Aguiardavidm:
Closes https://github.com/GSA/data.gov/issues/6271

## Summary
- Report link in notification emails is now opt-in per source (new `send_report_email` field on the harvest source form, dropdown, defaults off) so existing sources' notification emails don't change until explicitly enabled
- When opted in, the notification email inlines the report data directly (issues grouped by field, sample datasets) instead of linking out to a separate page, with per-dataset links to catalog.data.gov (via `CATALOG_BASE_URL`)
- Adds a "Catalog" column to the harvest source page's dataset table, linking each dataset to catalog.data.gov
- Drops a line from the notification email that incorrectly described a "test site" / shared login, which doesn't hold once this reaches deployed environments

The standalone `/harvest_job/<id>/report` page from earlier revisions of this PR has been split out to a separate branch (`6271-harvest-report-page`) so this PR can focus on the email-sending pattern and catalog links on existing pages.

## Test plan
- [x] `make test-unit` passes
- [x] `make test-integration` passes (pre-existing OpenSearch/migration failures unrelated to this change, confirmed via `git stash`)
- [x] Migration applies and reverses cleanly against a real db (`flask db upgrade`/`downgrade`)
- [x] Manual check: `send_report_email` dropdown defaults to False and persists on save; notification email includes the inline report section (with catalog links) only when opted in; harvest source page's dataset table shows catalog links
- [x] Manually rendered the notification email body against real fixture data to confirm formatting

Found but did not fix (pre-existing, unrelated): `tests/integration/database/test_migrations.py` doesn't clean up Alembic's `alembic_version` table between runs, so it can leave stale state that breaks itself and other tests run afterward in the same session. Reproduces identically on unmodified `main`.


jbrown-xentity (approved):
LGTM. I would like to see a screenshot of what this will look like in practice in email before we go live, but not critical @Aguiardavidm since it won't be "in use" until the switch is flipped.

PR #920, opened 2026-09-07
6292: add script to validate large catalogs bypassing the 10MB upload cap

### LINKED ISSUE #6292: Add a script to validate large DCAT catalogs bypassing the 10MB upload cap
What: Add a script (scripts/validate_catalog/validate_catalog.py in datagov-harvester) that validates a DCAT-US catalog URL against the schema, bypassing the 10MB upload cap on the hosted /validate page.

Why: The hosted /validate page rejects catalogs over 10MB (app/constants.py MAX_UPLOAD_MB), so large real-world catalogs can't be checked there. Example: King County's DCAT-US 3.0 feed (531 datasets) exceeds the cap.

Done When:

- [x] Script calls validate_records() directly, with no size limit
- [x] Supports --schema to pick dcatus3.0/dcatus1.1 federal/non-federal
- [x] Supports --output/-o to write the full error list to a file instead of stdout, defaulting to a gitignored path and overwriting on rerun

Aguiardavidm:
Closes https://github.com/GSA/data.gov/issues/6292

## Summary
- Adds `scripts/validate_catalog/validate_catalog.py`, which calls the same validation logic used by the hosted `/validate` page (`validate_records` in `app/util.py`) directly, with no size limit
- `--schema` selects the schema (defaults to `dcatus3.0 catalog`)
- `--output`/`-o` writes the full error list to a file instead of the terminal; given with no path, defaults to `scripts/validate_catalog/output/errors.txt` (gitignored), and overwrites on each run

## Test plan
- [x] Ran against a real large catalog (King County's DCAT-US 3.0 feed, 531 datasets) inside the app container, confirmed counts match manual validation
- [x] Confirmed `--output` writes to a host-visible path (repo is the only mounted volume, `/tmp` inside the container isn't), overwrites rather than appends, and never shows up in `git status`
- [x] `ruff check`, `black --check`, `isort --check` all pass


jbrown-xentity (approved):
LGTM. A way for users to access the tooling without taking down our own resources...

PR #922, opened 2026-09-08
6295: Dispatch DCAT-US 3 content warnings on CURIE-prefixed @type

### LINKED ISSUE #6295: DCAT-US 3 warning rules never fire on prefixed @type CURIEs (dct:Location, dcat:Dataset)

## Problem

DCAT-US 3 content warnings never fire when a source uses prefixed `@type` CURIEs, which real-world feeds do.

`_TYPE_RULES` ([dcat_warnings.py:644](https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/dcat_warnings.py#L644)) keys on bare class names (`"Dataset"`, `"Location"`, `"PeriodOfTime"`, …), and `detect_dcat_warnings` ([:671](https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/dcat_warnings.py#L671)) does an exact dict lookup on `obj["@type"]`. Every `@type` in the live OpenTopography feed is prefixed:

```
802  dcat:Dataset      802  org:Organization    802  vcard:Contact
802  dct:Location      802  dcat:Distribution     1  dcat:Catalog
```

Zero matches, so **the entire warning system is a no-op for that source**. Confirmed both directions:

```
prefixed dct:Location,     unresolvable bbox    -> no warnings
bare     Location,         unresolvable bbox    -> ['unresolvable_spatial_value']
prefixed dct:PeriodOfTime, endDate < startDate  -> no warnings
bare     PeriodOfTime,     endDate < startDate  -> ['date_out_of_order']
```

This is also why the 83 dropped-spatial records in #6294 emit no `unresolvable_spatial_value` warning.

## Why it wasn't caught

The v3.0 schema deliberately permits CURIEs — `@type` is `{"type": "string", "default": "Location"}` with no `enum`/`const`. But every fixture in `example_data/dcatus/dcatus3_0*.json`, the schema's own `complete_example.json`, and every assertion in `tests/unit/test_dcat_warnings.py` uses bare names, so nothing in the repo exercises the prefixed form.

## Suggested fix

Normalize `@type` before dispatch in `detect_dcat_warnings` — match on the local name after the last `:`, or strip a known prefix set (`dcat:`, `dct:`, `dcterms:`, `vcard:`, `org:`, `skos:`, `foaf:`, `adms:`, `locn:`). Matching on the local name is simpler and prefix-agnostic.

## Acceptance criteria

- [ ] `{"@type": "dct:Location", …}` dispatches to the same rules as `{"@type": "Location", …}`
- [ ] Same for `dcat:Dataset`, `dcat:Distribution`, `dct:PeriodOfTime`, `vcard:Contact`, `org:Organization`
- [ ] Bare `@type` behavior is unchanged (existing tests keep passing)
- [ ] A non-string or unrecognized `@type` is still skipped without raising
- [ ] Unit tests cover the prefixed form — there are currently none
- [ ] At least one `example_data` fixture uses prefixed `@type` so this can't regress

## Context

Found in #6038 (SPIKE-3). Blocks #6197. Related to the #5970 warnings epic.

cody-seibert-gsa:
## Summary

Real DCAT-US 3 feeds (e.g. `portal.opentopography.org`'s `catalog30.json`) send CURIE-prefixed `@type` values like `dcat:Dataset` rather than bare class names. `_TYPE_RULES` in `harvester/utils/dcat_warnings.py` keyed on bare names only, so `detect_dcat_warnings()` never matched and silently produced zero content-quality warnings for such sources — even though the DCAT-US 3.0 schema permits CURIEs (`@type` has no enum/const).

## What changed

- `detect_dcat_warnings` now matches on the `@type`'s local name (the part after the last `:`) before dispatching to `_TYPE_RULES`, so a prefixed value (`dct:Location`) dispatches identically to its bare form (`Location`). Bare `@type` and non-string/unrecognized `@type` behavior is unchanged.
- Added unit tests in `tests/unit/test_dcat_warnings.py` covering the prefixed form.
- Added `example_data/dcatus/dcatus3_0_warning_curie.json`, a CURIE-typed counterpart to the existing `dcatus3_0_warning.json` fixture, to guard against regressions.

Closes GSA/data.gov#6295

### DIFF
--- a/example_data/dcatus/dcatus3_0_warning_curie.json
+++ b/example_data/dcatus/dcatus3_0_warning_curie.json
@@ -1,0 +1,85 @@
+{
+  "@type": "dcat:Catalog",
+  "@id": "https://example.gov/data.json",
+  "conformsTo": {
+    "@type": "dct:Standard",
+    "title": "DCAT-US 3.0"
+  },
+  "title": "Example DCAT-US 3.0 Catalog (CURIE-prefixed @type warnings)",
+  "description": "Same shape as dcatus3_0_warning.json, but every @type is a CURIE (e.g. dcat:Dataset) rather than a bare class name, matching real-world feeds such as https://portal.opentopography.org/csw/catalog30.json (GSA/data.gov#6295).",
+  "dataset": [
+    {
+      "@type": "dcat:Dataset",
+      "title": "Curie Warned Dataset One",
+      "description": "Schema-valid, but issued is later than modified.",
+      "identifier": "https://example.gov/datasets/curie-warning-one",
+      "issued": "2030-01-01",
+      "modified": "2020-01-01",
+      "publisher": {
+        "@type": "org:Organization",
+        "name": "Test Agency"
+      },
+      "contactPoint": {
+        "@type": "vcard:Contact",
+        "fn": "Test Contact One",
+        "hasEmail": "mailto:one@example.gov"
+      }
+    },
+    {
+      "@type": "dcat:Dataset",
+      "title": "Curie Warned Dataset Two",
+      "description": "Schema-valid, but the distribution declares an unrecognized media type.",
+      "identifier": "https://example.gov/datasets/curie-warning-two",
+      "publisher": {
+        "@type": "org:Organization",
+        "name": "Test Agency"
+      },
+      "contactPoint": {
+        "@type": "vcard:Contact",
+        "fn": "Test Contact Two",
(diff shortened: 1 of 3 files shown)


jbrown-xentity (approved):
LGTM. I think the JSON-LD formats and types need to be more formally defined by OMB, hoping that happens in DCAT-US3.1...

PR #925, opened 2026-09-08
6296: pick the first usable Location in a spatial array, not first truthy

### LINKED ISSUE #6296: `spatial` arrays pick the first *truthy* Location, not the first *usable* one

## Problem

When a v3.0 `spatial` value is an array of Location objects, `_unwrap_location()` ([general_utils.py:1250-1251](https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/general_utils.py#L1250-L1251)) selects the first element that is *truthy*, not the first that is actually *usable* (i.e. has `geometry`/`bbox`/`centroid`):

```python
if isinstance(input_value, list):
    input_value = next((item for item in input_value if item), None)
```

A non-empty dict with only `prefLabel` is truthy, so it wins over a later element that actually carries geometry:

| Input | Result |
|---|---|
| `[{bbox: POLYGON(…)}, {prefLabel: "Alaska"}]` | ✅ Polygon |
| `[{prefLabel: "Alaska"}, {bbox: POLYGON(…)}]` | ❌ `NULL` |

Order in the source feed silently determines whether a dataset gets a `translated_spatial` at all.

Separately (lower priority, same function): even when selection works, only one of N array elements is ever kept, so a multi-region `spatial` silently drops every region but the first. `spatial_shape` is a `geo_shape` OpenSearch field, so combining into a `MultiPolygon`/`GeometryCollection` is possible if that's wanted — that part is a product call, not filed as a fix here.

## Suggested fix

In `_unwrap_location`, iterate the list and return the first element that actually unwraps to something usable (has a resolvable `geometry`/`bbox`/`centroid`, or is itself a bare GeoJSON dict), rather than the first truthy one.

## Acceptance criteria

- [ ] `[{no usable geometry}, {usable geometry}]` resolves using the second element
- [ ] `[{usable geometry}, {no usable geometry}]` is unchanged (still resolves using the first)
- [ ] Existing `test_translate_spatial_location_array_uses_first` still passes
- [ ] Unit test added for the reversed-order case — there is currently none

## Context

Found in #6038 (SPIKE-3), finding 3.

cody-seibert-gsa:
## Summary

`_unwrap_location()` selected the first *truthy* element of a `spatial` Location array, not the first *usable* one. A non-empty dict with no geometry (e.g. a `prefLabel`-only Location) is just as truthy as one with a `bbox`, so `[{prefLabel-only}, {bbox-bearing}]` resolved to `NULL` while the reverse order worked fine - whether the record kept its geometry depended on element order in the source feed, not on which element actually had any.

## What changed

- Split `_unwrap_location` into `_unwrap_single_location` (the existing single-Location logic, unchanged) plus a list loop that returns the first element `_unwrap_single_location` resolves to something truthy, instead of picking the first truthy raw element up front.
- Added unit tests for the previously-broken reversed order, and for an array where nothing is usable.

Closes GSA/data.gov#6296


tests/unit/test_utils.py
```diff
@@ -378,6 +378,23 @@ def test_translate_spatial_location_array_uses_first(self):
...
+            {"@type": "Location", "prefLabel": "Nebraska"},
+            {"@type": "Location", "geometry": "POINT (1.0 1.0)"},
+        ]
+        assert translate_spatial(locations) == (
+            '{"type": "Point", "coordinates": [1.0, 1.0]}'
+        )
+
+    def test_translate_spatial_location_array_all_unusable_is_empty(self):
+        locations = [
+            {"@type": "Location", "prefLabel": "Nebraska"},
```
rshewitt, line 423: we have the locations table to resolve named locations like "Nebraska". In production i would expect `translate_spatial` to most likely return a geojson object for that state.

tests/unit/test_utils.py
```diff
@@ -382,6 +382,67 @@ def test_translate_spatial_location_array_uses_first(self):
             '{"type": "Point", "coordinates": [0.0, 0.0]}'
         )
 
+    def test_translate_spatial_location_array_skips_unusable_leading_element(self):
```
rshewitt: is this test necessary anymore?

cody-seibert-gsa: doesn't seem like it, I'll remove it

rshewitt (approved):
looks good!

PR #927, opened 2026-09-08
munge spatial for ISO records and fix anti-meridian splitting logic

rshewitt:
# Pull Request

Related to [6280](https://github.com/GSA/data.gov/issues/6280)

## About
preprocesses ISO spatial values (munges them into a geojson polygon) in order to make use of anti-meridian splitting logic

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


harvester/harvest.py
```diff
@@ -1408,7 +1409,11 @@ def _dataset_payload(self, metadata: dict) -> dict:
             "last_harvested_date": self.date_finished,
         }
 
-        translated_spatial = translate_spatial_to_geojson(metadata.get("spatial"))
+        spatial_value = metadata.get("spatial")
+        if self.harvest_source.schema_type.startswith("iso19115"):
```
akuny: I think we need to add `if spatial_value and` to this if condition. Otherwise, it looks like an ISO record with no `spatial` key raises `AttributeError` here.

harvester/utils/general_utils.py
```diff
@@ -962,12 +995,22 @@ def spatial_wrap_around_meridian(geom):
     # Unclear why this is needed, but to work with the right hand rule.
     # https://medium.com/@jinagamvasubabu/solution-polygons-and-multipolygons-should-follow-the-right-hand-rule-27b96fa61c6
     new_geom["coordinates"][0][1].reverse()
+
+    new_geom["coordinates"] = [
```
akuny, line 1019: Now that the two boxes are two separate polygons, both need their points listed counter-clockwise, but the `.reverse()` above is left over from when the second one was a hole and only ever flips one of the two. So, one box always ends up backwards.

Paste the expected dict from `test_iso_dataset_payload_meridian` into `validate_geometries` and it fails `exterior_not_ccw`. Same for the two `test_translate_spatial_over_meridian_*` tests. Nothing flags it because `validate_geojson` hands this output back without checking it again.

Can we flip whichever box comes out backwards, and update the three expected values in `test_record_sync.py` and `test_utils.py` to match?

akuny (changes requested):
Looks good apart from two spots where we we might be able to tighten things up.

harvester/utils/general_utils.py
rshewitt: yeah sure good catch

harvester/harvest.py
rshewitt: dict.get should return None if the key isn't present and `translate_spatial_to_geojson` handles situations when the input is None so i think we're good?

akuny: Does the mdtranslator ISO writer always emit spatial, even when the source has no bounding box? If it can omit it or send null, this errors the record. Line 1414 hands the value to `munge_spatial` first, so the `if not spatial_input` guard in `translate_spatial_to_geojson` never gets reached.

rshewitt: ah yeah i see. i'll add the check

akuny (approved):

PR #930, opened 2026-09-09
Give distinct CatalogRecords distinct synthetic identifiers

soroush5:
`backfill_catalog_record_identifiers` synthesized the @id from `primaryTopic|modified` only, so two different records sharing those fields (or both missing them) got the exact same @id, and `filter_duplicate_identifiers` then wrongly dropped all but the first — silent data loss.

Now hashes the full record content: distinct records get distinct @ids, byte-identical records still collapse (true duplicates), and ids stay stable across runs. `TestBackfillCatalogRecordIdentifiers` (5 existing + 1 new): 6 passed.


soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

PR #934, opened 2026-09-10
6297: fall back to altLabel for named-place spatial values

### LINKED ISSUE #6297: Decide: should `Location.altLabel` reach the gazetteer lookup when `prefLabel` is absent? (v1.1 parity gap, altLabel half)

## Problem

`Location.prefLabel` / `altLabel` never reach the gazetteer lookup, so a named-place `spatial` value that would resolve under v1.1 resolves to `NULL` under v3.0.

**Update 2026-09-10:** [PR #925](https://github.com/GSA/datagov-harvester/pull/925) fixed this for `prefLabel`. The `altLabel` half described below is still open.

`_unwrap_location()` ([general_utils.py:1374](https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/general_utils.py#L1374)) now falls back to a Location's `prefLabel` via the new `_extract_pref_label()` helper ([general_utils.py:1354](https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/general_utils.py#L1354)) when no element in the input has usable `geometry`/`bbox`/`centroid`. That string flows into `translate_spatial()` ([general_utils.py:1402](https://github.com/GSA/datagov-harvester/blob/main/harvester/utils/general_utils.py#L1402)) and on into its existing `dbi.get_geo_from_string()` gazetteer lookup — the same path a v1.1 plain string takes.

`_extract_pref_label()` only reads `prefLabel`, though. It has no fallback to `altLabel`, so a Location with `altLabel` but no `prefLabel` anywhere in the input still returns `None` and resolves to `NULL`, exactly as before PR #925.

| Input | Result | Gazetteer lookup attempted |
|---|---|---|
| `"California"` (v1.1 plain string) | ✅ California polygon | `['California']` |
| `{"@type": "Location", "prefLabel": "California"}` (v3.0) | ✅ California polygon (fixed by #925) | `['California']` |
| `{"@type": "Location", "altLabel": "California"}` (v3.0, no `prefLabel`) | ❌ `NULL` | none |

So an agency that migrates `"spatial": "California"` to `"spatial": {"@type": "Location", "altLabel": "California"}` (using `altLabel` instead of `prefLabel`) still loses spatial search coverage on that record with no warning.

## Current test coverage

- `test_translate_spatial_location_pref_label_only_db_hit` / `_db_miss` (`tests/unit/test_utils.py`) cover the now-fixed `prefLabel`-only case.
- `test_translate_spatial_location_array_falls_back_to_pref_label_db_hit` / `_db_miss` include an `altLabel` field, but only ever alongside a `prefLabel` on a *different* array element — they assert the `prefLabel` wins, not that a standalone `altLabel` resolves.
- No test exercises an `altLabel`-only Location (no `prefLabel` present anywhere in the input). That case is unhandled today.

## Decision needed

- **Option A — extend parity to `altLabel`:** in `_extract_pref_label()` (or a renamed equivalent), fall back to `altLabel` when `prefLabel` is absent/blank, before giving up. Add a test for an `altLabel`-only Location.
- **Option B — keep current behavior:** named-place-only Locations using only `altLabel` stay unresolved. If kept, consider surfacing a warning so agencies relying on `altLabel` notice the coverage gap instead of it failing silently.

`altLabel` is `Optional` (repeatable) in the DCAT-US 3.0 `Location` definition; `prefLabel` is `Recommended`. It's plausible some feeds populate only `altLabel` for a named place, so this gap may still matter even with `prefLabel` parity restored.

## Context

Found in #6038 (SPIKE-3), finding 4. Related to #6294/#6295 (same SPIKE) and the `locations` gazetteer table. `prefLabel` half resolved by #925 (closes #6296).

cody-seibert-gsa:
## Summary
- Closes GSA/data.gov#6297.
- `Location.altLabel` never reached the gazetteer lookup, so a v3.0 `spatial` value using only `altLabel` (no `prefLabel`) resolved to `NULL` even when the equivalent v1.1 plain string would have resolved. Extends the `prefLabel` fallback from #925/#6296 to also try `altLabel` when `prefLabel` is absent anywhere in the input.
- Decided priority order for a Location's spatial fields (per maintainer input on #6297) is now `bbox > centroid > geometry > prefLabel > altLabel`, first valid value wins. Previously `_unwrap_single_location` checked a single Location's own fields in `geometry > bbox > centroid` order; that's now `bbox > centroid > geometry`.
- `_extract_pref_label` is generalized to `_extract_label(input_value, field_name)` so the same helper handles both `prefLabel` and `altLabel`.

## Test plan
- [x] `pytest tests/unit/test_utils.py -k translate_spatial` — 31 passed, including new tests:
  - `test_translate_spatial_location_bbox_centroid_geometry_all_present_prefers_bbox`
  - `test_translate_spatial_location_centroid_and_geometry_prefers_centroid`
  - `test_translate_spatial_location_array_pref_label_wins_over_earlier_alt_label`
  - `test_translate_spatial_location_array_falls_back_to_alt_label_db_hit`
  - `test_translate_spatial_location_alt_label_only_db_hit` / `_db_miss`
- [x] `pytest tests/unit/` — 571 passed (no regressions)

### DIFF
--- a/harvester/utils/general_utils.py
+++ b/harvester/utils/general_utils.py
@@ -1332,7 +1332,11 @@ def _get_geo_lookup_interface():
 
 def _unwrap_single_location(input_value):
     """Resolve a single (non-array) spatial value to its geometry-bearing
-    value, or None if it has nothing usable."""
+    value, or None if it has nothing usable.
+
+    When a Location defines more than one of these at once, bbox outranks
+    centroid outranks geometry (decided in #6297).
+    """
 
     if (
         isinstance(input_value, dict)
@@ -1342,7 +1346,7 @@ def _unwrap_single_location(input_value):
         }
         <= input_value.keys()
     ):
-        for field in ("geometry", "bbox", "centroid"):
+        for field in ("bbox", "centroid", "geometry"):
             value = input_value.get(field)
             if value:
                 return value
@@ -1351,9 +1355,9 @@ def _unwrap_single_location(input_value):
     return input_value
 
 
-def _extract_pref_label(input_value):
-    """Return a Location's prefLabel as a plain string, or None if absent,
-    blank, or not a dict.
+def _extract_label(input_value, field_name):
+    """Return a Location's `field_name` (prefLabel or altLabel) as a plain
+    string, or None if absent, blank, or not a dict.
 
     Consulted by _unwrap_location only when nothing in the whole input has
     usable geometry - real geometry always outranks a named-place fallback.
@@ -1365,9 +1369,9 @@ def _extract_pref_label(input_value):
 
     if not isinstance(input_value, dict):
         return None
-    pref_label = input_value.get("prefLabel")
-    if isinstance(pref_label, str) and pref_label.strip():
-        return pref_label
+    label = input_value.get(field_name)
+    if isinstance(label, str) and label.strip():
+        return label
     return None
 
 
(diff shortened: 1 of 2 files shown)


jbrown-xentity (approved):
LGTM

PR #935, opened 2026-09-10
6277: fix progress bar CSS class collision across job rows

### LINKED ISSUE #6277: Percentage on harvest incorrect with multiple jobs
When I ran the harvest for a couple hundred jobs whenever there were two or more actively processing jobs they would all show the same completion percent. 

## How to reproduce

1. Queue many longer force harvest jobs (Probably at least 3-5) 
2. Wait for them to be at different percent completion. 
3. Go to the metrics page and check the percent done value. 

## Expected behavior

[ ] Percent value should match what happens when you click on the row. 

## Actual behavior

[ ] Percent value are all the same for each row.

## Sketch

Aguiardavidm:
https://github.com/GSA/data.gov/issues/6277

- Every job row's progress bar rendered a `<style>` block targeting the same `.this-progress` class
- When multiple jobs were in progress, the browser cascade applied whichever row rendered last to every progress bar on the page
- Scoped each row's selector to `.progress-{{ job.id }}` so each bar reflects its own job's percent complete
- Kept the nonce'd `<style>` tag rather than an inline `style` attribute, since the app's CSP `style-src-attr` has no nonce support and would silently block inline attributes

## Test plan
- [x] Verified via curl against local docker env that each row now renders a unique `.progress-<job-id>` selector/class pair

### DIFF
--- a/app/templates/components/job-table/job_table.j2
+++ b/app/templates/components/job-table/job_table.j2
@@ -57,11 +57,11 @@
                         <div class="progress-meter">
                             {# [:-1] slices off "%" #}
                             <style nonce="{{ csp_nonce() }}">
-                              .this-progress {
+                              .progress-{{ job.id }} {
                                 --progress: {{ job.percent_complete[:-1] }};
                               }
                             </style>
-                            <div class="progress-percent this-progress"></div>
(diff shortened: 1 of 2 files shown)


jaredb96 (approved):
LGTM

PR #936, opened 2026-09-11
add theme keyword to opensearch mapping

rshewitt:
# Pull Request

Related to [6161](https://github.com/GSA/data.gov/issues/6161)

## About
adds theme keyword for opensearch mapping. catalog [PR](https://github.com/GSA/datagov-catalog/pull/386/)

## PR TASKS

- [ ] Code well documented
- [ ] Files linted

### DIFF
--- a/search/mappings.py
+++ b/search/mappings.py
@@ -53,6 +53,12 @@ MAPPINGS = {
             "type": "text",
             "analyzer": TEXT_ANALYZER,
             "search_analyzer": TEXT_ANALYZER,
+            "fields": {
+                "normalized": {
+                    "type": "keyword",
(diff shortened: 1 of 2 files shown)


jbrown-xentity (commented):
I'm unsure if this requires a re-index or not. We haven't done a change to a field type before... Gemini AI seems to think it would require that. If so, suggest pushing to dev with the rebuild index process to verify everything will go ok!

rshewitt:
@jbrown-xentity here's a [query](https://catalog-dev.data.gov/search?theme=card%20services%20program) on dev

jbrown-xentity (approved):
LGTM, re-index out of an abundance of caution.

PR #939, opened 2026-09-14
Document the OpenSearch reindex process

cody-seibert-gsa:
## Summary
- Add `docs/ops/opensearch-reindex.md`, leading with the actual process: open a PR, attach the `force re-index recommended` label, and merge it — the deploy pipeline handles the rest.
- The doc also covers the manual `flask search compare/reset-mapping/rebuild-index/delete-index` CLI commands and the three GitHub Actions workflows (Sync OpenSearch, Rebuild OpenSearch Index, Delete OpenSearch Physical Index) as ops-troubleshooting-only tools, distinct from the normal label-driven process.
- Link the new doc from `docs/ops/README.md` and add explicit links to the ops docs from the root `README.md`'s Documentation section (previously only reachable via the generic "docs folder in general" bullet).

## Test plan
- [x] Verified referenced CLI flags/behavior against `app/commands/search.py`
- [x] Verified referenced workflow behavior against the `.github/workflows/*.yml` files
- [x] Verified the `force re-index recommended` label string against `.github/scripts/detect-opensearch-migration-label.sh`
- [ ] Docs-only change; no functional testing required


rshewitt (approved):
Awesome thanks for this!

PR #940, opened 2026-09-14
Fix DCAT-US v3.0 temporal serialization for OpenSearch indexing

cody-seibert-gsa:
## Summary

`temporal` was grouped into `_normalize_dcat_dates`'s scalar `date_fields` list alongside `modified`/`issued`, so a non-string value fell through to `str(value)`. Under DCAT-US v1.1, `temporal` is a scalar interval string, so this was harmless. Under v3.0, `temporal` is a list of `PeriodOfTime` objects — `str()` on that produces a Python repr string (`"[{'@type': 'PeriodOfTime', ...}]"`), which isn't valid JSON and can't be parsed back.

Fix: serialize `temporal` the same way `spatial` already is — via `_serialize_dcat_value` (`json.dumps` for dict/list, `str()` otherwise) — instead of the scalar-date path.

Found in GSA/data.gov#6038 (SPIKE-3).

## Test plan

- [x] Added unit test asserting a v3.0 `PeriodOfTime` list normalizes to valid, `json.loads`-able JSON (previously produced an unparseable repr string)
- [x] Added end-to-end `dataset_to_document` test covering the same case
- [x] `pytest tests/integration/search/test_documents.py tests/unit/test_search_commands.py` — 87 passed (all tests touching this code path)


jbrown-xentity (approved):
I think I agree with this approach that removes temporal from munging and leaves it alone. If we want to add it to searchability, we should add it to opensearch with it's own field, not changing the provided value.

PR #942, opened 2026-09-16
5485: Standardize Validation Errors

akuny:
Addresses https://github.com/GSA/data.gov/issues/5485

Standardizes the stored type for validation failures in the `harvest_record_error` table.

No migration, since type is a free-form String column. No backfill. Existing rows keep ValidationError, and all read paths already accept both names.

Clean lint, targeted validation/exception/report tests.


jbrown-xentity (approved):
LGTM

PR #943, opened 2026-09-16
add logs around boundary of job start and cleanup

rshewitt:
# Pull Request

Related to [6324](https://github.com/GSA/data.gov/issues/6324)

## About
idle in transaction [alert](https://one.newrelic.com/alerts/issue?account=1601367&duration=259200000&state=ea467874-35ec-04b4-07f0-4187b54585f7) kicked off leading to an investigation into the queries responsible. the largest offender is likely `interface.pget_harvest_jobs` which occurs in the `start_job` function. however, both `_clean_old_jobs` and `_start_new_jobs` mix db calls and cf network calls. both also query the job table. the outcome of adding these logs is understanding how long each of those functions take to process and to compare that against idling transactions.

here's some info on `_clean_old_jobs`. it first makes a db call, then gets the running app tasks, then makes more db calls. here's a timestamp of how long that could take (just 1 test).
```terminal
>>> def f():
...  start = datetime.datetime.now()
...  cf_handler.get_running_app_tasks()
...  print(datetime.datetime.now() - start)
... 
>>> f()
0:00:12.739346
```

i also removed an assertion checking the exact number of logger.info calls because i think it's ridiculous for a test to fail just because i added logs

## PR TASKS

- [ ] Code well documented
- [ ] Files linted


jaredb96 (approved):
LGTM

PR #944, opened 2026-09-16
6338: Fix harvest error notification email showing false success

Aguiardavidm:
https://github.com/GSA/data.gov/issues/6338

## Summary
- When a harvest job fails during extraction (`ExtractExternalException`/`ExtractInternalException`), `report()` now passes the job's actual status through to `send_notification_emails`, and treats a job-level error as a trigger for the `on_error`/`on_error_or_update` notification frequencies (previously those only checked `records_errored`, which stays 0 when extraction fails before any record is processed).
- `send_notification_emails` branches on status: errored jobs get subject "Harvest Job Failed" and a body with the actual error message (pulled from `get_harvest_job_errors_by_job`), instead of the hardcoded "successfully completed" text with all-zero counts.

## Test plan
- [x] Added `test_send_notification_emails_reports_job_error` covering the error-path email content and subject.
- [x] Updated `test_critical_job_error_status_not_overwritten` to mock `send_notification_emails` (now invoked in that scenario) and assert it's called with `job_status="error"`.
- [x] `make lint-check` equivalent (ruff, black, isort) passes.
- [x] Ran affected integration suites locally; confirmed 5 unrelated pre-existing failures in `test_harvest_job_full_flow.py` also occur on `main`.


jbrown-xentity (approved):
LGTM, adding error to the title on job errors helps data providers know when something is really important.

PR #946, opened 2026-09-17
6342: Make organization logo optional, default to placeholder

Aguiardavidm:
https://github.com/GSA/data.gov/issues/6342

- Organization logo was `DataRequired()` in the form, blocking org creation/edit when a provider's logo URL was invalid or missing. The `logo` column is nullable at the DB level, so no migration is needed.
- Switched the `logo` field validator to `Optional()` + `URL()`, matching the existing `code_repo_url` pattern. This also removes the red required-asterisk, which the shared edit template derives automatically from the field's validators.
- Per Fuhu's feedback on the Slack thread: no default/placeholder logo on the harvester side. A blank logo is now stored as `None` — catalog.data.gov already renders its own placeholder for orgs with no logo, so keeping a second placeholder image in sync between harvester and catalog would be redundant. Removed the now-unused `app/static/assets/img/placeholder-organization.png` asset and its references in the CLI default and fixtures.

## Test plan
- [x] `poetry run pytest tests/integration/app/test_forms.py tests/integration/app/test_routes.py` — all passed, including a new test asserting a blank logo is stored as `None`
- [x] ruff / isort / black pass on changed files


FuhuXia:
no need to use placeholder in both apps (catalog and harvest), otherwise you have two placeholder images to synchronize. 

I would say remove placehold from harvest. harvest does not need any image.

FuhuXia (approved):
LGTM
trivial: the hint still says "Defaults to a placeholder if left blank". If you mean the org will displayed on catalog site with a placeholder image, then it is not wrong.

PR #948, opened 2026-09-18
6259: remove records total

akuny:
Addresses https://github.com/GSA/data.gov/issues/6259

The "Records total" row on the harvest job page has been removed. The `records_total` column stays in the model/DB. This change only affects the view.

## Changes

- `app/templates/view_job_data.html`: filter `records_total` out of the job-config table loop.
- `tests/playwright/test_harvest_job.py`: drop the now-invalid "Records total" / value assertions.

## Verification

- [x] Lint passes
- [x] Playwright passes


jbrown-xentity (approved):
LGTM

PR #949, opened 2026-09-18
5907: Fix order_by sort direction on list endpoints

akuny:
Addresses https://github.com/GSA/data.gov/issues/5907

## Issue

`order_by=-name` returned ascending order with no error. In `database/interface.py`, `order_by_helper` only recognized the literal `"asc"` and hard-coded the `date_created` column, discarding both the field name and the `-` prefix; `pget_db_query` also early-returned `organizations` and `harvest_sources` unordered since neither has `date_created`.

## Changes

- Rewrite `order_by_helper` to parse `field` (asc), `-field` (desc), and legacy `"asc"`/`"desc"`/empty (default column). Default is `date_created` if present, else `name`. Unknown columns fall back to the default without erroring.
- Route `organizations` and `harvest_sources` through the helper.

## Note

For `date_created` models, omitting `order_by` on the raw API now defaults to ascending (was descending). The UI is unaffected, and it passes `order_by="desc"` explicitly.


jaredb96 (approved):
LGTM

PR #950, opened 2026-09-18
6348: Show friendly error on duplicate harvest source URL edit

Aguiardavidm:
https://github.com/GSA/data.gov/issues/6348

- Editing a harvest source's URL to one already used by a different source threw an unhandled 500 (`sqlalchemy.exc.IntegrityError` on the `harvest_source_url_key` unique constraint), confirmed via New Relic logs on staging.
- The add flow already guards against this (`create_harvest_source` in `app/deps.py` does a `get_harvest_source_by_url` pre-check plus a friendly message) - the edit flow never got the same treatment.
- Added `HarvesterDBInterface.try_update_harvest_source`, mirroring the existing `try_add_harvest_source` pattern: catches the commit exception and translates it via the existing `harvest_source_save_error_message` helper instead of letting it propagate.
- `edit_harvest_source` now pre-checks for a URL conflict (naming the conflicting source's ID, same as add) and falls back to `try_update_harvest_source`'s error message for anything else, flashing a normal message and redirecting back to the edit form instead of crashing.
- Also fixed a second latent bug on the same line: if the update ever failed for any reason, the old code still dereferenced `source.id` on a `None` result on the next line, which would have raised `AttributeError` instead of showing the flashed message.

## Test plan
- [x] Added `test_edit_harvest_source_duplicate_url_shows_friendly_error`, asserting a 302 back to the edit page and that the source's URL is left unchanged
- [x] `poetry run pytest tests/integration/app/test_forms.py tests/integration/app/test_routes.py` - 87 passed
- [x] ruff / isort / black clean


jaredb96 (approved):
LGTM

PR #951, opened 2026-09-18
6117 access level field support

marcos-nieto-usds:
# Pull Request

Related to 
- https://github.com/GSA/data.gov/issues/6117

### Summary

- Adds a fallback to accessRights when deriving a dataset's normalized access_level for indexing, since accessLevel is not part of the DCAT-US v3.0 schema.

- DCAT-US v3.0 removes accessLevel in favor of accessRights, but the two guidance documents describing this transition are not fully aligned on the expected shape of accessRights:

- The [DCAT-US v3.0 Implementation Guide](https://resources.data.gov/assets/documents/dcat-us-3-implementation-guide.pdf) (authoritative for field definitions) allows accessRights to either directly replace accessLevel with the same three canonical values (public, restricted public, non-public), or contain descriptive free text explaining the access restriction, mirroring v1.1's separate rights field, per M-25-05 supplemental-info guidance.

- The [DCAT-US v3.0 Migration Guide](https://resources.data.gov/resources/dcat-us-3-migration/) is explicitly not authoritative for field definitions (this is mentioned in bold, as the first sentence under "**Details**." "**This page is a technical guide and not meant to provide implementation guidance for M-25-05 or specific field definitions.**"). 
   - The guide offers a sample [transform_access_rights()](https://github.com/GSA/dcat-us/blob/1ca074b67dd1ad3cea8f0dd4ae951864491fd104/jsonschema/transforms.py#L69) that maps each canonical level to one specific descriptive sentence.
- Because these two sources disagree, and the migration guide's sample sentences aren't binding, we can expect real-world harvested accessRights values to arrive in multiple shapes: clean canonical strings, the migration guide's sample sentences, or agency-specific free text conveying the same meaning with different wording.

### Changes
- transforms.py
   - Updated coerce_access_level() to:
      - Return "" when no value is provided (missing accessLevel and accessRights).
      - Return the value as-is (lowercased/cleaned) when it already matches a canonical level (public, restricted public, non-public).
      - Otherwise, infer the canonical level via keyword matching against common phrasing patterns (e.g., "not available", "restricted") to tolerate free-text accessRights values without depending on exact sentence matching.
      - If no keyword matches, return the original (lowercased/cleaned) value unchanged rather than raising or discarding it. Preserves visibility into unresolved values for downstream consumers/reporting rather than silently dropping data.
   - Updated docstrings (module-level and function-level) to describe this as a best-effort normalization rather than a strict three-value guarantee, and to document the rationale/sourcing for the fallback behavior.
- documents.py
   - index_fields["access_level"] now falls back to accessRights when accessLevel is absent:
      - coerce_access_level(dataset.dcat.get("accessLevel") or dataset.dcat.get("accessRights") or None)
   - accessLevel is preferred when both fields are present, consistent with existing behavior.
- test_dataset_document.py
   - Added/updated coverage for:
      - Setting access_level directly from accessLevel
      - Normalizing whitespace/casing on accessLevel
      - Preferring accessLevel over accessRights when both are present
      - Falling back to accessRights when accessLevel is absent, using realistic free-text values (not the bare enum string) to actually exercise the keyword-matching logic
      - Tolerating agency-specific/reworded accessRights phrasing that isn't an exact match to the migration guide's sample sentences.
- ACCESS_LEVEL_FILTER intentionally remains access_level-only, since the harvester's search API has no documented public contract requiring the accessLevel alias. [Source 1, harvester-api](https://resources.data.gov/harvester-api/) and [Source 2, harvester-swagger](https://harvest.data.gov/openapi/docs). 

### Testing
- All new and existing tests pass.
- Manually verified against sample DCAT-US v3.0 payloads with accessRights-only datasets (no accessLevel) to confirm correct fallback resolution for each canonical level.
- As shown in https://github.com/GSA/data.gov/issues/6117#issuecomment-5732152039, verified that attempting to use a non-existent accessLevel gets validated on the harvester side.


jaredb96:
@marcos-nieto-usds Same question I have for the other PR - merge this into develop and try to test there?

marcos-nieto-usds:
> @marcos-nieto-usds Same question I have for the other PR - merge this into develop and try to test there?

Will do.

PR #955, opened 2026-09-21
5293: Update waf datetime parsing for different servers

### LINKED ISSUE #5293: update waf datetime parsing for different servers
## User Story

In order to accurately derive file modified dates from waf source, datagov wants to update waf datetime parsing to handle different formats/servers

## Acceptance Criteria

- [ ] GIVEN a waf source \
  AND a datetime format not like "2025-01-22 23:21" (e.g. "05-Jun-2025 17:17") \
  WHEN waf datetime parsing happens \
  THEN the other format is parsed correctly \

## Background
- we consider the modified date in waf sources to avoid unnecessary downloads (if the source document is older than what we have then skip it)
- currently, our waf datime parsing function only handles "2025-01-22 23:21" format but we want to make sure other formats are supported to. that function will pad with a placeholder datetime but if we can correctly parse the date on the page that would be better.

## Sketch
- identify different datetime formats (based on server? what if the server type isn't in the response header?)
- implement those datetime formats based on the server type? or just try them all

jaredb96:
WIP


jbrown-xentity (approved):
LGTM, thanks for going through all the test cases!

